Issue #1, opened 2022-11-04
Move to upstream parboiled2 dependency


jrudolph:
Akka HTTP right now contains a copy of parboiled2 in its sources (with some slight additions). This decision was originally made on the grounds that Akka should not have many dependencies on Scala libraries, so it can be released quickly after a new (incompatible) Scala version has been released.

By now, parboiled2 does not depend on shapeless any more and the ecosystem has more stabilized (and after all, maintaining the fork has not helped at all with migration to new Scala versions...). So, I'd suggest we move to upstream parboiled2.

nvollmar:
That would make sense

pjfanning:
@jrudolph do you think this is best done now or should we wait till after the v1.0.0 release?

pjfanning:
There is some discussion about bringing this into v1.0.0 release in https://github.com/apache/incubator-pekko-http/issues/39.

It would simplify the license files in pekko-http if we just used the parboiled2 jar instead of embedding the source in our code base.

spangaer:
If there's no functional difference by using upstream, I don't think switching to the dependency is real violation of the v1.0 intent.

As long as Pekko 1.0.0 exhibits the same behavior as Akka 2.6.x feature for feature, bug for bug after the rename refactor in consuming code.

Seetaramayya:
## Approach 1: As Is approach (convert to dependency in `1.1.0`)

(+) As @spangaer mentioned, we can reason about bugs (if there are any) `1.0.0` does not do anything apart from different packaging. No need to worry about performance impacts. 
(+) No need to run heavy performance benchmarks before releasing
(-) As @pjfanning mentioned, licence files management would be easy (I really don't know how painful)

 ## Approach 2: Convert to dependency right now (in `1.0.0`)

(+) Licence file management would be easy
(-) IMHO: Performance bench marks needs to be run before releasing 
(-) Assume underlying dependency `parboiled` has (`functional` or `performance` ) bug which resulted lot of side effects in the system narrow down the issue would be bit more complicated than the approach (I am talking from relative point of view). 


I would prefer `Approach 1` unless there is a huge impact on licensing part (I don't know have knowledge in the area). IMHO: With this approach we are separating the concerns

spangaer:
For clarity, I favor approach 2 (do it in 1.0.0). I do believe some validation should be made, but best effort only, it shouldn't absorb all energy either, otherwise there's nothing to be gained.

jrudolph:
[...]

PR #14, opened 2022-12-10
moving to upstream parboiled2 dependency

luksow:
Hi,

since https://github.com/apache/incubator-pekko-http/issues/1 was rather easy, I took stab at it. The parboiled2 interface changed a bit over the time, so more changes were needed than just the imports.

Thing to consider is to remove akka-parsing completely but I think it was partially sorted out in https://github.com/apache/incubator-pekko-http/tree/scala-3 branch


pjfanning:
Thanks for the PR. We still need to decide if we want to make this change yet. For v1.0.0, we generally want minimal changes.
The scala-3 branch is shaping up to be the basis of the v1.1.0 release and we may end targeting this work to that release and possibly that branch.
The scalfmt checks are failing and it would be a good idea to change the order of the imports so that the org.parboiled imports are not with the Akka imports.

luksow:
@pjfanning thanks for your comments. I did scalafmt and also reorganized imports so that org.parboiled2 is always after akka's.

After you decide the faith of v1.0.0 - let me know. I'm happy to target v1.1.0 or scala-3 branch instead.

pjfanning:
@mdedetrich, @jrudolph, @nvollmar - I think we should consider treating the scala-3 branch as the v1.1.0 branch. If this PR was retargeted to the scala-3 branch, would any of you support approving the PR?

nvollmar:
I think that would make sense.

mdedetrich:
I would also support putting this into 1.1.x, will retarget the PR when the 1.1.x branch is out. @luksow Thanks a lot for the work, much appreciated!

pjfanning:
@luksow would you consider updating this PR to work with the latest main branch? The `akka` package classes have moved to `org.apache.pekko`. The main branch should be more stable now.

We are actually thinking about merging this into main for the v1.0.0 release.

mdedetrich:
> @luksow would you consider updating this PR to work with the latest main branch? The `akka` package classes have moved to `org.apache.pekko`. The main branch should be more stable now.
> 
> We are actually thinking about merging this into main for the v1.0.0 release.

I was also going to mention this. There are still a couple more big structural changes that need to be done (i.e. configuration change from `akka` to `pekko` as well as renaming types that have the akka in them. I don't think that these changes would cause further merge conflicts for this kind of PR but I am going to focus on this for the week.

pekko-http-core/src/main/scala/akka/http/impl/model/parser/HeaderParser.scala
```diff
@@ -108,25 +108,25 @@ private[http] object HeaderParser {
 
   object EmptyCookieException extends SingletonException("Cookie header contained no parsable cookie values.")
 
-  def lookupParser(headerName: String, settings: Settings = DefaultSettings): Option[String => HeaderParser#Result] =
-    dispatch.lookup(headerName).map { runner => (value: String) =>
```
jrudolph, line 112: We should keep this optimization, maybe we have to keep at our own version of `DynamicRuleDispatch`?

jrudolph: Maybe, let's try to upstream the optimization and go back to the simpler version for now (which adds an extra hashmap lookup per header parsed).

mdedetrich: Makes sense to me

jrudolph:
I added a merge commit to fix the conflicts.

mdedetrich:
> I added a merge commit to fix the conflicts.

Thanks, I just quickly checked the PR and I don't see any class files containing Akka being modified so there shouldn't be any major merge conflicts later

pekko-http-core/src/main/scala/akka/http/impl/model/parser/HeaderParser.scala
mdedetrich: > Maybe, let's try to upstream the optimization and go back to the simpler version for now (which adds an extra hashmap lookup per header parsed).

If we do this then lets make an issue on it.

http-core/src/main/scala/org/apache/pekko/http/impl/model/parser/CacheControlHeader.scala
```diff
@@ -5,11 +5,10 @@
 package org.apache.pekko.http.impl.model.parser
 
 import org.apache.pekko
-import pekko.parboiled2.Parser
 import pekko.http.scaladsl.model.headers._
 import CacheDirectives._
 
-private[parser] trait CacheControlHeader { this: Parser with CommonRules with CommonActions with StringBuilding =>
+private[parser] trait CacheControlHeader { this: HeaderParser =>
```
jrudolph, line 11: Somehow the Scala compiler got confused here about the more restricted self-type, not sure what's the issue, but that works for now.

http-core/src/main/scala/org/apache/pekko/http/impl/model/parser/HeaderParser.scala
```diff
@@ -110,11 +110,11 @@ private[http] object HeaderParser {
   object EmptyCookieException extends SingletonException("Cookie header contained no parsable cookie values.")
 
   def lookupParser(headerName: String, settings: Settings = DefaultSettings): Option[String => HeaderParser#Result] =
-    dispatch.lookup(headerName).map { runner => (value: String) =>
-      import pekko.parboiled2.EOI
+    Some { (value: String) =>
```
jrudolph, line 113: I changed that to keep the old signatures even if the new one might "lie" because it may still fail later with a `RuleNotFound` exception (only if we are adding ModeledHeaders without accompanying it with the correctly named parser).

jrudolph (approved):
lgtm but has to wait for a decision in #1.

pekko-http-core/src/main/scala/akka/http/impl/model/parser/HeaderParser.scala
jrudolph: Created https://github.com/apache/incubator-pekko-http/issues/45

jrudolph:
(Fixed the test fixes I accidentally reverted from @luksow's state)

luksow:
Wow, you're moving quick! I barely was able to have a look and all work is already done :) Glad to see it moving forward.

LGTM.

mdedetrich:
> Wow, you're moving quick! I barely was able to have a look and all work is already done :) Glad to see it moving forward.

Yeah this is a common problem we have....

There is still one issue though, which is that the PR is unable to be merged because it needs to be rebased due to conflicts (at least thats what github is complaining)

luksow:
@mdedetrich: here it says it's good to go :thinking: 
image

mdedetrich:
I need to download your version of github :D

image

spangaer:
Tie breaker
image


Maybe you should try light mode 😉 

Perhaps it's resolved by now, I saw the same yesterday evening, but it's not the first time I notice GitHub being "eventually consistent"

mdedetrich:
So I noticed that on my phone its fine and on the desktop its still broken, so this is probably the most liberal definition of eventual thatI have seen in a while.

Anyways it just means someone else will have to merge.

mdedetrich:
> The reason is probably that you selected "rebase and merge" which isn't possible cleanly due to my merge commit. But we can squash here.

You are right, this is a case of UI ergonomics because the drop down button which lets you change the merge strategy was kind of greyed out (implying it was disabled), probably it being mixed with dark theme did not help.

mdedetrich (approved):
lgtm

pjfanning:
could we hold off for a couple of days? - on #1, it has been suggested that we run a perf test - may be over cautious but may be no harm if someone has time to run the benchmarks in this project (with and without this change)

mdedetrich:
Sure

mdedetrich:
As per discussion at https://github.com/apache/incubator-pekko-http/issues/1#issuecomment-1432884633 I have applied the `upstream` label (indicating that the PR is waiting for some upstream changes).

jrudolph:
Cleaned up the commits into a single one based on top of current master.

jrudolph:
> As per discussion at [#1 (comment)](https://github.com/apache/incubator-pekko-http/issues/1#issuecomment-1432884633) I have applied the `upstream` label (indicating that the PR is waiting for some upstream changes).

I don't think we should hold off for any upstream changes as discussed above and in #1, the performance hit is currently minor and can partly be solved without upstream changes.

pjfanning (approved):
lgtm - the perf hit is small

pjfanning:
@luksow this change is large enough that we probably need a CLA. If you don't have one already, could you get an iCLA as documented in https://www.apache.org/licenses/contributor-agreements.html ?

jrudolph:
> @luksow this change is large enough that we probably need a CLA. If you don't have one already, could you get an iCLA as documented in [apache.org/licenses/contributor-agreements.html](https://www.apache.org/licenses/contributor-agreements.html) ?

AFAIU, we have to collect CLAs for every contributor. We have the problem here that if we don't get a CLA we will have to resubmit the exact same PR with a new author because it is so trivial that there is no other alternative to these exact changes. After all only 66 lines were updated or added here.

pjfanning:
In practice, ASF projects don't collect CLAs from every user who submits a PR. There is no magic criterion for what makes a PR significant enough to require one but the PR changes 70 files, even if the changes are not very large.

Let's see what @luksow says, if he doesn't have time to sort out the CLA, we are probably going to have to look at alternatives.

luksow:
@pjfanning @jrudolph I've just signed and sent ICLA.

mdedetrich:
> In practice, ASF projects don't collect CLAs from every user who submits a PR. There is no magic criterion for what makes a PR significant enough to require one but the PR changes 70 files, even if the changes are not very large.

My understanding is also the same, an ICLA is required for significant contributions but the definition of this is entirely subjective.

PR #15, opened 2023-01-05
Remove pekko prefix from modules

mdedetrich:
This PR makes pekko-http reflect the same dir structure as every other pekko project (including pekko core as of https://github.com/apache/incubator-pekko/pull/71). Unlike pekko-core the changes are much more trivial, aside from the dir renames only paradox markdown doc links and `build.sbt` needed to be updated.

In regards to published modules with `publishLocal`

Akka ASL 2.0 with org `com.typesafe.akka`
```
akka-http-bom_2.13
akka-http-caching_2.13
akka-http-core_2.13
akka-http-jackson_2.13
akka-http-scalafix-rules_2.13
akka-http-spray-json_2.13
akka-http-testkit_2.13
akka-http-xml_2.13
akka-http2-support_2.13
akka-http_2.13
akka-parsing_2.13
```

Pekko-http (this branch/PR) with org `org.apache.pekko`
```
pekko-http-bom_2.13
pekko-http-caching_2.13
pekko-http-core_2.13
pekko-http-jackson_2.13
pekko-http-scalafix-rules_2.13
pekko-http-spray-json_2.13
pekko-http-testkit_2.13
pekko-http-xml_2.13
pekko-http2-support_2.13
pekko-http_2.13
pekko-parsing_2.13
```

So everything is working as expected, exceptions have been noted as comments below.


build.sbt
```diff
@@ -62,11 +62,12 @@ lazy val aggregatedProjects: Seq[ProjectReference] = userProjects ++ List[Projec
   httpJmhBench,
   billOfMaterials)
 lazy val root = Project(
-  id = "pekko-http-root",
+  id = "pekko-http",
```
mdedetrich, line 65: This is to be consistent with pekko-core, see https://github.com/apache/incubator-pekko/blob/main/build.sbt#L88-L92

nvollmar (approved):
lgtm

gmethvin (commented):
should we also remove the `http` prefix? Since the `pekko-http` prefix is implied, then don't we want `core` instead of `http-core`, `caching` instead of `http-caching`, etc.? In your current change it looks like we're doing this already with `bill-of-materials` and `parsing`.

mdedetrich:
> should we also remove the `http` prefix? Since the `pekko-http` prefix is implied, then don't we want `core` instead of `http-core`, `caching` instead of `http-caching`, etc.? In your current change it looks like we're doing this already with `bill-of-materials` and `parsing`.

For me the redundant part is the root project (i.e. `pekko`), not what the project is implementing. The changes to `bill-of-materials` was done just to make it consistent with pekko core and the change to `parsing` wasn't exceptional in any way (it was `pekko-parsing` before, not `pekko-http-parsing`.

While I think I understand the sentiment I believe this is going too far, most importantly since there is a relation of the name to the published artifact if we hypothetically rename `http-core` to `core` (for example), the artifact still needs to be called `pekko-http-core` and not `pekko-core` (because otherwise it can conflict with other pekko repos under the same `org.apache.pekko`). Even the current proposed changes is stretching it a bit for current committers and to me what is making it acceptable is that at least for the modules that create artifacts (which `bill-of-materials`/`root`/`docs` are not) there is a logical relation/computation to the artifact

gmethvin:
Yeah, the fact that there are modules like `parsing` that aren't specifically for http functionality makes it weird to remove the http prefix everywhere, ok. I was thinking that the module name would always have `akka-http-` prefixed, but that won't work if we have modules that aren't http.

gmethvin (approved):

PR #19, opened 2023-01-06
put license and notice in all jars

pjfanning:
* I updated the NOTICE to match what we have in main pekko repo
* From my understanding, we do not need to include HPack license - or even include details in our NOTICE file
  * it is Apache licensed
  * https://github.com/twitter/hpack has no NOTICE so we don't have copy that non-existent notice into our notice


nvollmar (approved):
lgtm

jrudolph:
> From my understanding, we do not need to include HPack license - or even include details in our NOTICE file
> 
>     * it is Apache licensed
> 
>     * [twitter/hpack](https://github.com/twitter/hpack) has no NOTICE so we don't have copy that non-existent notice into our notice

Not sure I fully agree. 

 * We should note somewhere that we included code and the notice file seems a good place
 * Since we are adapting akka-http we still have to keep akka-http's NOTICE file (based on the Apache License terms), so we cannot just delete it because we think the information is not useful to keep...

PR #20, opened 2023-01-06
Record fork base commit and version in README

kw217:
It's useful for potential users of Pekko to know what version of Akka it corresponds to. This PR adds that info to the README. It also adds the precise commit of the fork, since that's likely to be of interest to us and to others in the future.

I'd like to propose tagging that commit as well, as `akka-fork-base` or similar - do you agree this would be a good idea?

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,8 @@ general toolkit for providing and consuming HTTP-based services. While
 interaction with a browser is of course also in scope it is not the primary
 focus of Pekko HTTP.
 
+Pekko HTTP is a fork of [Akka HTTP](https://github.com/akka/akka-http) 10.2.x release, prior to the Akka project's adoption of the Business Source License.
+
 Pekko HTTP follows a rather open design and many times offers several different
 API levels for "doing the same thing". You get to pick the API level of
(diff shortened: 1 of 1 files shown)


pjfanning:
I've added a git tag - see https://github.com/apache/incubator-pekko-http/tags and would be -1 on this PR - I think the tag is enough.

kw217:
Thanks for the tag. Would you accept the PR if I remove the commit link? I think the base version is very important - I just worked it out so I can see whether my current project will be able to move across smoothly or not.

pjfanning:
I'm happy with the current pekko repo text.

```
Pekko is a fork of [Akka](https://github.com/akka/akka) 2.6.x, prior to the Akka project's adoption of the Business Source License.
```

So for pekko-http, we can use.

```
Pekko-Http is a fork of [Akka-Http](https://github.com/akka/akka-http) 10.2.x, prior to the Akka project's adoption of the Business Source License.
```

kw217:
Thanks @pjfanning - done. That, plus the new tag, is sufficient for me :-). Please can you approve the workflow?

pjfanning (approved):

PR #30, opened 2023-01-27
Update scalafmt

mdedetrich:
Updates scalafmt and like with the other pekko projects also adds the scalafmt github action while removing the now unnecessary scalafmt checks in `validate-and-test.yml`


pjfanning (approved):
lgtm

PR #31, opened 2023-01-28
Add git blame ignore revs

mdedetrich:
Similar to pekko core, adds a `.git-blame-ignore-revs` which contains the hashes for the previous scalafmt commits.


pjfanning (approved):
lgtm

PR #43, opened 2023-02-14
Rename akka-http user-agent plus server header to pekko-http

mdedetrich:
Currently the `User-Agent` and server headers are set to `akka-http`, this PR changes it to `pekko-http`.


jrudolph (approved):
lgtm

PR #44, opened 2023-02-14
build: run GHA jobs only on main repository and not on forks

jrudolph:


pjfanning (approved):
lgtm - should help GHA save some money

PR #47, opened 2023-02-14
setup publishing to apache nexus

jrudolph:


mdedetrich (approved):
lgtm. Will also need to add the `DISCLAIMER` file using the sbt-apache-sonatype plugin similarly to how it was done with pekko-core but that can be done seperately.

jrudolph:
> lgtm. Will also need to add the `DISCLAIMER` file using the sbt-apache-sonatype plugin similarly to how it was done with pekko-core but that can be done seperately.

That already seems to be working out of the box!?

mdedetrich:
> > lgtm. Will also need to add the `DISCLAIMER` file using the sbt-apache-sonatype plugin similarly to how it was done with pekko-core but that can be done seperately.
> 
> That already seems to be working out of the box!?

Its implemented manually in https://github.com/apache/incubator-pekko-http/blob/main/project/MetaInfLicenseNoticeCopy.scala#L16-L29 but sbt-apache-sonatype handles this for you as well, see https://github.com/apache/incubator-pekko/blob/main/project/AddMetaInfLicenseFiles.scala#L18-L42 (note that `apacheSonatypeDisclaimerFile` is optional)

jrudolph:
> Its implemented manually in [`main`/project/MetaInfLicenseNoticeCopy.scala#L16-L29](https://github.com/apache/incubator-pekko-http/blob/main/project/MetaInfLicenseNoticeCopy.scala?rgh-link-date=2023-02-14T12%3A45%3A10Z#L16-L29) but sbt-apache-sonatype handles this for you as well, see [apache/incubator-pekko@`main`/project/AddMetaInfLicenseFiles.scala#L18-L42](https://github.com/apache/incubator-pekko/blob/main/project/AddMetaInfLicenseFiles.scala?rgh-link-date=2023-02-14T12%3A45%3A10Z#L18-L42) (note that `apacheSonatypeDisclaimerFile` is optional in the plugin which is why we need to specify it in Pekko projects).
> 
> As mentioned before, this makes sense to do in another PR since its not directly related.

Ah, I see, we could basically simplify. Let's do that later, indeed.

jrudolph:
> Test run at [apache/incubator-pekko-http/actions/runs/4173990714](https://github.com/apache/incubator-pekko-http/actions/runs/4173990714)

which seems to run in general but in this case errored out with a 502 from the Apache Nexus...

PR #48, opened 2023-02-14
Replace Akka with Pekko for classes and functions

mdedetrich:
Self explanatory

### DIFF
--- a/docs/src/test/java/docs/http/javadsl/server/PekkoHttp1020MigrationExample.java
+++ b/docs/src/test/java/docs/http/javadsl/server/PekkoHttp1020MigrationExample.java
@@ -13,7 +13,7 @@ import org.apache.pekko.stream.ActorMaterializer;
 import org.apache.pekko.stream.Materializer;
 
 @SuppressWarnings("deprecation")
-public class AkkaHttp1020MigrationExample {
+public class PekkoHttp1020MigrationExample {
     public static void main(String[] args) {
         {
             //#old-binding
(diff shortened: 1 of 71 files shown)


nvollmar (approved):
lgtm

jrudolph (approved):
lgtm

mdedetrich:
@jrudolph You happy with me merging this now?

docs/src/test/java/docs/http/javadsl/server/PekkoHttp1020MigrationExample.java
```diff
@@ -13,7 +13,7 @@
 import org.apache.pekko.stream.Materializer;
 
 @SuppressWarnings("deprecation")
-public class AkkaHttp1020MigrationExample {
+public class PekkoHttp1020MigrationExample {
```
spangaer, line 16: Just accidentally stumbled over this, but seems like this one would correctly be named Akka, as there is no Pekko 10.2 Migration

pjfanning: I vote for keeping the new name - it easier to explain all the names now have Pekko in them - as opposed to explaining on a case by case basis. One extra option is to just to remove this test.

spangaer: I don't have any firm opinion, but unfirmly: removing it seems like the most right option then.

mdedetrich: Its likely these will be removed as they don't make sense in Pekko, but trademarks (and other legal matters) generally takes precedence over other matters.

PR #50, opened 2023-02-14
fix config in HttpServerLatencyMultiNodeSpec

jrudolph:
Seems to be an oversight from renaming

### DIFF
--- a/http-tests/src/multi-jvm/scala/org/apache/pekko/http/PekkoHttpServerLatencyMultiNodeSpec.scala
+++ b/http-tests/src/multi-jvm/scala/org/apache/pekko/http/PekkoHttpServerLatencyMultiNodeSpec.scala
@@ -36,8 +36,9 @@ object PekkoHttpServerLatencyMultiNodeSpec extends MultiNodeConfig {
         stream.materializer.debug.fuzzing-mode = off
 
         testconductor.barrier-timeout = 30m
-
-        test.AkkaHttpServerLatencySpec {
+      }
+      pekko {
+        test.PekkoHttpServerLatencySpec {
           enable = off
 
           rate = 10000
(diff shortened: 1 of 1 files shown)


http-tests/src/multi-jvm/scala/org/apache/pekko/http/PekkoHttpServerLatencyMultiNodeSpec.scala
```diff
@@ -36,8 +36,9 @@ object AkkaHttpServerLatencyMultiNodeSpec extends MultiNodeConfig {
         stream.materializer.debug.fuzzing-mode = off
 
         testconductor.barrier-timeout = 30m
-
-        test.AkkaHttpServerLatencySpec {
+      }
+      pekko {
```
jrudolph, line 40: This is the fix, the rest is just extra renamings.

mdedetrich:
@jrudolph I already made a PR https://github.com/apache/incubator-pekko-http/pull/48 that renames the classes/functions that have akka in their name, should I merge that first and then you can rebase this PR against it?

http-tests/src/multi-jvm/scala/org/apache/pekko/http/PekkoHttpServerLatencyMultiNodeSpec.scala
```diff
@@ -145,13 +146,13 @@ class AkkaHttpServerLatencyMultiNodeSpec extends MultiNodeSpec(AkkaHttpServerLat
       })
   }
 
-  val enableSpec = system.settings.config.getBoolean("org.apache.pekko.test.AkkaHttpServerLatencySpec.enable")
+  val enableSpec = system.settings.config.getBoolean("pekko.test.PekkoHttpServerLatencySpec.enable")
```
jrudolph, line 149: The problem was that these referenced the config above which had the keys still under `akka` (and also would never move to `org.apache.pekko`).

jrudolph:
Ugh `scalafmt` does not work for multi-jvm tests...

mdedetrich:
> Ugh scalafmt does not work for multi-jvm tests...

I think this is because we are using using https://github.com/apache/incubator-pekko-http/blob/main/.scalafmt.conf#L77 which assumes a default sbt/scala project layout but it can be completely customized, see https://scalameta.org/scalafmt/docs/configuration.html#projectlayout

jrudolph:
> I think this is because we are using using [`main`/.scalafmt.conf#L77](https://github.com/apache/incubator-pekko-http/blob/main/.scalafmt.conf?rgh-link-date=2023-02-14T13%3A12%3A56Z#L77) which assumes a default sbt/scala project layout but it can be completely customized, see [scalameta.org/scalafmt/docs/configuration.html#projectlayout](https://scalameta.org/scalafmt/docs/configuration.html#projectlayout)

No, it's because the sbt plugin works differently than the command line tool. Fix in #51

mdedetrich (approved):
lgtm

pjfanning (approved):
lgtm

PR #52, opened 2023-02-14
disable publishing of http2-support module

jrudolph:
This used to carry the http2 implementation but we moved the implementation to http-core a long time ago. We continued publishing to avoid people missing the module but it doesn't seem to make sense to keep publishing for pekko-http.


mdedetrich:
Should we rename the `http2-support` module to `http2-tests` while we are at it so its more clear?

jrudolph:
Yes, good point.

jrudolph:
Let's do that in the next step.

mdedetrich (commented):
lgtm

PR #56, opened 2023-02-14
publish nightlies only nightly and not for every main commit

jrudolph:
At least until we know that it will not create issues for the apache infrastructure.


pjfanning (approved):
lgtm

PR #65, opened 2023-02-16
Add github PR check to make sure code is formatted

mdedetrich:
Adds a github PR status check which disallows PR's from being merged if the code is not formatted. Note that its been verified that this works in another pekko module (see https://github.com/apache/incubator-pekko-sbt-paradox/blob/main/.asf.yaml#L28-L35)

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -28,7 +28,13 @@ github:
     rebase:  true
 
   protected_branches:
-    main: { }
+    main:
+      required_status_checks:
+        # strict means "Require branches to be up to date before merging".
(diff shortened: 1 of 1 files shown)


jrudolph:
We should also add "Checks" and "Compile and test (2.13, 1.8)".

jrudolph (approved):
lgtm

mdedetrich:
> We should also add "Checks" and "Compile and test (2.13, 1.8)".

Same response as https://github.com/apache/incubator-pekko/pull/199#issuecomment-1432922427

PR #66, opened 2023-02-16
Use incubator-pekko-sbt-paradox

mdedetrich:
I have tested this locally and everything appears to work (i.e. the new material theme is being used).


jrudolph (approved):
lgtm

Issue #62, opened 2023-02-16
Figure out license files wrt Twitter HPACK which we include vendored in http-core


jrudolph:
The copy is shaded but also modified so that using the upstream version is not possible. We could probably extract the java source into its own submodule if that simplifies the license situation.

jrudolph:
In #19 the existing notes in the NOTICE.txt file were deleted. Was that intentional?

jrudolph:
> In #19 the existing notes in the NOTICE.txt file were deleted. Was that intentional?

According to the #19 description it was intentional.

pjfanning:
I am not an expert on licensing and I am not a lawyer.

I've seen what the Incubator mentors and PMC members say about other instances in our code base and in other Incubator projects but when I try to apply what I think is the reasoning to scenarios, I frequently get it wrong.

With the NOTICE file, we have https://infra.apache.org/licensing-howto.html#mod-notice - that we should only modify the standard Apache NOTICE when we have to. It's worth reading that full doc (not just the mod-notice paragraph).

There is this: `If the dependency supplies a NOTICE file, its contents must be analyzed and the relevant portions bubbled up into the top-level NOTICE file.` but I could not find HPACK's NOTICE file - if there is one that I've missed - we need to abide by it in our NOTICE.

With the LICENSE file. we do have the leeway to mention code that is not ours and what the license for that code is. See the incubator-pekko repo LICENSE and the unmerged https://github.com/apache/incubator-pekko/pull/159. https://infra.apache.org/licensing-howto.html#alv2-dep says that we do not normally need to list Apache licensed code in our LICENSE.

jrudolph:
Ok, I'm fine with being relaxed about the notes, it would be good though to some place to list all external bundled dependencies. The document you linked mentions:

> ## Bundling an Apache 2-0-licensed dependency

> Assuming that the bundled dependency itself contains no bundled sub-components under other licenses, so the ALv2 applies uniformly to all files, there is no need to modify LICENSE. However, for completeness it is useful to list the products and their versions, as is done for products under other licenses.

So, it would be nice to keep a note in the LICENSE file about our inclusion of twitter hpack.

pjfanning:
Checking again, the source files have copyrights in their headers.

```
 * Copyright 2014 Twitter, Inc.
```

So it seems that we should mention this in our LICENSE. I can do a PR over the coming week.

pjfanning:
relates to #35

pjfanning:
[...]

PR #67, opened 2023-02-16
include hpack details in license

pjfanning:
Relates to #62

### DIFF
--- a/project/AddMetaInfLicenseFiles.scala
+++ b/project/AddMetaInfLicenseFiles.scala
@@ -17,8 +17,17 @@ import org.mdedetrich.apache.sonatype.SonatypeApachePlugin.autoImport._
  */
 object AddMetaInfLicenseFiles extends AutoPlugin {
 
+  private lazy val baseDir = LocalRootProject / baseDirectory
+
   override lazy val projectSettings = Seq(
+    apacheSonatypeLicenseFile := baseDir.value / "legal" / "StandardLicense.txt",
     apacheSonatypeDisclaimerFile := Some((LocalRootProject / baseDirectory).value / "DISCLAIMER"))
+
+  /**
+   * Settings specific for Pekko http-core subproject which require a different license file.
+   */
+  lazy val httpCoreSettings = Seq(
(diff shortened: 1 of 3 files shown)


project/AddMetaInfLicenseFiles.scala
```diff
@@ -17,9 +17,18 @@ import org.mdedetrich.apache.sonatype.SonatypeApachePlugin.autoImport._
...
+
   override lazy val projectSettings = Seq(
+    apacheSonatypeLicenseFile := baseDir.value / "legal" / "StandardLicense.txt",
     apacheSonatypeDisclaimerFile := Some((LocalRootProject / baseDirectory).value / "DISCLAIMER"))
 
+  /**
+   * Settings specific for Pekko http-core subproject which require a different license file.
+   */
+  lazy val httpCoreSettings = Seq(
+    apacheSonatypeLicenseFile := baseDir.value / "LICENSE")
```
jrudolph, line 30: Does that work to override the `projectSettings`? It's somewhat brittle to define the same setting twice for a project...

jrudolph (approved):
lgtm, have you tested that the LICENSE files end up correctly in the binaries?

project/AddMetaInfLicenseFiles.scala
mdedetrich: This is the correct way of doing it because of how its structured for sbt-apache-sonatype and there is a test for it (see https://github.com/mdedetrich/sbt-apache-sonatype/tree/main/src/sbt-test/sbt-apache-sonatype/multi-project-different-license-test).

sbt-apache-sonatype is designed to set `apacheSonatypeLicenseFile` for every project and then a user is meant to override that value on a per project basis. Unless there is some non-determinism I am not aware of it should work fine (and can be verified with `show apacheSonatypeLicenseFile`.

pjfanning: I'll merge this as is. We won't really know if this is the right solution until the Incubator PMC members look at our initial source releases.

PR #73, opened 2023-02-20
Add copyright headers

mdedetrich:
`CopyrightHeader` has also been updated to make sure the header check is consistently applied.

### DIFF
--- a/project/CopyrightHeader.scala
+++ b/project/CopyrightHeader.scala
@@ -14,7 +14,8 @@
 package akka
 
 import sbt._, Keys._
-import de.heikoseeberger.sbtheader.{ CommentCreator, HeaderPlugin }
+import de.heikoseeberger.sbtheader.{ CommentCreator, HeaderPlugin, NewLine }
+import org.apache.commons.lang3.StringUtils
 
 object CopyrightHeader extends AutoPlugin {
   import HeaderPlugin.autoImport._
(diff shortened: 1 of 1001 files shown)


pjfanning:
not a big deal but could you update the conf files as https://github.com/apache/incubator-pekko-connectors/pull/31 does?

mdedetrich:
@pjfanning Are you sure this necessary? I was under the impression that source files are treated completely separately from configuration files (such as `.properties` or `.conf` files).

I haven't really seen copyright header on configuration files because then you get into problems with people copying and then modifying those configuration files as a base conf

pjfanning (approved):
lgtm - we can look at conf files later

project/CopyrightHeader.scala
```diff
@@ -1,11 +1,21 @@
...
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * license agreements; and to You under the Apache License, version 2.0:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * This file is part of the Apache Pekko project, derived from Akka.
+ */
+
+/*
+ * Copyright (C) 2018-2022 Lightbend Inc. <https://www.lightbend.com>
```
jrudolph, line 11: Was that update intentional?

mdedetrich:
> lgtm - we can look at conf files later

Mailing list discussion started at https://lists.apache.org/thread/9plz30o8g2258jmnb9mym3bfsdfyh9pz

jrudolph (changes requested):
lgtm, there's at least one update of the Lightbend copyright header that might have been accidental?

project/CopyrightHeader.scala
mdedetrich: I have just force pushed the changes, please recheck

jrudolph (approved):
lgtm

PR #74, opened 2023-02-21
make sure that no pekko version specified to build against is same as "default"

jrudolph:
That fixes the nightly tests which only accidentally worked before and when fixing the akka->pekko prefix in f47c735e20d62fa58ce81e43cb4f5d82a1c9272b broke because "default" started to refer to pekko 1.0.0 which is not yet released.

### DIFF
--- a/project/PekkoDependency.scala
+++ b/project/PekkoDependency.scala
@@ -45,9 +45,8 @@ object PekkoDependency {
             // (typically for the docs that require 2.6)
             if (defaultVersion.startsWith("1.0")) Artifact(determineLatestSnapshot("0.0.0"), true)
             else Artifact(defaultVersion)
-          case Some("default") => Artifact(defaultVersion)
-          case Some(other)     => Artifact(other, true)
-          case None            =>
+          case Some(other)            => Artifact(other, true)
+          case Some("default") | None =>
             // TODO: Revert to Artifact(defaultVersion) when release of Pekko is made
             mainSnapshot
         }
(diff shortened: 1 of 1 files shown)


project/PekkoDependency.scala
```diff
@@ -45,9 +45,8 @@ object PekkoDependency {
             // (typically for the docs that require 2.6)
             if (defaultVersion.startsWith("1.0")) Artifact(determineLatestSnapshot("0.0.0"), true)
             else Artifact(defaultVersion)
-          case Some("default") => Artifact(defaultVersion)
-          case Some(other)     => Artifact(other, true)
-          case None            =>
+          case Some(other)            => Artifact(other, true)
+          case Some("default") | None =>
```
mdedetrich, line 49: Can we add a `NOTE` here stating that this should be reverted once a pekko release is made?

mdedetrich (changes requested):
lgtm, would just like to a see a comment so we don't forget.

project/PekkoDependency.scala
jrudolph: It doesn't have to be changed later, these two are actually supposed to be the same.

jrudolph: But actually the ordering of clauses was now wrong... Fixed.

mdedetrich: 👍

mdedetrich (approved):
lgtm

mdedetrich:
Code is not formatted

jrudolph:
:+1: Nightlies green again: https://github.com/apache/incubator-pekko-http/actions/runs/4232112501/jobs/7351447416

PR #78, opened 2023-02-21
Renames headers workflow

mdedetrich:
The `headers.yml` filename is more appropriate (incase we need to add other steps that isn't strictly related to checks).


pjfanning (approved):
lgtm

PR #79, opened 2023-02-22
remove Akka specific release train docs

pjfanning:


jrudolph (approved):
lgtm, I think having a release train issue is quite worthwhile but we will have to redo it

pjfanning:
@jrudolph we'll definitely need new docs but I think it is tidier to recreate them when needed. We can always look back on the Akka docs for inspiration.

PR #80, opened 2023-02-22
use sbt-hackers-digest

jrudolph:
This is a new plugin I'm working on. It will annotate compilation errors, warnings, and test failures on
the relevant files in Github Actions.

I added a test failure to show the result here.

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -36,6 +36,7 @@ addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.10.0")
 addSbtPlugin("com.github.sbt" % "sbt-ci-release" % "1.5.10")
 addSbtPlugin("org.mdedetrich" % "sbt-apache-sonatype" % "0.1.1")
 addSbtPlugin("com.github.pjfanning" % "sbt-source-dist" % "0.1.5")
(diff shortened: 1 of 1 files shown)


jrudolph:
Interesting, our test reporter also reports the tests as annotations.

jrudolph:
Right now the output is a bit verbose on the "Files Changed" tab because we have multiple tools and multiple test runs annotate the same (or mostly the same) lines multiple times. I think in summary it's still better than sifting through failed run logs:

image

For compilation warnings/errors:

image

I'm now removing the test failure commit to un-draft this PR.

jrudolph:
(Test annotations still available as a sample of the output at https://github.com/apache/incubator-pekko-http/commit/0b85f0cf4b635bb7d99885fbd3e955b95908ef07).

mdedetrich:
This is looking really nice! Something that you might want to consider later is that there is merit in putting this sought of plugin into sbt org, I have already done this for https://github.com/sbt/sbt-github-actions (which I maintain).

mdedetrich (approved):
lgtm

mdedetrich:
On another note, I was wondering if you combine this sbt plugin with `addSbtPlugin("com.hpe.sbt" % "sbt-pull-request-validator" % "1.0.0")` (which is used in pekko), would it only report annotations for code that has changed in the PR?

PR #82, opened 2023-02-22
fix some warnings

jrudolph:
These are somewhat genuine as we are using an `AnyRef` field for different value types for performance reasons (and in lack of cheap union types in Scala 2).


mdedetrich (approved):
lgtm

PR #83, opened 2023-02-22
update to Scala 2.13.10

jrudolph:
Recent Scala 2.13 versions deprecated the -target option which then fails the docs module compilation which has warnings as errors.

(Extracted from #81 because it was more complicated than expected)

### DIFF
--- a/docs/src/test/scala/docs/http/scaladsl/server/directives/JsonStreamingExamplesSpec.scala
+++ b/docs/src/test/scala/docs/http/scaladsl/server/directives/JsonStreamingExamplesSpec.scala
@@ -125,7 +125,7 @@ class JsonStreamingExamplesSpec extends RoutingSpec with CompileOnlySpec {
       Marshalling.WithFixedContentType(ContentTypes.`text/csv(UTF-8)`,
         () => {
           val txt = t.txt.replaceAll(",", ".")
-          val uid = t.uid
+          val uid = t.uid.toString
           ByteString(List(uid, txt).mkString(","))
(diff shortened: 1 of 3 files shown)


docs/src/test/scala/docs/http/scaladsl/server/directives/JsonStreamingExamplesSpec.scala
```diff
@@ -125,7 +125,7 @@ class JsonStreamingExamplesSpec extends RoutingSpec with CompileOnlySpec {
       Marshalling.WithFixedContentType(ContentTypes.`text/csv(UTF-8)`,
         () => {
           val txt = t.txt.replaceAll(",", ".")
-          val uid = t.uid
+          val uid = t.uid.toString
```
jrudolph, line 128: This led to a new fatal warning in 2.13.10.

mdedetrich (approved):
lgtm, I am just a bit torn and whether this should target `1.0.0` (i.e. `main`) or `1.1.0`. Original idea was that `1.1.x` was no behaviour changes but I guess with the parboiled2 chjages this concept is kind of thrown out the window

jrudolph:
We don't really expect behavior changes here, it should be less intrusive than all the package renamings associated changes.

mdedetrich:
> We don't really expect behavior changes here, it should be less intrusive than all the package renamings associated changes.

👍 Feel free to merge, also since we are also dealing a http server I think this is less of a concern compared to pekko core

PR #87, opened 2023-02-26
Update naming for status codes 413 and 422

Philippus:
I think the naming changed in RFC 9110, see https://www.rfc-editor.org/rfc/rfc9110.html#name-413-content-too-large and https://www.rfc-editor.org/rfc/rfc9110.html#name-422-unprocessable-content.

### DIFF
--- a/http-core/src/main/java/org/apache/pekko/http/javadsl/model/StatusCodes.java
+++ b/http-core/src/main/java/org/apache/pekko/http/javadsl/model/StatusCodes.java
@@ -97,10 +97,19 @@ public final class StatusCodes {
       org.apache.pekko.http.scaladsl.model.StatusCodes.LengthRequired();
   public static final StatusCode PRECONDITION_FAILED =
       org.apache.pekko.http.scaladsl.model.StatusCodes.PreconditionFailed();
+
+  public static final StatusCode CONTENT_TOO_LARGE =
+      org.apache.pekko.http.scaladsl.model.StatusCodes.ContentTooLarge();
+
+  /** @deprecated deprecated in favor of CONTENT_TOO_LARGE since 1.1.0 */
+  @Deprecated
   public static final StatusCode PAYLOAD_TOO_LARGE =
       org.apache.pekko.http.scaladsl.model.StatusCodes.PayloadTooLarge();
 
-  /** @deprecated deprecated in favor of PAYLOAD_TOO_LARGE */
+  /**
+   * @deprecated deprecated in favor of PAYLOAD_TOO_LARGE, which was later deprecated in favor of
+   *     CONTENT_TOO_LARGE (please switch to CONTENT_TOO_LARGE)
+   */
   @Deprecated
   public static final StatusCode REQUEST_ENTITY_TOO_LARGE =
       org.apache.pekko.http.scaladsl.model.StatusCodes.RequestEntityTooLarge();
@@ -131,8 +140,14 @@ public final class StatusCodes {
       org.apache.pekko.http.scaladsl.model.StatusCodes.EnhanceYourCalm();
   public static final StatusCode MISDIRECTED_REQUEST =
       org.apache.pekko.http.scaladsl.model.StatusCodes.MisdirectedRequest();
+  public static final StatusCode UNPROCESSABLE_CONTENT =
+      org.apache.pekko.http.scaladsl.model.StatusCodes.UnprocessableContent();
+
+  /** @deprecated deprecated in favor of UNPROCESSABLE_CONTENT since 1.1.0 */
+  @Deprecated
   public static final StatusCode UNPROCESSABLE_ENTITY =
       org.apache.pekko.http.scaladsl.model.StatusCodes.UnprocessableEntity();
+
   public static final StatusCode LOCKED = org.apache.pekko.http.scaladsl.model.StatusCodes.Locked();
   public static final StatusCode FAILED_DEPENDENCY =
       org.apache.pekko.http.scaladsl.model.StatusCodes.FailedDependency();
--- a/http-core/src/main/scala/org/apache/pekko/http/scaladsl/model/StatusCode.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/scaladsl/model/StatusCode.scala
@@ -151,9 +151,11 @@ object StatusCodes extends ObjectRegistry[Int, StatusCode] {
   val Gone                         = reg(c(410)("Gone", "The resource requested is no longer available and will not be available again."))
   val LengthRequired               = reg(c(411)("Length Required", "The request did not specify the length of its content, which is required by the requested resource."))
   val PreconditionFailed           = reg(c(412)("Precondition Failed", "The server does not meet one of the preconditions that the requester put on the request."))
-  val PayloadTooLarge              = reg(c(413)("Payload Too Large", "The request payload is larger than the server is willing or able to process."))
-  @deprecated("deprecated in favor of PayloadTooLarge", "Akka HTTP 10.1.11")
-  val RequestEntityTooLarge        = PayloadTooLarge
+  val ContentTooLarge              = reg(c(413)("Content Too Large", "The request content is larger than the server is willing or able to process."))
+  @deprecated("deprecated in favor of ContentTooLarge", "1.1.0")
+  val PayloadTooLarge              = ContentTooLarge
+  @deprecated("deprecated in favor of PayloadTooLarge, which was later deprecated in favor of ContentTooLarge (please switch to ContentTooLarge)", "Akka HTTP 10.1.11")
+  val RequestEntityTooLarge        = ContentTooLarge
   val UriTooLong                   = reg(c(414)("URI Too Long", "The URI provided was too long for the server to process."))
   @deprecated("deprecated in favor of UriTooLong", "Akka HTTP 10.1.11")
   val RequestUriTooLong            = UriTooLong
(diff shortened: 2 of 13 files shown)


pjfanning:
Are there any tests maintained for the legacy status codes? We need to prove that users can still respond with old status code variant.

http-core/src/main/java/org/apache/pekko/http/javadsl/model/StatusCodes.java
```diff
@@ -66,10 +66,16 @@ private StatusCodes() {}
     public static final StatusCode GONE = org.apache.pekko.http.scaladsl.model.StatusCodes.Gone();
     public static final StatusCode LENGTH_REQUIRED = org.apache.pekko.http.scaladsl.model.StatusCodes.LengthRequired();
     public static final StatusCode PRECONDITION_FAILED = org.apache.pekko.http.scaladsl.model.StatusCodes.PreconditionFailed();
+    public static final StatusCode CONTENT_TOO_LARGE = org.apache.pekko.http.scaladsl.model.StatusCodes.ContentTooLarge();
+
+    /**
+     * @deprecated deprecated in favor of CONTENT_TOO_LARGE
```
pjfanning, line 72: Could you include `since = "1.0.0"` on all new deprecations?

Philippus: not sure how to do this in the java comment. just add " since 1.0.0" to the comment?

pjfanning: probably best to read https://www.baeldung.com/javadoc-version-since

pjfanning: https://www.scala-lang.org/api/2.13.10/scala/deprecated.html describes how to set the 'since value' in Scala deprecations

jrudolph:
Nice addition, thanks @Philippus.

Is that something we would consider for 1.0.0? There's little risk but currently I'd say let's postpone to 1.1.x on principle since it is not strictly necessary right now?

pjfanning:
there is no hurry on this as we are unlikely to include this in the pekko-http 1.0.0 release (and add it later instead)

mdedetrich:
Indeed take your time, its not urgent

Philippus:
I'll fix it up once the first pekko-http is out of the door.

He-Pin:
@Philippus You need rebase your PR.

http-core/src/main/scala/org/apache/pekko/http/scaladsl/model/StatusCode.scala
```diff
@@ -151,8 +151,10 @@ object StatusCodes extends ObjectRegistry[Int, StatusCode] {
   val Gone                         = reg(c(410)("Gone", "The resource requested is no longer available and will not be available again."))
   val LengthRequired               = reg(c(411)("Length Required", "The request did not specify the length of its content, which is required by the requested resource."))
   val PreconditionFailed           = reg(c(412)("Precondition Failed", "The server does not meet one of the preconditions that the requester put on the request."))
+  val ContentTooLarge              = reg(c(413)("Content Too Large", "The request content is larger than the server is willing or able to process."))
+  @deprecated("deprecated in favor of ContentTooLarge", "1.0.0")
   val PayloadTooLarge              = reg(c(413)("Payload Too Large", "The request payload is larger than the server is willing or able to process."))
-  @deprecated("deprecated in favor of PayloadTooLarge", "Akka HTTP 10.1.11")
+  @deprecated("deprecated in favor of ContentTooLarge", "1.0.0")
   val RequestEntityTooLarge        = PayloadTooLarge
```
pjfanning, line 158: can you change this to this
```
@deprecated("deprecated in favor of PayloadTooLarge, which was later deprecated in favor of ContentTooLarge (please switch to ContentTooLarge)", "Akka HTTP 10.1.11")
```


* can you a comment - `// RequestEntityTooLarge was originally deprecated in favour of PayloadTooLarge in Akka HTTP 10.1.11` ?
* This will remind us that this line can actually be removed soon - possibly in Pekko 1.1.0 - because this line was already deprecated - but it does make sense to adjust the deprecation message now

pjfanning (approved):
lgtm

jrudolph (approved):
lgtm

pjfanning (changes requested):
Apologies - I was going around approving PRs but I forgot about the Scala 3 compile issue in this PR - hopefully, we;ll get around to fixing that soon

mdedetrich:
Was about to say that the tests are failing with `java.lang.NoClassDefFoundError: Could not initialize class org.apache.pekko.http.scaladsl.model.StatusCodes$`

jrudolph:
> Are there any tests maintained for the legacy status codes? We need to prove that users can still respond with old status code variant.

This is only an API change and for those changes we mostly rely on Mima to catch problems. We also have the (currently disabled) "http-compatibility-tests" module which has logic to test backward compatibility in cases which warrant more testing.

http-core/src/main/scala/org/apache/pekko/http/scaladsl/model/StatusCode.scala
```diff
@@ -151,8 +151,10 @@ object StatusCodes extends ObjectRegistry[Int, StatusCode] {
   val Gone                         = reg(c(410)("Gone", "The resource requested is no longer available and will not be available again."))
   val LengthRequired               = reg(c(411)("Length Required", "The request did not specify the length of its content, which is required by the requested resource."))
   val PreconditionFailed           = reg(c(412)("Precondition Failed", "The server does not meet one of the preconditions that the requester put on the request."))
+  val ContentTooLarge              = reg(c(413)("Content Too Large", "The request content is larger than the server is willing or able to process."))
+  @deprecated("deprecated in favor of ContentTooLarge", "1.0.0")
   val PayloadTooLarge              = reg(c(413)("Payload Too Large", "The request payload is larger than the server is willing or able to process."))
```
jrudolph, line 156:
```suggestion
  val PayloadTooLarge              = ContentTooLarge
```

http-core/src/main/scala/org/apache/pekko/http/scaladsl/model/StatusCode.scala
```diff
@@ -151,8 +151,10 @@ object StatusCodes extends ObjectRegistry[Int, StatusCode] {
...
-  @deprecated("deprecated in favor of PayloadTooLarge", "Akka HTTP 10.1.11")
+  @deprecated("deprecated in favor of PayloadTooLarge, which was later deprecated in favor of ContentTooLarge (please switch to ContentTooLarge)", "Akka HTTP 10.1.11")
   val RequestEntityTooLarge        = PayloadTooLarge
```
jrudolph, line 158:
```suggestion
  val RequestEntityTooLarge        = ContentTooLarge
```

jrudolph (approved):
Needs the below fixes to fix the runtime errors.

http-core/src/main/scala/org/apache/pekko/http/scaladsl/model/StatusCode.scala
```diff
@@ -165,6 +167,8 @@ object StatusCodes extends ObjectRegistry[Int, StatusCode] {
   val ImATeapot                    = reg(c(418)("I'm a teapot", "The resulting entity body MAY be short and stout."))
   val EnhanceYourCalm              = reg(c(420)("Enhance Your Calm", "You are being rate-limited.")) // Twitter only
   val MisdirectedRequest           = reg(c(421)("Misdirected Request", "The request was directed at a server that is not able to produce a response.")) // HTTP/2 only. https://tools.ietf.org/html/rfc7540#section-9.1.2
+  val UnprocessableContent         = reg(c(422)("Unprocessable Content", "The request was well-formed but was unable to be followed due to semantic errors."))
+  @deprecated("deprecated in favor of UnprocessableContent", "1.0.0")
   val UnprocessableEntity          = reg(c(422)("Unprocessable Entity", "The request was well-formed but was unable to be followed due to semantic errors."))
```
jrudolph, line 172: Same here:

```suggestion
  val UnprocessableEntity          = UnprocessableContent
```

http-core/src/main/java/org/apache/pekko/http/javadsl/model/StatusCodes.java
```diff
@@ -97,10 +97,19 @@ private StatusCodes() {}
       org.apache.pekko.http.scaladsl.model.StatusCodes.LengthRequired();
   public static final StatusCode PRECONDITION_FAILED =
       org.apache.pekko.http.scaladsl.model.StatusCodes.PreconditionFailed();
+
+  public static final StatusCode CONTENT_TOO_LARGE =
+      org.apache.pekko.http.scaladsl.model.StatusCodes.ContentTooLarge();
+
+  /** @deprecated deprecated in favor of CONTENT_TOO_LARGE since 1.0.0 */
```
pjfanning, line 104: this should be `since 1.1.0` - 1.0.0 was released without this change

http-core/src/main/scala/org/apache/pekko/http/scaladsl/model/StatusCode.scala
```diff
@@ -151,9 +151,11 @@ object StatusCodes extends ObjectRegistry[Int, StatusCode] {
   val Gone                         = reg(c(410)("Gone", "The resource requested is no longer available and will not be available again."))
   val LengthRequired               = reg(c(411)("Length Required", "The request did not specify the length of its content, which is required by the requested resource."))
   val PreconditionFailed           = reg(c(412)("Precondition Failed", "The server does not meet one of the preconditions that the requester put on the request."))
-  val PayloadTooLarge              = reg(c(413)("Payload Too Large", "The request payload is larger than the server is willing or able to process."))
-  @deprecated("deprecated in favor of PayloadTooLarge", "Akka HTTP 10.1.11")
-  val RequestEntityTooLarge        = PayloadTooLarge
+  val ContentTooLarge              = reg(c(413)("Content Too Large", "The request content is larger than the server is willing or able to process."))
+  @deprecated("deprecated in favor of ContentTooLarge", "1.0.0")
```
pjfanning, line 155: this should be `"1.1.0"` - 1.0.0 was released without this change

pjfanning (changes requested):

http-core/src/main/scala/org/apache/pekko/http/scaladsl/model/StatusCode.scala
pjfanning: all the refs in this PR to deprecations since `1.0.0` should be changed to `1.1.0`

http-core/src/main/java/org/apache/pekko/http/javadsl/model/StatusCodes.java
jrudolph: :+1: good point

jrudolph:
(btw. test error is likely unrelated)

He-Pin (approved):
lgtm

pjfanning (approved):
lgtm

jrudolph:
Thanks again, @Philippus!

Issue #49, opened 2023-02-14
Replace http requests to akka website with our own


mdedetrich:
A fair few tests in Pekko make http requests against Lightbend/Akka related sites (i.e. https://akka.io/), we should change these tests to make requests against our own site (likely needs to be done after we actually setup a site).

mdedetrich:
@raboof Do you know if there would be any security implications with this?

jrudolph:
Fixed in #90

PR #90, opened 2023-02-28
replace usages of akka.io in tests

jrudolph:
Refs #49

### DIFF
--- a/docs/src/test/java/docs/http/javadsl/ClientSingleRequestExample.java
+++ b/docs/src/test/java/docs/http/javadsl/ClientSingleRequestExample.java
@@ -35,7 +35,7 @@ public class ClientSingleRequestExample {
 
     final CompletionStage<HttpResponse> responseFuture =
       Http.get(system)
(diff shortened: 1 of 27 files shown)


mdedetrich (approved):
lgtm

mdedetrich:
So some tests are failing, I suspect that the tests may have been relying on behaviour from akka.io

jrudolph:
> So some tests are failing, I suspect that the tests may have been relying on behaviour from akka.io

Probably only the expected error messages need to be fixed

mdedetrich:
Might be excessive and definitely something later down the track, but there may be merit in using something like https://dev.to/iurysza/setting-up-mockserver-3kbe

jrudolph:
> Might be excessive and definitely something later down the track, but there may be merit in using something like [dev.to/iurysza/setting-up-mockserver-3kbe](https://dev.to/iurysza/setting-up-mockserver-3kbe)

I'm not even sure if any of the tests even actually run. Most are either using the URL as an example string or showing usages in the docs.

PR #95, opened 2023-03-01
stray akka ref in conf

pjfanning:
A lot of the other ones seem to be resolved by in-flight PRs. I couldn't find one that handles this case.


nvollmar (approved):
lgtm

PR #96, opened 2023-03-01
Use setup-java in Github Actions

mdedetrich:
`setup-scala` is deprecated so lets use `setup-java` instead


nvollmar (approved):
lgtm

.github/workflows/nightly.yml
```diff
@@ -23,11 +23,10 @@ jobs:
           fetch-depth: 0
 
       - name: Set up JDK ${{ matrix.JABBA_JDK }}
-        uses: olafurpg/setup-scala@v10
-        env:
-          JABBA_INDEX: 'https://github.com/typelevel/jdk-index/raw/main/index.json'
```
mdedetrich, line 28: So I don't know why we were using a custom index, I assume its because at some point the JDK lagged behind but I don't think this is an issue anymore?

jrudolph: I don't know or cannot remember, in any case, it shouldn't matter.

jrudolph (approved):
lgtm, this is great, in particular, because `setup-java` knows how to use the Java instances that are already part of the base images, so the setup task does not even have to download anything.

jrudolph:
Reverting since this broke PR validation.

PR #108, opened 2023-03-06
regenerate SSL certificates with new example domain name

jrudolph:


http-core/src/test/scala/org/apache/pekko/http/impl/engine/client/TlsEndpointVerificationSpec.scala
```diff
@@ -146,7 +146,8 @@ class TlsEndpointVerificationSpec extends PekkoSpecWithMaterializer("""
     val handler: HttpRequest => HttpResponse = { req =>
       // verify Tls-Session-Info header information
       val name = req.attribute(AttributeKeys.sslSession).flatMap(_.localPrincipal).map(_.getName)
-      if (name.exists(_ == "CN=akka.example.org,O=Internet Widgits Pty Ltd,ST=Some-State,C=AU")) HttpResponse()
+      println(name)
```
nvollmar, line 149: remove println

nvollmar (approved):
lgtm

http-core/src/test/scala/org/apache/pekko/http/impl/engine/client/TlsEndpointVerificationSpec.scala
jrudolph: Thanks, good catch!

PR #109, opened 2023-03-06
Revert "Use setup-java in Github Actions"

jrudolph:
Reverts apache/incubator-pekko-http#96

This broke the GHA workflows probably because JVM settings are not applied any more.


jrudolph:
It probably did not run any tests because the PR did not touch any code.

mdedetrich (approved):
lgtm, lets make an issue to investigate. Ideally every one of our pekko projects should be using `setup-java` eventually.

PR #111, opened 2023-03-06
run full validation when github workflows change

jrudolph:
To make sure that changes to the validation task really run all tests.


project/ValidatePullRequest.scala
```diff
@@ -111,7 +111,7 @@ object ValidatePullRequest extends AutoPlugin {
   val additionalTasks = settingKey[Seq[TaskKey[_]]]("Additional tasks for pull request validation")
 
   // The set of (top-level) files or directories to watch for build changes.
-  val BuildFilesAndDirectories = Set("project", "build.sbt", ".github/workflows")
```
jrudolph, line 114: The old syntax did not work since the diff set might just contain `.github` in the case that a workflow changed.

jrudolph:
Merging so we can test #113 properly.

Issue #28, opened 2023-01-21
new g8 template (pekko-http-quickstart-scala)


pjfanning:
introduction.md mentions
```
sbt new akka/pekko-http-quickstart-scala.g8
```

luksow:
Probably not very helpful but I converted my OSS template (quite popular, by the way) and the transition was really smooth: https://github.com/theiterators/akka-http-microservice/pull/171

pjfanning:
@luksow I would be open to adding a link to your template. It would allow us to long finger the task of replacing akka-http-quickstart-scala and concentrate on other tasks. When you have time, would you be able to create a pekko-http-microservice from your existing PR? We could mention that in our docs.

luksow:
@pjfanning thanks for getting back to me! Actually, if `pekko-http-quickstart-scala` is endgame goal, I could devote some time to fork & convert it, as it shouldn't be too hard. Is anyone working on it already? Is Java version required?

Another question is, do we want to have a 1:1 version? What I dislike about `akka-http-quickstart` is that it mixes actors with akka http which is very often a bad idea - overkill for most of the use cases.

Also, I'm happy to create pekko-http-microservice from my PR. There are two things to consider. First of all, it's Scala 3 version, which means I have to use `for3Use2_13` which is fine but it also requires me to use circe, instead of spray-json (which has poor/no Scala 3 support). This means I need https://github.com/hseeberger/akka-http-json equivalent. From what I have learnt so far, it's no longer maintained and thus no support for Pekko is planned (https://github.com/hseeberger/akka-http-json/issues/780). So we have two options here:
1. I create pekko-http-microservice using Scala 2, spray-json.
2. I'll create & maintain a fork of `akka-http-json` for Pekko (which I think will be very important at some point) and stick to Scala 3.
Which approach is better in your opinion?

Lastly, having said that, looking forward to Scala 3, I think it's important to consider the role of spray-json in the core of pekko-http as it's basically obsolete.

pjfanning:
Thanks @luksow - any direction that you choose will be useful. A Scala pekko version of akka-quickstart will be easier to integrate with the pekko docs but we can work with anything. Regarding JSON support, pekko-http-spray-json may be ok for now but cloning Heiko Seeburger's libs may be a good longer term plan. I plan to look at this latter option but if anyone else gets there first - that's great.

pjfanning:
@luksow I forked akka-http-json and created https://github.com/pjfanning/pekko-http-json

[...]

PR #112, opened 2023-03-06
reference the pekko g8 templates

pjfanning:
relates to #28

### DIFF
--- a/docs/src/main/paradox/introduction.md
+++ b/docs/src/main/paradox/introduction.md
@@ -54,16 +54,14 @@ For Scala (sbt)
 @@@ div { .group-java }
 For Java (Maven or Gradle)
 :  ```sh
-    sbt new akka/akka-http-quickstart-java.g8
+    sbt new theiterators/pekko-http-quickstart-java.g8
     ```
 From there on the prepared project can be built using Gradle or Maven.
-Unfortunately, we do not yet have a Java template but it should be easy to migrate an Akka project to use
-Pekko instead.
 @@@
 
 More instructions can be found on the @scala[[template
 project](https://github.com/theiterators/pekko-http-quickstart-scala.g8)]@java[[template
-project](https://github.com/akka/akka-http-quickstart-java.g8)].
+project](https://github.com/theiterators/pekko-http-quickstart-java.g8)].
 
 ## Routing DSL for HTTP servers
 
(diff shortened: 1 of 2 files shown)


mdedetrich (approved):
lgtm, is there a plan to bring these templates into apache/pekko org at some point?

jrudolph (commented):
Do we plan to provide these ourselves? Since those are displayed so prominently, it's somewhat risky to directly delegate to a third-party repo which we do not have any control over. This is, of course, no concrete concern towards Olsza or theiterators, whose work on getting the templates migrated is much appreciated, more being generally careful towards potential supply chain attack vectors.

pjfanning:
We'll need code grants etc. My preference is to foster a community around Pekko and not to have everything in Apache repos. If anyone we link starts doing things we don't like, we can remove the links.

jrudolph:
> My preference is to foster a community around Pekko and not to have everything in Apache repos

Yes, of course, but then we have to mark it in the documentation and not make it the main thing everyone will do who starts with pekko.

> We'll need code grants etc.

No, I don't think we have to do anything but fork the project as suggested by Olsza. The original akka version and also this version is put into public domain, so there's nothing to do here.

> Imagine having to do an Apache release every time we make small mods to the g8 templates. Votes in PPMC and in Incubator PMC.

There's no release necessary since the command will pull directly from Github.

pjfanning:
* We do need code grants to accept new git repos. 
* We do have to follow Apache release processes - even if they may not be tailored to our needs in all cases.
* I'm dreading getting to the stage where we need to start releasing our existing pekko repos. All the votes will seriously fill up our time and the so much Incubator PMC time.

How about we remove the g8 documentation from core documentation and just put links in our 'Examples and Extensions' page? We can put a health warning on the top of that page to warn people that the external pages have been reviewed at some stage but they may have changed after they have been reviewed.

mdedetrich:
> How about we remove the g8 documentation from core documentation and just put links in our 'Examples and Extensions' page? We can put a health warning on the top of that page to warn people that the external pages have been reviewed at some stage but they may have changed after they have been reviewed.

Im fine with this

pjfanning:
@jrudolph would you be happy if I make the change to move the g8 templates references to the 'Examples and Extensions' page?

pjfanning:
I'm going to merge this as is because there is an existing ref to TheIterators g8 template (for Scala). I'll raise an issue to revisit.

pjfanning:
raised https://github.com/apache/incubator-pekko-http/issues/124

PR #116, opened 2023-03-06
move to pekko in more various places

jrudolph:


docs/src/test/java/docs/http/javadsl/Http2ClientApp.java
```diff
@@ -58,7 +59,7 @@ public static void main(String[] args) {
 
     dispatch.apply(
         HttpRequest.create(
-            "https://pekko.apache.org/api/akka/current/akka/actor/typed/scaladsl/index.html").withHeaders(
+            "https://pekko.apache.org/api/pekko/current/org/apache/pekko/actor/typed/scaladsl/index.html").withHeaders(
```
jrudolph, line 62: These links are currently arbitrary and the example may fail if actually run but that wasn't guaranteed before as well.

mdedetrich (approved):
lgtm

PR #118, opened 2023-03-07
avoid making the workspace dirty by setting up .jvmpopts

jrudolph:
Otherwise, the version number will contain a timestamp because of the dirty workspace.


mdedetrich (approved):
lgtm. Should do this change in other workshops for both consistency reasons and also if they happen to be using a similar pull request validator mechanism.

PR #129, opened 2023-03-27
build: cleanup PekkoDependency and also build nightly against latest snapshot

jrudolph:


mdedetrich:
So I understand the argument of re-enabling builds using the latest snapshot but this is actually what caused the issue, due to some dependency resolution for pekko-projection it ended up resolving multiple versions of akka/akka-streams which is not allowed.

That being said, as long as there is one version of pekko-http that we know has a dependency with a specific version of akka then its okay

mdedetrich:
Ill approve PR once I see that the matrix works out as expected

jrudolph:
> So I understand the argument of re-enabling builds using the latest snapshot but this is actually what caused the issue, due to some dependency resolution for pekko-projection it ended up resolving multiple versions of akka/akka-streams which is not allowed (see [apache/incubator-pekko-projection/actions/runs/4518661785/jobs/7958676851?pr=19](https://github.com/apache/incubator-pekko-projection/actions/runs/4518661785/jobs/7958676851?pr=19))
> 
> That being said, as long as there is one version of pekko-http that we know has a dependency with a specific version of pekko core then its okay, right now I am trying to make sure that every module is built against pekko `0.0.0+26623-85c2a469-SNAPSHOT` specifically.

pekko-http still declares all pekko core dependencies as `provided`, so it is somewhat unlikely that an issue in a downstream project has anything to do with whatever we use here.

mdedetrich:
> pekko-http still declares all pekko core dependencies as provided, so it is somewhat unlikely that an issue in a downstream project has anything to do with whatever we use here.

I don't know precisely how it happened, personally I did this approach because overall its faster and I don't want to change the projection build just to make it work with snapshots (because our situation with snapshots is temporary).

mdedetrich (approved):
lgtm. I just double checked locally by checking out this branch, doing `publishLocal` to make sure that the deployed artifact still has `0.0.0+26623-85c2a469-SNAPSHOT` as its dependency and everything seems to work out.

Merging.

PR #135, opened 2023-04-14
missing license headers

pjfanning:
some files I found without license headers


He-Pin (approved):
lgtm，so the header checking action missed those？

pjfanning:
they were missed and I'm not sure why - I reran `sbt +headerCheckAll` and these files were not highlighted as issues

PR #137, opened 2023-04-17
some pekko naming issues

pjfanning:


build.sbt
```diff
@@ -405,11 +405,11 @@ lazy val docs = project("docs")
       "scaladoc.org.apache.pekko.base_url" -> s"https://pekko.apache.org/api/pekko/${PekkoDependency.docs.link}",
       "scaladoc.org.apache.pekko.link_style" -> "direct",
       "javadoc.org.apache.pekko.http.base_url" -> {
-        val v = if (!isSnapshot.value) "current" else version.value
+        val v = if (isSnapshot.value) "current" else version.value
```
jrudolph, line 408: Did we discuss changing the naming/URL scheme somewhere? The rationale behind the existing scheme is to have a single URL `/current` (in addition to the versioned ones) which will always point to the latest version of the documentation.

pjfanning: The boolean expression seems wrong. Builds for snapshots are getting the full snapshot version string here. Everything else for snapshot builds uses 'current'. I agree that for non-snapshot builds, having the version number makes sense - but that's not what this expression is doing.

jrudolph: We could choose to remove the condition and link to the versioned variant in all cases. It does not really make sense to use `/current` for snapshots in particular because `current` at least used to mean "latest released version".

Using `/current` for all properly released version (`!isSnapshot`) has the benefit that when people use this path and link to it, they will always get the most recent version of the scaladoc/documentation. The disadvantage would be that people actually might want to link to an older version of the scaladoc/javadoc but with a link under `/current` get an URL that will automatically be updated to the latest version when a new version comes out. 

In fact, it's a disadvantage of linking to versioned/non-`/current` URLs that there's some implied assumption that we will keep an archive of old documentation/javadoc/scaladoc alive basically for ever. This is not completely unreasonable but as we have discussed before the documentation/scaladoc/javadoc archive can quickly get to a non-negligible size and keeping the archive available can become a burden. By linking into the archive ourselves we might make our lives harder than necessary when we want to clean up later because it will be hard to understand where we placed links to the archive and where we might end up with dead links (or need manual redirects). This can become a SEO issue not only because of potential downranking because of dead links but also because of keeping around (and linking to) an archive of mostly duplicate documentation content will mislead search engines and bring users to the wrong versions of documentation pages. It's important, therefore, to exclude the archive from search engine crawlers.

I think there was some confusion about what `/current` meant in Akka and what we might think it should mean for pekko and I might have missed discussions about how we think it should be used.

jrudolph: > The boolean expression seems wrong. Builds for snapshots are getting the full snapshot version string here. Everything else for snapshot builds uses 'current'. I agree that for non-snapshot builds, having the version number makes sense - but that's not what this expression is doing.

To answer that directly: it might seem wrong but it is as it was intended for the above reasons.

The reasoning for using versions in snapshots version explicitly is because snapshot documentation is not for broad consumption but for explicit checking of the documentation state of the `main` branch. In that particular case, we want to link to the exact version of scaladoc/javadoc not least because we might have written documentation that refers to new classes that are only available on the `main` branch so linking to `/current` will link to an *older* stable version that does not represent the state of the `main` branch.

(You could make the same argument for stable releases, as well, i.e. that linking to versioned javadoc/scaladoc will lead to a more consistent view of documentation at the point of when a release was made. However, there's a trade-off regarding the above reasons and the above problems with search engine and we decided for Akka that having simple and up-to-date documentation and links that work well with search engines trumps versioned links. Its amazing how many people would end up on years-old documentation because of the search engines couldn't figure where to link for particular content...)

pjfanning:
I think I've removed anything that was contentious.

PR #146, opened 2023-05-12
Regenerate documentation images using pekko implementations.

### LINKED ISSUE #99: Regenerate images in documentation which have akka references
Can't remember precisely where but there are images in the current pekko Paradox docs which refer to Akka. Some examples I can remember are a visalized flamegraph and also a screenshot of a http webserver that still has Akka in it.

We should regenerate the screenshots so that they have Pekko naming instead and update the docs to reflect this.

Math-ias:
Hoping to help close #99 .

Using this command (src https://stackoverflow.com/a/24879385) I found all the images in the repo ...
```
find . -type f -exec file --mime-type {} \+ | awk -F: '{if ($2 ~/image\//) print $1}'
```

1. DispatcherBehaviourOnBadCode.png & DispatcherBehaviourOnGoodCode.png

These are used in `handling-blocking-operations-in-pekko-http-routes.md` to document the effect of blocking operations.

To regenerate these ...
 * I created a sample project using pekko libraries.
 * I used the routes documented in the article.
 * I concatenated a `get` route to allow for traffic with GET requests.
 * I used the following script trying to follow along with the traffic the article was describing ...
```
for i in $(seq 256);
do
  sleep 0.05
  curl http://localhost:8080 &
done
wait
for i in $(seq 256);
do
  sleep 0.05
  curl -X POST http://localhost:8080 &
  sleep 0.05
  curl http://localhost:8080 &
done
```
 * I started a recording from Java Mission Control.
 * I viewed that recording in the Legacy Threads view.
 * I clicked on the default or blocking dispatcher threads.
 * I stored and set that selection.
 * I then went into the new Threads view and zoomed in.

I changed the language in the doc to match my recording setup. That language includes mention of color and the tools used.

In trying out the code of the article I found I needed changes ... I've estimated what the proper fix is for the Java example, though I haven't checked it in my example app.

```
[error] /Users/[...]]/Desktop/pekko-example/src/main/scala/com/example/QuickstartApp.scala:20:65: type mismatch;
[error]  found   : String("my-blocking-dispatcher")
[error]  required: org.apache.pekko.actor.typed.DispatcherSelector
```

This code was committed over 7 years ago
4ae775007f29ad2a13da37e7150e7d665087583e
and I anticipate the API has changed since then.

2. DispatcherBehaviourProperBlocking.png

This image isn't in use today ...

```
% git grep DispatcherBehaviourProperBlocking.png
```

And has never been used ...

```
% git log -S "DispatcherBehaviourProperBlocking"
```

Not even when the file was first checked in
4ae775007f29ad2a13da37e7150e7d665087583e.

3. akka-http-file-listing.png

This one was easy enough.
I added a route to return `getFromBrowsableDirectory`. I pointed that to some example directory I created, and took a screenshot from Safari.


docs/src/test/java/docs/http/javadsl/server/BlockingInHttpExamples.java
```diff
@@ -60,7 +61,7 @@ void blockingHttpDedicatedDispatcher() {
         //#blocking-example-in-dedicated-dispatcher
         // GOOD (the blocking is now isolated onto a dedicated dispatcher):
         final Route routes = post(() -> {
-            final MessageDispatcher dispatcher = system.dispatchers().lookup("my-blocking-dispatcher");
```
pjfanning, line 63: the existing code compiles for me - why change it?

docs/src/main/paradox/handling-blocking-operations-in-pekko-http-routes.md
```diff
@@ -63,8 +65,8 @@ provided dispatcher would be used. Both these dispatchers are ForkJoin pools by
 not best suited for blocking operations.
 @@@
 For example, the above screenshot shows an Apache Pekko FJP dispatchers threads,
-named "`default-akka.default-dispatcher2,3,4`" going into the blocking state, after having been idle. 
-It can be observed that the number of new threads increases, "`default-akka.actor.default-dispatcher 18,19,20,...`" 
+named "`m-pekko.actor.default-dispatcher-6,7`" going into the blocking state, after having been idle.
```
pjfanning, line 68: why 'm-pekko'? - there is also an 'default-akka' on the next like - should these not both become 'default-pekko'?

docs/src/main/paradox/akka-http-file-listing.png
pjfanning, line 1: could this be renamed to pekko-http-file-listing.png ?

Math-ias: Good catch. I can move this and change all the references.

docs/src/test/java/docs/http/javadsl/server/BlockingInHttpExamples.java
Math-ias: I originally thought changes were necessary because I couldn't paste this code into a sample project. On closer inspection (and realizing this file does compile 🤦) I found my original mistake. In my sample project I was using the `typed` actor system which does have API changes. This is also the source of the odd `m` prefix in dispatcher names: the typed ActorSystem apply method demands a name. I will revert these changes and regenerate screenshots using the classic actor system.

docs/src/main/paradox/handling-blocking-operations-in-pekko-http-routes.md
Math-ias: Good catch on the mismatched names. See other comment about why `m-pekko` originally: I was using the typed actor system. Will fix shortly.

Math-ias:
Force pushed to change my email on the commits.

pjfanning:
there is a merge conflict, I'm afraid

Math-ias:
Performed a merge back with main to resolve these conflicts. Required some manual intervention.

.asf.yaml
```diff
@@ -13,18 +13,18 @@ github:
...
-
+  
   features:
     # Enable wiki for documentation
     wiki: false
     # Enable issue management
     issues: true
     # Enable projects for project management boards
     projects: true
-
```
pjfanning: please revert this change - this file has nothing to do with the doc changes

pjfanning: the whole file should left as is

.git-blame-ignore-revs
```diff
@@ -1,2 +1,8 @@
 # scalafmt
 90f8bf7e6d6f7a2d6c63313944613a95ff60091a
+
+# javafmt
```
pjfanning: revert these file changes - the Pekko contributors will decide what goes in here

pjfanning (approved):
lgtm - thanks - merging

PR #150, opened 2023-05-24
Scala 3 support

mdedetrich:
Had to make a new branch because I don't have permission to push onto @jrudolph s branch.


.asf.yaml
```diff
@@ -35,11 +35,6 @@ github:
         # contexts are the names of checks that must pass
         contexts:
           - Code is formatted
-          - Check headers
-      required_pull_request_reviews:
```
pjfanning, line 39: We probably don't want to remove this

.github/workflows/nightly.yml
```diff
@@ -18,7 +18,7 @@ jobs:
         PEKKO_VERSION: [default, main]
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v2
```
pjfanning, line 21: can we avoid this revert?

build.sbt
```diff
@@ -11,7 +11,7 @@ import sbtdynver.GitDescribeOutput
 import spray.boilerplate.BoilerplatePlugin
 import com.lightbend.paradox.apidoc.ApidocPlugin.autoImport.apidocRootPackage
 
-sourceDistName := "incubating-pekko-http"
+sourceDistName := "incubator-pekko-http"
```
pjfanning, line 14: we can't do this revert

pjfanning (dismissed):

mdedetrich:
Okay I need to clean this up, evidently something happened

http-core/src/main/scala/org/apache/pekko/http/impl/model/parser/SimpleHeaders.scala
```diff
@@ -111,14 +112,14 @@ private[parser] trait SimpleHeaders { this: Parser with CommonRules with CommonA
 
   // http://tools.ietf.org/html/rfc7233#section-4.2
   def `content-range` = rule {
-    (`byte-content-range` | `other-content-range`) ~ EOI ~> (`Content-Range`(_, _))
+    `byte-content-range` ~ EOI ~> (`Content-Range`(_, _)) | `other-content-range` ~ EOI ~> (`Content-Range`(_, _))
```
mdedetrich, line 115: This change is somewhat suspicious, let me look into it

http-core/src/main/scala/org/apache/pekko/http/impl/util/Rendering.scala
```diff
@@ -232,7 +233,7 @@ private[http] trait Rendering {
   /**
    * Renders the given string in double quotes.
    */
-  def ~~#!(s: String): this.type = ~~('"').putEscaped(s) ~~ '"'
+  def ~~#!(s: String): this.type = this.~~('"').putEscaped(s, Rendering.`\"`, '\\').~~('"')
```
mdedetrich, line 236: Another suspicious change

project/Dependencies.scala
```diff
@@ -28,15 +28,16 @@ object Dependencies {
...
   val specs2Version = "4.10.6"
-  val scalaCheckVersion = "1.14.3"
+  val scalaCheckVersion = "1.15.4"
 
   val scalafixVersion = _root_.scalafix.sbt.BuildInfo.scalafixVersion // grab from plugin
 
   val scala212Version = "2.12.17"
   val scala213Version = "2.13.10"
-  val allScalaVersions = Seq(scala213Version, scala212Version)
+  val scala3Version = "3.2.2"
```
pjfanning, line 39: will we go ahead and upgrade to Scala 3.3.0?

mdedetrich: I would prefer to do this in a separate PR.

project/Doc.scala
```diff
@@ -76,7 +76,13 @@ object Scaladoc extends AutoPlugin {
       // Workaround https://issues.scala-lang.org/browse/SI-10028
       "-skip-packages", "org.apache.pekko.pattern:org.specs2",
       "-doc-canonical-base-url", "https://pekko.apache.org/api/pekko-http/current/") ++
-      plugins.map(plugin => "-Xplugin:" + plugin)
+      plugins.map(plugin => "-Xplugin:" + plugin) ++
+      // Workaround https://issues.scala-lang.org/browse/SI-10028
+      (if (scalaBinaryVersion == "3")
+         // https://github.com/lampepfl/dotty/issues/14939
+         List("-skip-packages:akka.pattern:org.specs2")
```
pjfanning, line 83: should be 'org.apache.pekko.pattern' - this 'akka.pattern' issue appears in `else` too

docs/src/test/scala/docs/http/scaladsl/server/directives/JsonStreamingExamplesSpec.scala
```diff
@@ -218,7 +222,7 @@ class JsonStreamingExamplesSpec extends RoutingSpec with CompileOnlySpec {
               .runFold(0) { (cnt, _) => cnt + 1 }
 
           complete {
-            measurementsSubmitted.map(n => Map("msg" -> s"""Total metrics received: $n"""))
+            measurementsSubmitted.map(n => s"""Total metrics received: $n""")
```
mdedetrich, line 225: Another suspicious change

parsing/src/main/scala/org/apache/pekko/macros/LogHelper.scala
```diff
@@ -13,11 +13,8 @@
 
 package org.apache.pekko.macros
 
-import org.apache.pekko
-import pekko.annotation.InternalApi
-import pekko.event.LoggingAdapter
-
-import scala.reflect.macros.blackbox
+import org.apache.pekko.annotation.InternalApi
```
pjfanning, line 16: maybe fix this import - move with rest of pekko ones and remove org.apache bit

mdedetrich: Fixed and pushed

project/Doc.scala
mdedetrich: Fixed and pushed

docs/src/test/scala/docs/http/scaladsl/server/directives/JsonStreamingExamplesSpec.scala
```diff
@@ -235,7 +239,7 @@ class JsonStreamingExamplesSpec extends RoutingSpec with CompileOnlySpec {
 
     Post("/metrics", entity = data) ~> route ~> check {
       status should ===(StatusCodes.OK)
-      responseAs[String] should ===("""{"msg":"Total metrics received: 2"}""")
+      responseAs[String] should ===("Total metrics received: 2")
```
mdedetrich, line 242: Another suspicious change

docs/src/test/scala/docs/http/scaladsl/server/directives/RouteDirectivesExamplesSpec.scala
mdedetrich, line 1: Changes in this entire file are suspicious

docs/src/test/scala/docs/http/scaladsl/server/directives/JsonStreamingFullExamples.scala
```diff
@@ -41,21 +42,21 @@ class JsonStreamingFullExamples extends AnyWordSpec {
...
+    implicit val userFormat: JsonFormat[User] = jsonFormat2(User.apply)
 
     val `vnd.example.api.v1+json` =
       MediaType.applicationWithFixedCharset("vnd.example.api.v1+json", HttpCharsets.`UTF-8`)
     val ct = ContentType.apply(`vnd.example.api.v1+json`)
 
     implicit def userMarshaller: ToEntityMarshaller[User] = Marshaller.oneOf(
-      Marshaller.withFixedContentType(`vnd.example.api.v1+json`) { organisation =>
-        HttpEntity(`vnd.example.api.v1+json`, organisation.toJson.compactPrint)
+      Marshaller.withFixedContentType(`vnd.example.api.v1+json`) { (user: User) =>
```
mdedetrich, line 52: Another suspicious change

docs/src/test/scala/docs/http/scaladsl/server/directives/MarshallingDirectivesExamplesSpec.scala
```diff
@@ -53,6 +54,7 @@ class MarshallingDirectivesExamplesSpec extends RoutingSpec with CompileOnlySpec
 
   "example-entity-with-raw-json" in {
     // #example-entity-with-raw-json
+    import spray.json.JsValue
```
mdedetrich, line 57: Another suspicious change albeit this one seems harmless

docs/src/test/scala/docs/http/scaladsl/server/directives/RouteDirectivesExamplesSpec.scala
mdedetrich, line 1: This entire file change is suspicious.

mdedetrich, line 1: Entire file change is suspicious

project/Dependencies.scala
```diff
@@ -45,11 +46,19 @@ object Dependencies {
...
-    val scalaReflect = ScalaVersionDependentModuleID.versioned("org.scala-lang" % "scala-reflect" % _ % "provided") // Scala License
+    val scalaReflect = ScalaVersionDependentModuleID.fromPF {
+      case v if v.startsWith("2.") => "org.scala-lang" % "scala-reflect" % v % "provided" // Scala License
+    }
   }
 
   object Compile {
-    val scalaXml = "org.scala-lang.modules" %% "scala-xml" % "1.3.0" // Scala License
+    val scalaXml = {
+      val xml = "org.scala-lang.modules" %% "scala-xml" // Scala License
```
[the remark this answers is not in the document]
mdedetrich: iirc there was issues with scala-xml on JDK 8 but that is running in CI so lets see if it makes an impact

http-core/src/main/scala/org/apache/pekko/http/impl/model/parser/SimpleHeaders.scala
raboof: this change was part of 0a979b87f0 and indeed back then (on scala 3.1.3) was required to compile. The old syntax seems to work again on scala 3.2.2 and 3.3.0, though, so this change could be reverted.

http-core/src/main/scala/org/apache/pekko/http/impl/util/Rendering.scala
raboof: This change was part of ebd17d4dce and indeed still seems required to compile

mdedetrich: Thanks. Not sure if it makes sense to go through every comment that I marked as suspicious, I was just concerned whether a regression was made when dealing with the merge conflicts.

If you think all of the suspicious parts I marked are legitimate just let me know and we can finally review and merge the PR.

docs/src/test/scala/docs/http/scaladsl/HttpClientExampleSpec.scala
```diff
@@ -318,9 +323,10 @@ class HttpClientExampleSpec extends AnyWordSpec with Matchers with CompileOnlySp
     import pekko.http.scaladsl.unmarshalling.Unmarshal
     import pekko.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
     import spray.json.DefaultJsonProtocol._
+    import spray.json.RootJsonFormat
 
     case class Pet(name: String)
-    implicit val petFormat = jsonFormat1(Pet)
+    implicit val petFormat: RootJsonFormat[Pet] = jsonFormat1(Pet.apply)
```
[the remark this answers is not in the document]
pjfanning: thinking about this again, this is spray-json code so it should be covered by spray-json tests - we probably don't really need to do extra testing in pekko-http

docs/src/test/scala/docs/http/scaladsl/server/directives/JsonStreamingExamplesSpec.scala
pjfanning: this does seem like an important change - I think we need to understand why this needs to change

mdedetrich: Agreed, it basically does see like some test/business logic was updated and not related to scala 3 at all.

mdedetrich: Pinging @raboof

docs/src/test/scala/docs/http/scaladsl/server/directives/JsonStreamingExamplesSpec.scala
raboof: This change was part of 63b3437048. It might indicate a change in behavior, but likely it's a change in behavior between Scala versions rather than between before and after the changes to the module. Since the `Map` to json encoding doesn't seem super relevant to this particular test it seemed more productive to just simplify it to a regular string.

docs/src/test/scala/docs/http/scaladsl/server/directives/JsonStreamingExamplesSpec.scala
raboof: (counterpart of https://github.com/apache/incubator-pekko-http/pull/150#discussion_r1206637535)

docs/src/test/scala/docs/http/scaladsl/server/directives/JsonStreamingExamplesSpec.scala
mdedetrich: Right, so if I understand correctly there probably was some issue with construction of `Map` between Scala 2 and Scala 3 and the solution was "screw it, lets just not use a `Map` because its not important and its just in tests anyways.

raboof: Yes - specifically in having the implicits correctly pick up that the `Map` can automatically be converted to a Json object and be Marshallable. It's not useless to figure out and show how to make that work again, but the risk that this is a bug that needs to be fixed rather than just something that would need a change in the test code to work on Scala 3 feels relatively low.

http-core/src/main/scala/org/apache/pekko/http/impl/util/Rendering.scala
raboof: So far all changes seem motivated and not introduced while dealing with the merge conflicts, but I'll verify a couple more to increase our confidence they are all legitimate...

docs/src/test/scala/docs/http/scaladsl/server/directives/JsonStreamingExamplesSpec.scala
mdedetrich: I wouldn't be surprised if this was since solved in later Scala versions, there is some merit in spending a low amount of time to figure out if its solved but I will look into this when a future PR updating pekko to Scala 3.3 LTS is out.

Thanks for spending time looking into it!

raboof: I checked that "just" switching back to a map is not sufficient even on 3.3.0. Might be just a matter of bringing into scope the right implicit, but I don't think I'll dive into that now.

mdedetrich:
@pjfanning @He-Pin @raboof So I think the PR is ready to review, it turns out that none of the issues were really suspicious but changes legitimately necessary for Scala 3.

Regarding Scala 3.3, I changed my mind and I will try and update it in this PR if its trivial to do so.

docs/src/test/scala/docs/http/scaladsl/server/directives/RouteDirectivesExamplesSpec.scala
raboof: This change was part of 63b34370483aab3e10716e031d98c75c1bfc2e40. The `Content-Type` case class constructor is `private[http]`, because people are supposed to use the HttpEntity subclass to access it, rather than looking at the headers themselves. Apparently this was somehow not always correctly enforced on Scala 2.

On Scala 3 this appears to have been fixed, so the test wouldn't compile as-is anymore, and it seemed more 'realistic' to switch to a different header to test the route directives.

If it turns out there are legitimate use cases for constructing `Content-Type` header objects we might consider removing the `private[http]` from that constructor, but for now it seems conservative to keep it.

mdedetrich: Thanks!

docs/src/test/scala/docs/http/scaladsl/server/directives/JsonStreamingFullExamples.scala
raboof: Also part of 63b3437048, seems to be a case of scala3 not inferring the type of this parameter without it. Still seems to be the case for Scala 3.3.0

docs/src/test/scala/docs/http/scaladsl/server/directives/MarshallingDirectivesExamplesSpec.scala
raboof: Part of 63b3437048, likely not strictly needed but in general we try to include relevant imports in code snippets that end up in the docs.

pjfanning:
I've been reviewing this and it seems ready to go. I'll approve when the scala 3.3 change is in.

mdedetrich:
Okay ill push the commit now, its a trivial one liner.

raboof (approved):
Great, nice group effort ;) .

Looks like this breaks the scalafix build setup, but that seems worth it IMHO.

mdedetrich:
> Looks like this breaks the scalafix build setup, but that seems worth it IMHO.

Indeed I will make an issue on this and we can solve it later.

PR #160, opened 2023-05-26
upgrade pekko snapshot

pjfanning:
to get a version that uses Scala 3.3.0

### DIFF
--- a/project/PekkoDependency.scala
+++ b/project/PekkoDependency.scala
@@ -53,7 +53,7 @@ object PekkoDependency {
   }
 
   // Default version updated only when needed, https://pekko.apache.org/docs/pekko/current/project/downstream-upgrade-strategy.html
-  val minimumExpectedPekkoVersion = "0.0.0+26671-d5a1b2ef-SNAPSHOT"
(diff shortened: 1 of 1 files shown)


project/PekkoDependency.scala
```diff
@@ -53,7 +53,7 @@ object PekkoDependency {
   }
 
   // Default version updated only when needed, https://pekko.apache.org/docs/pekko/current/project/downstream-upgrade-strategy.html
-  val minimumExpectedPekkoVersion = "0.0.0+26656-898c6970-SNAPSHOT"
+  val minimumExpectedPekkoVersion = "0.0.0+26671-d5a1b2ef-SNAPSHOT"
```
mdedetrich, line 56: Can we change this to `0.0.0+26669-ec5b6764-SNAPSHOT`, otherwise we get dependency conflicts and I remember you changing some other pekko module to `0.0.0+26669-ec5b6764-SNAPSHOT` when Scala 3 was merged into it.

mdedetrich (changes requested):
One change

mdedetrich (approved):
lgtm

project/PekkoDependency.scala
pjfanning: switched

PR #170, opened 2023-06-03
Use java NIO to create temp file

mdedetrich:
`java.io.File.createTempFile` creates files on Linux that are readable by everyone, we should be using `java.nio.Files.createTempFile` instead.

### DIFF
--- a/http/src/main/scala/org/apache/pekko/http/scaladsl/server/directives/FileUploadDirectives.scala
+++ b/http/src/main/scala/org/apache/pekko/http/scaladsl/server/directives/FileUploadDirectives.scala
@@ -14,6 +14,7 @@
 package org.apache.pekko.http.scaladsl.server.directives
 
 import java.io.File
+import java.nio.file.Files
 
 import org.apache.pekko
 import pekko.Done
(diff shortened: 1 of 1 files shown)


http/src/main/scala/org/apache/pekko/http/scaladsl/server/directives/FileUploadDirectives.scala
```diff
@@ -175,9 +176,12 @@ trait FileUploadDirectives {
       implicit val ec = ctx.executionContext
 
       def tempDest(fileInfo: FileInfo): File = {
-        val dest = File.createTempFile("pekko-http-upload", ".tmp")
-        dest.deleteOnExit()
-        dest
+        val dest = Files.createTempFile("pekko-http-upload", ".tmp")
+        Runtime.getRuntime.addShutdownHook(new Thread() {
```
mdedetrich, line 180: This is the equivalent of what `dest.deleteOnExit` does behind the scenes.

mdedetrich:
Pinging @raboof since its possibly security related

pjfanning:
Should an Akka issue be raised?

http/src/main/scala/org/apache/pekko/http/scaladsl/server/directives/FileUploadDirectives.scala
He-Pin, line 180: Will there have many hook threads?

mdedetrich: > Will there have many hook threads?

I don't think so but note that this is the same implementation as `deleteOnExit` so it should behave the same as current version.

pjfanning (approved):
Lgtm

PR #171, opened 2023-06-05
Cache snapshots to prevent overloading Apache Nexus repo

mdedetrich:
The Apache Snapshots nexus repo is sometimes giving us request refused, likely because we are overloading the repo with excessive amount of requests.  This is due to the fact that we are using `SNAPSHOT` versions which by default don't cache because they are mutable (i.e. you can reupload snapshots with different implementations as much as you like).

Since for both pekko and its project forks the snapshot version is derived from git hash, there is no disadvantage to caching the snapshot.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -42,7 +42,9 @@ inThisBuild(Def.settings(
   },
   scalafixScalaBinaryVersion := scalaBinaryVersion.value,
   apacheSonatypeProjectProfile := "pekko",
-  versionScheme := Some("semver-spec")))
+  versionScheme := Some("semver-spec"),
+  // TODO: Remove when Pekko has a proper release
+  updateOptions := updateOptions.value.withLatestSnapshots(false)))
 
 // When this is updated the set of modules in Http.allModules should also be updated
 lazy val userProjects: Seq[ProjectReference] = List[ProjectReference](
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -26,8 +26,10 @@ addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.10.0-RC1") // for 
 addSbtPlugin("io.spray" % "sbt-boilerplate" % "0.6.1")
 addSbtPlugin("com.lightbend.sbt" % "sbt-bill-of-materials" % "1.0.2")
 addSbtPlugin("com.github.sbt" % "sbt-license-report" % "1.5.0")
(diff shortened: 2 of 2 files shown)


build.sbt
```diff
@@ -42,7 +42,9 @@ inThisBuild(Def.settings(
   },
   scalafixScalaBinaryVersion := scalaBinaryVersion.value,
   apacheSonatypeProjectProfile := "pekko",
-  versionScheme := Some("semver-spec")))
+  versionScheme := Some("semver-spec"),
+  // TODO: Remove when Pekko has a proper release
+  updateOptions := updateOptions.value.withLatestSnapshots(false)))
```
pjfanning, line 47: We'll have staged releases to test soon. This change would have been useful in the past. Now its benefit is reduced.

mdedetrich: We are currently effectively DDOSing the Apache snapshots repo so I think this takes precedence

pjfanning: Can you remove the TODO? This change is probably fine as long as we keep using sbt-dynver

pjfanning: * you occasionally might need to try out a pre-release change in one of our downstream repos
* and then you have to remember to add back `updateOptions := updateOptions.value.withLatestSnapshots(false))`

mdedetrich: Sure, honestly the real problem here is that `updateOptions.value.withLatestSnapshots(false))` is not granular. In an idea world you would be able to set `updateOptions.value.withLatestSnapshots(false))` for artifacts that are only download remotely where we can verify that snapshots are immutable since they are version with git hashes from `main` branch.

I think we are scraping the bottom of the barrel a bit here, this is meant to be a temporary work around so we don't overload the ASF repo with requests and while there may be arguments for leaving `updateOptions.value.withLatestSnapshots(false))` after release there is also a reason why the default of `withLatestSnapshots` is `true` in that its incredibly frustrating/surprising when you actually happen to have mutable snapshots. The overloading of requests is not going to be an issue if a person occasionally uses snapshots locally, its an issue if we have 12+ repo's building nightly snapshots resolving against pekko snapshots.

@pjfanning Is this PR still a real blocker for you? We can always revert/revisit this later but the equivalent of these changes have already been merged in other pekko modules.

pjfanning (approved):
lgtm

mdedetrich:
Thanks!

PR #178, opened 2023-06-10
make Content-Type and Content-Length constructors accessible in pekko packages

pjfanning:
Pekko-Connectors S3 uses these constructors.

See https://github.com/apache/incubator-pekko-connectors/pull/167

Scala 3 is stricter about private and package private access.


mdedetrich (approved):
lgtm

jrudolph:
They are private because they must only be used internally. AFAICS there's no good reason to use the headers in the S3 connector in that way, in any case, it's not related to the module's use of pekko-http, and there the HTTP headers are just used for convenience.

PR #180, opened 2023-06-16
add build docs

pjfanning:
* copied from https://github.com/apache/incubator-pekko
* in anticipation that this repo will soon head to RC

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ The documentation is available at
 
 ### Running the Build
 - Open a command window and change directory to your preferred base directory
-- Use git to clone the [repo](https://github.com/apache/incubator-pekko) or download a source release from https://pekko.apache.org (and unzip or untar it, as appropriate)
+- Use git to clone the [repo](https://github.com/apache/incubator-pekko-http) or download a source release from https://pekko.apache.org (and unzip or untar it, as appropriate)
 - Change directory to the directory where you installed the source (you should have a file called `build.sbt` in this directory)
 - `sbt compile` compiles the main source for project default version of Scala (2.13)
     - `sbt +compile` will compile for all supported versions of Scala
--- a/README.md
+++ b/README.md
@@ -31,9 +31,8 @@ The documentation is available at
 ## Building from Source
 
 ### Prerequisites
-- Make sure you have installed a Java Development Kit (JDK) version 11 or later.
+- Make sure you have installed a Java Development Kit (JDK) version 8 or later.
 - Make sure you have [sbt](https://www.scala-sbt.org/) installed and using this JDK.
-- Make sure you also have Java 8 installed (JDK or JRE).
 - [Graphviz](https://graphviz.gitlab.io/download/) is needed for the scaladoc generation build task, which is part of the release.
 
 ### Running the Build
@@ -49,7 +48,6 @@ The documentation is available at
 - `sbt publishLocal` will push the jars to your local Apache Ivy repository
 - `sbt publishM2` will push the jars to your local Apache Maven repository
 - `sbt docs/paradox` will build the docs (the ones describing the module features)
-     - Requires Java 11 minimum
      - `sbt docs/paradoxBrowse` does the same but will open the docs in your browser when complete
      - the `index.html` file will appear in `target/paradox/site/main/`
 - `sbt unidoc` will build the Javadocs for all the modules and load them to one place (may require Graphviz, see Prerequisites above)
--- a/README.md
+++ b/README.md
@@ -32,8 +32,10 @@ The documentation is available at
 
 ### Prerequisites
 - Make sure you have installed a Java Development Kit (JDK) version 8.
-    - there is an issue with the jars produced by higher JDK versions  
-- Make sure you have [sbt](https://www.scala-sbt.org/) installed and using this JDK.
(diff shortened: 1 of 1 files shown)


mdedetrich (changes requested):
Only Pekko needs dual JDK's, currently Pekko-Http only needs JDK 8.

pjfanning:
> Only Pekko needs dual JDK's, currently Pekko-Http only needs JDK 8.

@mdedetrich I made the change to remove the mention of having 2 JDKs

README.md
```diff
@@ -22,21 +21,50 @@ might require you to write more application code.
...
 [Scala](https://pekko.apache.org/docs/pekko-http/current/scala/http/) and
 [Java](https://pekko.apache.org/docs/pekko-http/current/java/http/).
 
-
-Community
----------
+## Building from Source
+
+### Prerequisites
+- Make sure you have installed a Java Development Kit (JDK) version 8 or later.
```
mdedetrich, line 34: I woukd sayJDK 8 only (i.e. remove the later part) because Pekko Http build needs to produce JDK 8 bytecode. If you build with after JDK then JDK 8 users cannot run Pekko Http.


This can be resolved with the `-release` flag which pekko http uses but with conditionals which may not be relevant now (i.e. latest versions of Scala 2 fixes some problems) but the point is we need to be careful about this and verify it.

mdedetrich (changes requested):
See comment

README.md
pjfanning: our pekko repo says JDK 11+

mdedetrich: The critical one, i.e. building jar artifacts is JDK 8 and we inherited that from Akka https://github.com/apache/incubator-pekko-http/blob/main/.github/workflows/publish.yml#L19 .

Also a distinction should be made between building release artifacts to validate release process and just working on the code locally.

I am on my phone atm so it's a bit hard to read entire thing in context but for me the important part is that it's clear that when creating release artifacts (for verification or doing the release yourself as a release manager) that you use JDK 8.

And it might well be that it cresting release artifacts for JDK 11 or higher is fine but after going through the process of validating/improving/clarifying this in root pekko i would rather be safe

pjfanning: this appears in our build.sbt

```
    (if (isJdk8) Seq.empty
     else if (scalaBinaryVersion.value == "2.12") Seq("-target:jvm-1.8")
     else Seq("-release", "8")),
```

mdedetrich: Yes I know but this was for older versions of Scala. `-target` in the latest versions of Scala is either deprecated or removed and ontop of the the behaviour of `-release` has changed.

pjfanning: I built pekko-http on my laptop with JDK11 and was then able to use the jars that I published to MavenLocal with JDK 8 builds of downstream projects. The classes appear to be Java 8 format.

mdedetrich: I am more concerned about stuff like unsafe or missing references to documentation. I'll check this properly when I get home

mdedetrich:
@pjfanning So I haven't had proper time to check this, but the issue I am concerned about specifically is https://github.com/classgraph/classgraph/issues/403#issuecomment-586770739 and https://github.com/apache/incubator-pekko/pull/258/files#diff-c11a10bd9e773399e4272cd1ee5dff72545afb2f8bed23f2d1bbc49d4e1fc03fL60-L67 . JDK 9+ introduced a very subtle but possible breaking change from the JDK 8 standard library. If the problem is there, you won't hit at compile time, you hit them at runtime and only if you hit specific codepaths i.e. code that calls `byteBuffer.clear()` (and I briefly checked pekko-http and it does call this code).

To confirm that there isn't a problem I have to manually inspect the bytecode, just like I did here https://github.com/apache/incubator-pekko/pull/258#issuecomment-1477908162

mdedetrich:
@pjfanning Okay so I have successfully run [jardiff](https://github.com/lightbend-labs/jardiff) on two versions of pekko-http, one built with JDK 8 and the other built with JDK 11 using

```
jardiff -g ~/github/diff-temp/ ~/github/incubator-pekko-http/http-core/target/scala-2.12/ ~/github/incubator-pekko-http-2/http-core/target/scala-2.12/
```

and as I suspected we have the problem I noted before

image
(this is with `HttpHeaderParser$.class`)

mdedetrich:
@pjfanning I managed to fix the issue that I described earlier that will let you produce jars correctly on JDK 11, will create the PR soon (having dinner now)

mdedetrich:
@pjfanning I have just created https://github.com/apache/incubator-pekko-http/pull/187 which allows building correct artifacts for pekko-http with JDK 11 so you can use the original documentation that you provided in this PR.

Once that PR is merged I can approve this one.

mdedetrich (approved):
Since https://github.com/apache/incubator-pekko-http/pull/187 is merged I will approve this PR, you may want to update it since I have also verified its safe to build with JDK 11 (newer versions than that I haven't checked yet).

README.md
jrudolph: The safe option is to build with JDK 8. The compatibility mode of JDK 11 will probably work most of the times but we ran into issues in the past, where the JDK signature storage that allows JDK 11 to figure out which methods are present in JDK 8 was not completely accurate.

PR #181, opened 2023-06-16
use latest pekko snapshot

pjfanning:
with a Pekko Core RC coming soon, it is important to verify everything


pjfanning:
@mdedetrich if we get this merged, I can run the publish job and note down the pekko-http version - allowing me to update more repos - ones that need both pekko and pekko-http

PR #182, opened 2023-06-16
compile warning about var that could be a val

pjfanning:
```
[warn] /Users/pj.fanning/code/incubator-pekko-http/http/src/main/scala/org/apache/pekko/http/scaladsl/coding/GzipCompressor.scala:75:23: local var crc32 in <$anon: GzipDecompressor.this.ParsingLogic> is never updated: consider using immutable val
[warn]     private[this] var crc32: CRC32 = new CRC32
```


mdedetrich (approved):
lgtm

Issue #183, opened 2023-06-16
Scala 3 only: `java.lang.NoSuchMethodError: ...TLSProtocol$SessionBytes$.unapply(...)`


mkurz:
[This scripted test](https://github.com/playframework/playframework/tree/main/dev-mode/sbt-plugin/src/sbt-test/play-sbt-plugin/http-backend-akka-http-http2), migrated to pekko in https://github.com/playframework/playframework/pull/11847

* Works fine with Scala 2.13 :heavy_check_mark: 
* Fails with Scala 3 :x: 

Seems like this has to do with how `unapply` methods changed in Scala 3 (?)
* https://docs.scala-lang.org/scala3/guides/migration/incompat-other-changes.html#explicit-call-to-unapply
* https://docs.scala-lang.org/scala3/reference/changed-features/pattern-matching.html

Snapshot used:
* pekko-http: `0.0.0+4411-6fe04045-SNAPSHOT`
* pekko: `0.0.0+26669-ec5b6764-SNAPSHOT`

```
[error] Uncaught error from thread [application-pekko.actor.default-dispatcher-4]: 'org.apache.pekko.stream.TLSProtocol$SessionBytes org.apache.pekko.stream.TLSProtocol$SessionBytes$.unapply(org.apache.pekko.stream.TLSProtocol$SessionBytes)', shutting down JVM since 'pekko.jvm-exit-on-fatal-error' is enabled for ActorSystem[application]
[error] java.lang.NoSuchMethodError: 'org.apache.pekko.stream.TLSProtocol$SessionBytes org.apache.pekko.stream.TLSProtocol$SessionBytes$.unapply(org.apache.pekko.stream.TLSProtocol$SessionBytes)'
[error]         at org.apache.pekko.http.impl.engine.http2.ProtocolSwitch$$anon$1$$anon$2.onPush(ProtocolSwitch.scala:65)
[error]         at org.apache.pekko.stream.impl.fusing.GraphInterpreter.processPush(GraphInterpreter.scala:555)
[error]         at org.apache.pekko.stream.impl.fusing.GraphInterpreter.execute(GraphInterpreter.scala:433)
[error]         at org.apache.pekko.stream.impl.fusing.GraphInterpreterShell.runBatch(ActorGraphInterpreter.scala:662)
[error]         at org.apache.pekko.stream.impl.fusing.GraphInterpreterShell$AsyncInput.execute(ActorGraphInterpreter.scala:532)
[...]

PR #184, opened 2023-06-16
avoid unapply due to issue reported by Play team (scala 3 only)

pjfanning:
* relates to https://github.com/apache/incubator-pekko-http/issues/183
* simplifies the code, we don't really need the unapply

### DIFF
--- a/http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/ProtocolSwitch.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/ProtocolSwitch.scala
@@ -62,10 +62,10 @@ private[http] object ProtocolSwitch {
               new InHandler {
                 def onPush(): Unit =
                   grab(netIn) match {
-                    case first @ SessionBytes(session, bytes) =>
+                    case first: SessionBytes =>
                       val chosen = chosenProtocolAccessor(first)
                       chosen match {
-                        case "h2" => install(http2Stack.addAttributes(HttpAttributes.tlsSessionInfo(session)), first)
+                        case "h2" => install(http2Stack.addAttributes(HttpAttributes.tlsSessionInfo(first.session)), first)
                         case _    => install(http1Stack, first)
                       }
                     case SessionTruncated =>
--- a/http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/ProtocolSwitch.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/ProtocolSwitch.scala
@@ -65,8 +65,9 @@ private[http] object ProtocolSwitch {
                     case first: SessionBytes =>
                       val chosen = chosenProtocolAccessor(first)
                       chosen match {
-                        case "h2" => install(http2Stack.addAttributes(HttpAttributes.tlsSessionInfo(first.session)), first)
-                        case _    => install(http1Stack, first)
+                        case "h2" =>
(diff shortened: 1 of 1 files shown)


mkurz:
Nice that was fast :+1:

mkurz (approved):
LGTM

mdedetrich:
Nice work!

Is it possible to write a small unit test that fails before this change and passes with this change?

pjfanning:
> Nice work!
> 
> Is it possible to write a small unit test that fails before this change and passes with this change?

I added https://github.com/apache/incubator-pekko-http/pull/185 to see if any tests went through this code. Plenty do.
I suspect that something more is going on.

@mkurz mentioned that where he saw the issue was in sbt scripted tests. Maybe sbt could be doing something like instrumenting the classes under test.

mdedetrich:
Well since this is a compiler change I am not that fussed about a unit test for this specific part of the code, I just suggested it because its easier.

What would be more ideal is a normal/integration test that happens to cover this, i.e. we should be testing this part of the code one way or another.

mdedetrich (approved):
lgtm

pjfanning:
The code is actually well tested. So other than adding the sbt scripted test from the Play Framework, I am not really sure how to extend the tests here.

This is fairly minor refactor, since the unapply is not really necessary. We already have the match case matching to a variable called `first`.

mdedetrich:
> The code is actually well tested. So other than adding the sbt scripted test from the Play Framework, I am not really sure how to extend the tests here.
> 
> This is fairly minor refactor, since the unapply is not really necessary. We already have the match case matching to a variable called `first`.

Well unless I am missing something, evidently we dont have a test that goes through this code path because otherwise we would have hit it. Or is the case that we aren't running tests with Scala 3 specifically (in which case we should?)

pjfanning:
We have lots of tests going through the code and we do test with Scala 3.

With sbt scripted tests, there could be asm manipulation of the classes or something like that. I had to investigate an issue in Jackson recently where a user found a problem when using Gradle with Jackson and it turned out that Gradle was using asm and the issue was stemming from there.

mdedetrich:
> With sbt scripted tests, there could be asm manipulation of the classes or something like that.

I don't think this is the case, sbt-scripted just compiles the code and generates the jar just like a normal user would and just runs the code, in essence its closer to a smoke test. As a hypothetical example if we had inlining in pekko-http that could be an example, i.e. inlining only enabled on CI/build but not when running normally.

My hunch is honestly that we are not going through this code path, but looks like you are already on it here https://github.com/apache/incubator-pekko-http/pull/185

He-Pin:
A Scala 3 bug?

http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/ProtocolSwitch.scala
```diff
@@ -62,11 +62,12 @@ private[http] object ProtocolSwitch {
               new InHandler {
                 def onPush(): Unit =
                   grab(netIn) match {
-                    case first @ SessionBytes(session, bytes) =>
+                    case first: SessionBytes =>
```
He-Pin, line 65: lgtm, fire an issue in dotty too?

He-Pin (approved):
lgtm

pjfanning:
> A Scala 3 bug?

Probably - but I haven't been able to come up with a minimised test case that could be reported to Dotty team.

PR #186, opened 2023-06-16
remove unused variable

pjfanning:
saw this declaration when investigating #183 

I'm not sure why the `logic` var was added. I appears to be unused. No compile or test issues occurred after removing this line.


mdedetrich (approved):
lgtm

PR #188, opened 2023-06-19
Fix low hanging fruit from warnings

mdedetrich:
Low hanging fruit safe to apply fixes from scalac warnings from various Scala versions


http/src/main/scala/org/apache/pekko/http/javadsl/unmarshalling/Unmarshaller.scala
```diff
@@ -119,7 +119,7 @@ object Unmarshaller extends pekko.http.javadsl.unmarshalling.Unmarshallers {
     unmarshalling.Unmarshaller.firstOf(u1.asScala, u2.asScala, u3.asScala, u4.asScala, u5.asScala)
   }
 
-  @nowarn("msg=parameter value mi in method adaptInputToJava is never used")
+  @nowarn("msg=mi in method adaptInputToJava is never used")
```
mdedetrich, line 122: The warning is slightly different depending on the Scala version, i.e. on one version its `parameter value mi in method adaptInputToJava is never used` and on another its `parameter mi in method adaptInputToJava is never used` so this will correctly supress the warnings on all versions of Scala.

pjfanning (approved):
lgtm

PR #189, opened 2023-06-19
Remove unneeded parser settings

mdedetrich:
As noted in the comments these options were needed for Parboiled but since we are now using upstream parboiled rather than manually inlined source (see https://github.com/apache/incubator-pekko-http/pull/14) the settings are no longer relevant.


pjfanning (approved):
lgtm

PR #196, opened 2023-06-20
use link to pekko code

pjfanning:
tidier to link to the pekko code

The Incubator PMC are searching for code for any links like this and they are causing us trouble with getting releases done.


mdedetrich (approved):
lgtm

PR #198, opened 2023-06-20
add utf8-decoder license

pjfanning:
* pekko-http-core already uses the root LICENSE because so far all mods to the standard license relate to pekko-http-core
* all other jars use the standard Apache License (which may change)


mdedetrich (approved):
lgtm

jrudolph (approved):
lgtm

PR #199, opened 2023-06-21
jsoniter-scala license

pjfanning:
* http-tests is not published as a jar - so there is no need to set up special build code to put a specific license in its jar


mdedetrich (approved):
lgtm

PR #204, opened 2023-06-27
add link validator CI job

pjfanning:
* we have similar CI jobs in other repos


pjfanning:
currently broken - https://github.com/apache/incubator-pekko-http/actions/runs/5388318117/jobs/9780805476

pjfanning:
with https://github.com/apache/incubator-pekko-http/pull/204 - we're getting to the point where these broken Java API links are the main remaining issue

mdedetrich:
@pjfanning Can you confirm that link validator works when https://github.com/apache/incubator-pekko-http/pull/231 is merged?

mdedetrich:
Hmm problem is still there, let me look into it

.github/workflows/link-validator.yml
```diff
@@ -0,0 +1,42 @@
...
+        if: github.event.pull_request
+        run: |-
+          git fetch origin pull/${{ github.event.pull_request.number }}/merge:scratch
+          git checkout scratch
+
+      - name: Setup Java 11
+        uses: actions/setup-java@v3
+        with:
+          distribution: temurin
+          java-version: 11
```
pjfanning, line 29: @mdedetrich this new yml already uses java 11

mdedetrich: Yes I can replicate this locally with JDK 11 as well, looking into it

mdedetrich: @pjfanning I figured out the core problem, see  https://github.com/scala/scala/pull/8663 and https://github.com/scala/bug/issues/11839 for why doc generation is broken on JDK 11+.

The doc generation is broken for JDK 11 and it works fine if you set the JDK to 8 which is what I would recommend since its also more correct (pekko-http supports JDK 8+).

This also explains why originally pekko-http used JDK 8 for generating docs, which means I also need to revert https://github.com/apache/incubator-pekko-http/pull/231 since right now doc generation is actually broken.

You can replicate this yourself by generating docs with `sbt -Dpekko.genjavadoc.enabled=true   "Javaunidoc/doc; Compile/unidoc"` using JDK 8 vs JDK 11. On JDK 8 it works and on 11 its broken. Updating this PR to JDK 8 should also solve the problem.

mdedetrich: @pjfanning Perfect, setting this to JDK 8 seems to have fixed this specific issue (see https://github.com/apache/incubator-pekko-http/actions/runs/5575596994/jobs/10185743202?pr=204). There are other problems i.e.
```
`[https://www.scala-lang.org/api/2.13.11/scala/runtime/AbstractFunction1.html`](https://www.scala-lang.org/api/2.13.11/scala/runtime/AbstractFunction1.html%60) status 404 Not Found
 - ../../../../../target/scala-2.13/unidoc/org/apache/pekko/http/scaladsl/server/index.html
 ```

But thats another problem to solve (looking into it now).

mdedetrich:
For the `AbstractFunction1` issue, it needs to be manually ignored (see https://github.com/apache/incubator-pekko/blob/34a3fa704c3ec4cbabe1956bf3ffc769ba6fc5ba/scripts/link-validator.conf#L59)

.github/workflows/link-validator.yml
pjfanning: added an exclude for this

PR #206, opened 2023-06-28
remove broken link from docs - cannot find another link that works

pjfanning:
@kazuhiro if you could provide an alternate link for the broken one, we will use it

https://kazuhiro.github.io/scala/akka/akka-http/akka-streams/2016/01/31/connection-pooling-with-akka-http-and-source-queue.html


mdedetrich (approved):
lgtm

kazuhiro:
Thanks for pinging. However, I shut down this blog for good.

PR #208, opened 2023-06-29
Add pekko-http-cors module

mdedetrich:
This PR adds pekko-http-cors support to pekko-http via the merge of akka-http-cors/pekko-http-cors (see https://github.com/lomigmegard/pekko-http-cors/issues/33). Notable points

* Legal disclaimers have been added (both to the root `LEGAL` file and to the `META-INF` jar indicating that the code was granted to ASF via SGA)
* The code is exactly the same as https://github.com/lomigmegard/pekko-http-cors with the expected package/import changes with these exceptions
  * One breaking code change where `org.apache.pekko.http.cors.javadsl.settings.CorsSettings.getMaxAge` was changed from `Optional[Long]` to `OptionalLong` since this is consistent within the rest of the Akka/Pekko ecosystem (and is also idiomatic for Java use)
  * A non breaking change where explicit return types where added to the `get*` public Java API methods in `CorsSettings`
  * ~~`application.conf` was renamed to `reference.conf` (using `application.conf` was likely an oversight, it should be used for actual applications where as `reference.conf` should be used for libraries).~~ This was an accident on my part
  * The root typesafe config was changed from `pekko-http-cors` to `pekko.http.cors` to be consistent with the rest of the Pekko Http.
* Migration notes have been added including the previously listed changes
* Docs have been added to the standard place (i.e. `docs/src/main/paradox/common/cors.md`) which is largely inspired by https://github.com/lomigmegard/pekko-http-cors/blob/master/README.md however I have made some liberties in changes so its more consistent with Pekko documentation, specifically
  * The original docs had inline source code that was Scala, this was updated to use different groupings for Java/Scala which means that Java equivalents had to be added.
  * The mention of default values has been removed since this is already documented in `reference.conf` and has been added to `docs/src/main/paradox/configuration.md`.
  * The sample code is directly included using the Scala/java `@@snip` rather than just referencing it.
  * In general there is an argument that the various sub headings such as `#### allowGenericHttpRequests / getAllowGenericHttpRequests` should also be removed because they are duplicated in the scala docs and is also not typical with Pekko Http documentation (let me know what you think on this point).
  * This is the first case where have new code derived from somewhere outside of Akka via a SGA which means that our standard sbt-header-check needed to be updated in order to ignore the `pekko-http-cors` project. Note that only sources have been ignored, resources (such as `reference.conf`) still get checked mainly for the `# SPDX-License-Identifier: Apache-2.0` header. This is on the presumption that headers are not recommended/necessary for non derived/SGA approved Apache source code.
  
@lomigmegard Pinging you so you can also review the PR and let me know your thoughts. The most critical thing here is any changes that you think would need to be done now otherwise they would break users in the future (i.e. the `Optional[Long]` to `OptionalLong` is an example of this) which you weren't able to do in akka-http-cors due to breaking downstream users.

### DIFF
--- a/docs/src/main/paradox/migration-guide/index.md
+++ b/docs/src/main/paradox/migration-guide/index.md
@@ -11,3 +11,8 @@
 * Where class names have "Akka" in the name, the Pekko ones have "Pekko" - e.g. PekkoException instead of AkkaException
 * Configs in `application.conf` use "pekko" prefix instead of "akka"
 * We will soon provide a more detailed guide for migrating from Akka HTTP v10.2 to Apache Pekko HTTP
+* If you happen to be using [akka-http-cors](https://github.com/lomigmegard/akka-http-cors) this has been directly integrated into
+  Apache Pekko Http which means you should use the @ref:[pekko-http-cors artifact](../common/cors.md) instead. Note that the root configuration naming 
+  has also been updated (i.e. changing from `akka-http-cors` to `pekko.http.cors`) to make it consistent with the rest of Apache Pekko Http.
+  * In addition there is a single breaking change, the return type of the `org.apache.pekko.http.cors.javadsl.settings.CorsSettings.getMaxAge`
+  method has been changed from `java.util.Optional<Long>` to `java.util.OptionalLong` since it's more idiomatic.
--- a/docs/src/test/java/docs/http/javadsl/server/cors/CorsServerExample.java
+++ b/docs/src/test/java/docs/http/javadsl/server/cors/CorsServerExample.java
@@ -1,0 +1,82 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package docs.http.javadsl.server.cors;
+
+// #cors-server-example
+import org.apache.pekko.http.javadsl.model.StatusCodes;
+import org.apache.pekko.http.javadsl.server.*;
+
+import java.util.NoSuchElementException;
+import java.util.concurrent.ExecutionException;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+import static org.apache.pekko.http.cors.javadsl.CorsDirectives.cors;
+import static org.apache.pekko.http.cors.javadsl.CorsDirectives.corsRejectionHandler;
+
+public class CorsServerExample extends HttpApp {
+
+  public static void main(String[] args) throws ExecutionException, InterruptedException {
+    final CorsServerExample app = new CorsServerExample();
+    app.startServer("127.0.0.1", 9000);
+  }
+
+  protected Route routes() {
+
+    // Your CORS settings are loaded from `application.conf`
+
+    // Your rejection handler
+    final RejectionHandler rejectionHandler =
+        corsRejectionHandler().withFallback(RejectionHandler.defaultHandler());
+
+    // Your exception handler
+    final ExceptionHandler exceptionHandler =
+        ExceptionHandler.newBuilder()
+            .match(
+                NoSuchElementException.class,
+                ex -> complete(StatusCodes.NOT_FOUND, ex.getMessage()))
+            .build();
+
+    // Combining the two handlers only for convenience
+    final Function<Supplier<Route>, Route> handleErrors =
+        inner ->
+            Directives.allOf(
+                s -> handleExceptions(exceptionHandler, s),
+                s -> handleRejections(rejectionHandler, s),
+                inner);
+
+    // Note how rejections and exceptions are handled *before* the CORS directive (in the inner
+    // route).
+    // This is required to have the correct CORS headers in the response even when an error occurs.
+    return handleErrors.apply(
+        () ->
+            cors(
+                () ->
+                    handleErrors.apply(
+                        () ->
+                            concat(
+                                path("ping", () -> complete("pong")),
+                                path(
+                                    "pong",
+                                    () ->
+                                        failWith(
+                                            new NoSuchElementException(
+                                                "pong not found, try with ping")))))));
+  }
+}
+// #cors-server-example
--- a/http-cors/src/main/scala/org/apache/pekko/http/cors/javadsl/CorsDirectives.scala
+++ b/http-cors/src/main/scala/org/apache/pekko/http/cors/javadsl/CorsDirectives.scala
@@ -1,0 +1,50 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.pekko.http.cors.javadsl
+
+import java.util.function.Supplier
+
+import org.apache.pekko
+import pekko.http.javadsl.server.{ RejectionHandler, Route }
+import pekko.http.javadsl.server.directives.RouteAdapter
+import pekko.http.cors.CorsJavaMapping.Implicits._
+import pekko.http.cors.javadsl.settings.CorsSettings
+import pekko.http.cors.scaladsl
+import pekko.http.impl.util.JavaMapping
+
+object CorsDirectives {
+
+  import pekko.http.cors.scaladsl.{ CorsDirectives => D }
+
+  def cors(inner: Supplier[Route]): Route =
+    RouteAdapter {
+      D.cors() {
+        inner.get.delegate
+      }
+    }
+
+  def cors(settings: CorsSettings, inner: Supplier[Route]): Route =
+    RouteAdapter {
+      D.cors(JavaMapping.toScala(settings)) {
+        inner.get.delegate
+      }
+    }
+
+  def corsRejectionHandler: RejectionHandler =
+    new RejectionHandler(scaladsl.CorsDirectives.corsRejectionHandler)
+}
--- a/http-cors/src/main/scala/org/apache/pekko/http/cors/javadsl/settings/CorsSettings.scala
+++ b/http-cors/src/main/scala/org/apache/pekko/http/cors/javadsl/settings/CorsSettings.scala
@@ -1,0 +1,77 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.pekko.http.cors.javadsl.settings
+
+import java.util.OptionalLong
+import org.apache.pekko
+import pekko.actor.ActorSystem
+import pekko.annotation.DoNotInherit
+import pekko.http.javadsl.model.HttpMethod
+import pekko.http.cors.javadsl.model.{ HttpHeaderRange, HttpOriginMatcher }
+import pekko.http.cors.scaladsl
+import pekko.http.cors.scaladsl.settings.CorsSettingsImpl
+import com.typesafe.config.Config
+
+/**
+ * Public API but not intended for subclassing
+ */
+@DoNotInherit
+abstract class CorsSettings { self: CorsSettingsImpl =>
+
+  def getAllowGenericHttpRequests: Boolean
+  def getAllowCredentials: Boolean
+  def getAllowedOrigins: HttpOriginMatcher
+  def getAllowedHeaders: HttpHeaderRange
+  def getAllowedMethods: java.lang.Iterable[HttpMethod]
+  def getExposedHeaders: java.lang.Iterable[String]
+  def getMaxAge: OptionalLong
+
+  def withAllowGenericHttpRequests(newValue: Boolean): CorsSettings
+  def withAllowCredentials(newValue: Boolean): CorsSettings
+  def withAllowedOrigins(newValue: HttpOriginMatcher): CorsSettings
+  def withAllowedHeaders(newValue: HttpHeaderRange): CorsSettings
+  def withAllowedMethods(newValue: java.lang.Iterable[HttpMethod]): CorsSettings
+  def withExposedHeaders(newValue: java.lang.Iterable[String]): CorsSettings
+  def withMaxAge(newValue: OptionalLong): CorsSettings
+}
+
+object CorsSettings {
+
+  /**
+   * Creates an instance of settings using the given Config.
+   */
+  def create(config: Config): CorsSettings = scaladsl.settings.CorsSettings(config)
+
+  /**
+   * Creates an instance of settings using the given String of config overrides to override settings set in the class
+   * loader of this class (i.e. by application.conf or reference.conf files in the class loader of this class).
+   */
+  def create(configOverrides: String): CorsSettings = scaladsl.settings.CorsSettings(configOverrides)
+
+  /**
+   * Creates an instance of CorsSettings using the configuration provided by the given ActorSystem.
+   */
+  def create(system: ActorSystem): CorsSettings = scaladsl.settings.CorsSettings(system)
+
+  /**
+   * Settings from the default loaded configuration. Note that application code may want to use the `apply()` methods
+   * instead to have more control over the source of the configuration.
+   */
+  @deprecated("Use other CorsSettings constructors", "1.0.0")
+  def defaultSettings: CorsSettings = scaladsl.settings.CorsSettings.defaultSettings
+}
--- a/http-cors/src/main/scala/org/apache/pekko/http/cors/scaladsl/model/HttpHeaderRange.scala
+++ b/http-cors/src/main/scala/org/apache/pekko/http/cors/scaladsl/model/HttpHeaderRange.scala
@@ -1,0 +1,39 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.pekko.http.cors.scaladsl.model
+
+import org.apache.pekko
+import pekko.http.cors.javadsl
+import pekko.util.Helpers
+
+import scala.collection.immutable.Seq
+
+abstract class HttpHeaderRange extends javadsl.model.HttpHeaderRange
+
+object HttpHeaderRange {
+  case object `*` extends HttpHeaderRange {
+    def matches(header: String) = true
+  }
+
+  final case class Default(headers: Seq[String]) extends HttpHeaderRange {
+    private val lowercaseHeaders: Seq[String] = headers.map(Helpers.toRootLowerCase)
+    def matches(header: String): Boolean = lowercaseHeaders.contains(Helpers.toRootLowerCase(header))
+  }
+
+  def apply(headers: String*): Default = Default(Seq(headers: _*))
+}
--- a/http-cors/src/main/scala/org/apache/pekko/http/cors/scaladsl/model/HttpOriginMatcher.scala
+++ b/http-cors/src/main/scala/org/apache/pekko/http/cors/scaladsl/model/HttpOriginMatcher.scala
@@ -1,0 +1,90 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.pekko.http.cors.scaladsl.model
+
+import org.apache.pekko
+import pekko.http.javadsl.{ model => jm }
+import pekko.http.scaladsl.model.headers.HttpOrigin
+import pekko.http.cors.javadsl
+
+import scala.collection.immutable.Seq
+
+/**
+ * HttpOrigin matcher.
+ */
+abstract class HttpOriginMatcher extends javadsl.model.HttpOriginMatcher {
+  def matches(origin: HttpOrigin): Boolean
+
+  /** Java API */
+  def matches(origin: jm.headers.HttpOrigin): Boolean = matches(origin.asInstanceOf[HttpOrigin])
+}
+
+object HttpOriginMatcher {
+  case object `*` extends HttpOriginMatcher {
+    def matches(origin: HttpOrigin) = true
+  }
+
+  final case class Default(origins: Seq[HttpOrigin]) extends HttpOriginMatcher {
+    private class StrictHostMatcher(origin: HttpOrigin) extends (HttpOrigin => Boolean) {
+      override def apply(origin: HttpOrigin): Boolean = origin == this.origin
+    }
+
+    private class WildcardHostMatcher(wildcardOrigin: HttpOrigin) extends (HttpOrigin => Boolean) {
+      private val suffix: String = wildcardOrigin.host.host.address.stripPrefix("*")
+      override def apply(origin: HttpOrigin): Boolean = {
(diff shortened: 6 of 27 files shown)


LICENSE
```diff
@@ -276,3 +276,9 @@ pekko-http-core contains code based on code by Tim Kienzle with his permission t
 (C) 2000 by Tim Kientzle <kientzle@acm.org>
 
 ---------------
+
+pekko-http-cors contains code that was donated by Lomig Mégard to the Apache Software Foundation
+via a Software Grant Agreement <https://www.apache.org/licenses/contributor-agreements.html#grants>
+See <https://github.com/lomigmegard/pekko-http-cors/issues/33>.
```
pjfanning, line 282: We don't need this if the code was granted to us

pjfanning: if we include something, it should be in the NOTICE file but I don't think we need to include anything

legal/pekko-http-cors-jar-license.txt
```diff
@@ -0,0 +1,207 @@
...
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+
+---------------
+
+pekko-http-cors contains code that was donated by Lomig Mégard to the Apache Software Foundation
+via a Software Grant Agreement <https://www.apache.org/licenses/contributor-agreements.html#grants>
+See <https://github.com/lomigmegard/pekko-http-cors/issues/33>.
```
pjfanning, line 207: I don't think we need this - StandardLicense should be ok for this module

pjfanning (changes requested):
the license changes need to be thought through

mdedetrich:
@justinmclean Is it possible to comment on https://github.com/apache/incubator-pekko-http/pull/208#pullrequestreview-1505300594, i.e. do we need to mention in `LICENSE` code that has been granted via a SGA?

mdedetrich:
Compile failures are due to pekko RC2 being dropped, when @pjfanning creates pekko RC3 this will be resolved.

One thing that I noticed is that we need to do modifications to sbt-header, this is because our standard header i.e.

```
/*
 * Licensed to the Apache Software Foundation (ASF) under one or more
 * license agreements; and to You under the Apache License, version 2.0:
 *
 *   https://www.apache.org/licenses/LICENSE-2.0
 *
 * This file is part of the Apache Pekko project, which was derived from Akka.
 */
 ```
 
 Is not correct for pekko-http-cors since it was not derived from Akka (emphasis on `This file is part of the Apache Pekko project, which was derived from Akka.`). I will look at this in the afternoon.

**Comment by pjfanning** [comment]
RC3 jars area ready - see https://github.com/apache/incubator-pekko-http/pull/209

File: LICENSE (hunk 1 above)
Comment (He-Pin) [reply LICENSE:282]: Better we have a record somewhere?

File: docs/src/main/paradox/migration-guide/index.md [hunk 3]
```diff
@@ -11,3 +11,9 @@
 * Where class names have "Akka" in the name, the Pekko ones have "Pekko" - e.g. PekkoException instead of AkkaException
 * Configs in `application.conf` use "pekko" prefix instead of "akka"
 * We will soon provide a more detailed guide for migrating from Akka HTTP v10.2 to Apache Pekko HTTP
+* If you happen to be using [akka-http-cors](https://github.com/lomigmegard/akka-http-cors) this has been directly integrated into
+  Apache Pekko Http which means you should use the [pekko-http-cors artifact](common/cors.md) instead. Note that the root configuration naming 
+  has for cors has also been updated (i.e. changing from `akka-http-cors` to `pekko.http.cors`) to make it consistent with the
```
lomigmegard, line 16: typo: remove `has for cors`

```suggestion
  has also been updated (i.e. changing from `akka-http-cors` to `pekko.http.cors`) to make it consistent with the
```

http-cors/src/main/scala/org/apache/pekko/http/cors/javadsl/CorsDirectives.scala
```diff
@@ -0,0 +1,36 @@
...
+          case ra: RouteAdapter => ra.delegate
+        }
+      }
+    }
+
+  def cors(settings: CorsSettings, inner: Supplier[Route]): Route =
+    RouteAdapter {
+      // Currently the easiest way to go from Java models to their Scala equivalent is to cast.
+      // See https://github.com/akka/akka-http/issues/661 for a potential opening of the JavaMapping API.
+      val scalaSettings = settings.asInstanceOf[scaladsl.settings.CorsSettings]
```
lomigmegard, line 26: Now that the code lives inside pekko, we could actually use `JavaMapping`, if this comment is still relevant. At least the comment should be updated to not point to akka?

http-cors/src/main/scala/org/apache/pekko/http/cors/scaladsl/settings/CorsSettings.scala
```diff
@@ -0,0 +1,236 @@
...
+   * loader of this class (i.e. by application.conf or reference.conf files in the class loader of this class).
+   */
+  def apply(configOverrides: String): CorsSettings =
+    apply(ConfigFactory.parseString(configOverrides).withFallback(ConfigFactory.load(getClass.getClassLoader)))
+
+  /**
+   * Creates an instance of CorsSettings using the configuration provided by the given ActorSystem.
+   */
+  def apply(system: ActorSystem): CorsSettings =
+    // From private pekko.http.impl.util.SettingsCompanionImpl implementation
```
lomigmegard, line 187: The `SettingsCompanionImpl` could now be used, to remove the copy-pasted code

mdedetrich:
I have fixed the issue of sbt-header applying on files that it shouldn't, see the last point in https://github.com/apache/incubator-pekko-http/pull/208#issue-1780808574 for more details.

docs/src/test/java/docs/http/javadsl/server/cors/CorsServerExample.java
```diff
@@ -0,0 +1,65 @@
+package docs.http.javadsl.server.cors;
```
pjfanning, line 1: all our other `docs` package scala files have license headers - could this be given the standard Apache license header?

pjfanning (changes requested):

lomigmegard:
Thanks for the PR @mdedetrich. 

I reviewed it and found only a couple places where some utilities from pekko could be re-used, but it can be done in another commits (or even PRs). 

Regarding your question for breaking changes, the main point I see are the default values for the configuration. Especially `allowed-origins` and `allowed-headers` which are very permissive. It's a tradeoff between safer defaults, and defaults that "just work" for simple servers.

lomigmegard:
>  application.conf was renamed to reference.conf (using application.conf was likely an oversight, it should be used for actual applications where as reference.conf should be used for libraries).

It was already [reference.conf](https://github.com/lomigmegard/pekko-http-cors/blob/master/pekko-http-cors/src/main/resources/reference.conf) :)

http-cors/src/main/scala/org/apache/pekko/http/cors/javadsl/settings/CorsSettings.scala
```diff
@@ -0,0 +1,59 @@
+package org.apache.pekko.http.cors.javadsl.settings
```
pjfanning, line 1: all source files must have license headers - please use the standard Apache one

mdedetrich:
@lomigmegard 

> Regarding your question for breaking changes, the main point I see are the default values for the configuration. Especially allowed-origins and allowed-headers which are very permissive. It's a tradeoff between safer defaults, and defaults that "just work" for simple servers.

Thanks, when I was talking about breaking changes I was alluding to binary compatibility and not behavioural. However the point about cors permissions being overly permissive is one that should be noted.

> It was already [reference.conf](https://github.com/lomigmegard/pekko-http-cors/blob/master/pekko-http-cors/src/main/resources/reference.conf) :)

Hmm, I must have gave it the wrong name when copying, my bad!

docs/src/main/paradox/migration-guide/index.md
```diff
@@ -11,3 +11,9 @@
...
+  rest of Apache Pekko Http.
+  * In addition there is a single breaking change, the return type of the `org.apache.pekko.http.cors.javadsl.settings.CorsSettings.getMaxAge`
+  method has been changed from `java.util.Optional<Long>` to `java.util.OptionalLong` since its more idiomatic.
```
lomigmegard, line 19:
```suggestion
  method has been changed from `java.util.Optional<Long>` to `java.util.OptionalLong` since it's more idiomatic.
```

http-cors/src/main/scala/org/apache/pekko/http/cors/javadsl/settings/CorsSettings.scala
mdedetrich: I swear I saw it somewhere that headers for original/SGA code was a convention but I might be misremembering. I will fix this after I resolve @lomigmegard review comments.

pjfanning: the code is granted to us - we can add headers and we need an Apache header

if the original code has had headers, we' d have to keep them or get @lomigmegard 's permission to change them

LICENSE
pjfanning: Since the code is granted to us, I don't see how this differs much a standard PR that we get. And with PRs generally, we do not modify the LICENSE and we insist that all source files are given Apache license headers.

The LICENSE addenda are to point out 3rd party code that has different licenses or that has copyright declarations in those licenses (if the original code was ASL but had a specific copyright in it).

I would favour highlighting the source of the code but I would suggest that we do that in the source files adding something like:

```
// based on code from akka-http-cors <https://github.com/lomigmegard/akka-http-cors> which has been granted
// to the Apache Pekko project - the Software Grant agreement is on record with the Secretary of the ASF
```

We could link to https://github.com/lomigmegard/pekko-http-cors but there is a risk that @lomigmegard could delete that repo.

lomigmegard: > We could link to https://github.com/lomigmegard/pekko-http-cors but there is a risk that @lomigmegard could delete that repo.

I was thinking of just archiving it, thus any link will still work.

http-cors/src/main/scala/org/apache/pekko/http/cors/javadsl/settings/CorsSettings.scala
mdedetrich: So just to confirm, since the original code has no headers I need to add the standard Apache one?

http-cors/src/main/scala/org/apache/pekko/http/cors/scaladsl/settings/CorsSettings.scala
mdedetrich: Fixed in latest push

http-cors/src/main/scala/org/apache/pekko/http/cors/javadsl/CorsDirectives.scala
mdedetrich: Fixed in latest push

docs/src/main/paradox/migration-guide/index.md
mdedetrich: Fixed in latest push

docs/src/main/paradox/migration-guide/index.md
mdedetrich: Fixed in latest push

mdedetrich:
@lomigmegard I addressed all of your comments and sent a re-review but when going through the code I noticed some things.

http-cors/src/main/scala/org/apache/pekko/http/cors/javadsl/settings/CorsSettings.scala
```diff
@@ -0,0 +1,59 @@
...
+  /**
+   * Creates an instance of CorsSettings using the configuration provided by the given ActorSystem.
+   */
+  def create(system: ActorSystem): CorsSettings = scaladsl.settings.CorsSettings(system)
+
+  /**
+   * Settings from the default loaded configuration. Note that application code may want to use the `apply()` methods
+   * instead to have more control over the source of the configuration.
+   */
+  def defaultSettings: CorsSettings = scaladsl.settings.CorsSettings.defaultSettings
```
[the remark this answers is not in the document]
lomigmegard: This `defaultSettings` was kept for backward compatibility when I worked on the [version 1.0.0](https://github.com/lomigmegard/akka-http-cors/releases/tag/v1.0.0). It's indeed not idiomatic in akka/pekko-http. 

I think it would be fine removing it (but that require more migration notes), or simply deprecating it. Looking at the downloads from `akka-http-cors_2.13`, around 1% are coming from versions 0.x. 
Screenshot 2023-06-29 at 18 02 24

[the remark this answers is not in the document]
mdedetrich: Thanks I will deprecate it then, will wait for the CI to pass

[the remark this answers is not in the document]
mdedetrich: I have just pushed the deprecations, i.e.

```scala
@deprecated("Use other CorsSettings constructors", "Akka Http Cors 1.2.0")
```

Let me know if this message makes sense, if so I will resolve the thread.

jrudolph (commented):
This is a great addition. Thanks a lot, @lomigmegard, and @mdedetrich.

Has it been considered whether we should keep this as a separate module or merge into the http modules proper?

Separate module:
 - + keeps functionality separate and users can choose whether they need it
 - + precedence in http-caching to keep such functionality outside
 - - some extra imports needed to be used
 - - adds complexity to builds to keep the module on the same version
 - - might become difficult to merge into main modules later if we make a release once because of compatibility problems (though, we could still publish an empty jar to make eviction work)

Merged in:
 - + common enough functionality to warrant inclusion into main modules
 - + slightly easier consumption because directives are automatically in scope
 - + easier integration with potential "default setup" should we want to provide that
 - - even bigger, harder to maintain main modules

I'm currently leaning towards leaving it as is in this PR, because the work is already done and the downsides are relatively minor (though we are mostly pushing/keeping them to users).

http-cors/src/main/scala/org/apache/pekko/http/cors/javadsl/settings/CorsSettings.scala
pjfanning: could you use the full header from https://www.apache.org/legal/src-headers.html ?

mdedetrich:
@jrudolph 

> Has it been considered whether we should keep this as a separate module or merge into the http modules proper?

I did briefly think about it and I came to similar pros/cons as you did but I opted to put it as a separate module for the following reasons

* While its definitely true that cors is commonly used, I am not sure if its common enough to warrant inclusion into core mainly based on the fact that it wasn't included into core in the first place and it seems like it wasn't too much of a problem?
  * I was also copying http-caching here, which you could also claim is a fairly frequent problem in HTTP however is also its own separate module
* Its less friction/work rather than putting it into core

That being said if you feel strongly that it should go into core then its something that can be explored

@pjfanning I have fixed all of the legal/license/header stuff you mentioned before (turns out I entirely forgot that sbt-header will accept source files with standard apache header). Can you re-review it?

http-cors/src/main/scala/org/apache/pekko/http/cors/javadsl/settings/CorsSettings.scala
mdedetrich: I just updated the PR, I ended up using

```
/*
 * Licensed to the Apache Software Foundation (ASF) under one or more
 * license agreements; and to You under the Apache License, version 2.0:
 *
 *   https://www.apache.org/licenses/LICENSE-2.0
 */
 ```
 
 which is the same as the standard header but without the "derived from Akka" part

**Comment by pjfanning** [comment]
> @pjfanning I have fixed all of the legal/license/header stuff you mentioned before (turns out I entirely forgot that sbt-header will accept source files with standard apache header). Can you re-review it?

I think we should use the full header from https://www.apache.org/legal/src-headers.html

**Comment by mdedetrich** [comment]
> I think we should use the full header from https://www.apache.org/legal/src-headers.html

I noticed your comment after I pushed PR, fixing it now

**Comment by mdedetrich** [comment]
@pjfanning Just pushed with the full Apache license header

**Review by pjfanning — APPROVED** [review]
lgtm

**Comment by pjfanning** [comment]
@mdedetrich before we release, we probably need to raise an issue at https://issues.apache.org/jira/projects/LEGAL/issues to discuss how to acknowledge this code. Pertinent points are:

* We have a Software Grant Agreement on hand with the Secretary of the ASF
* https://github.com/lomigmegard/pekko-http-cors has Apache License, v2.0 with no custom copyright declaration
* pekko-http-cors has no NOTICE file
* pekko-http-cors source code has no license headers

My suggestion would be a short mention in our NOTICE file. Something a bit like the BEA mention in https://github.com/apache/poi/blob/trunk/legal/NOTICE

**Comment by mdedetrich** [comment]
> My suggestion would be a short mention in our NOTICE file. Something a bit like the BEA mention in https://github.com/apache/poi/blob/trunk/legal/NOTICE

@pjfanning  I just added a mention in the NOTICE file, if the text needs to be updated let me know.

**Comment by pjfanning** [comment]
@mdedetrich I think that NOTICE statement looks good

**Comment by mdedetrich** [comment]
Perfect I will leave this PR open till early next week incase @jrudolph @lomigmegard have any further additional comments, then I will merge it.

File: http-cors/src/main/scala/org/apache/pekko/http/cors/javadsl/settings/CorsSettings.scala (hunk 9 above)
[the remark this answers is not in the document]
Comment (lomigmegard) [reply http-cors/src/main/scala/org/apache/pekko/http/cors/javadsl/settings/CorsSettings.scala:76]: Looks good for the message. For the "since", would it make more sense to reference the future first release of pekko? `Pekko Http Cors 1.0.0`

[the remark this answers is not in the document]
Comment (mdedetrich) [reply http-cors/src/main/scala/org/apache/pekko/http/cors/javadsl/settings/CorsSettings.scala:76]: Ah right, nice pickup will fix this now.

File: http-cors/src/main/scala/org/apache/pekko/http/cors/javadsl/CorsDirectives.scala [hunk 10]
```diff
@@ -0,0 +1,50 @@
...
+import pekko.http.impl.util.JavaMapping
+
+object CorsDirectives {
+
+  import pekko.http.cors.scaladsl.{ CorsDirectives => D }
+
+  def cors(inner: Supplier[Route]): Route =
+    RouteAdapter {
+      D.cors() {
+        inner.get().delegate
```
lomigmegard, line 37: nit: choose between `get()` and `get` (line 44), for consistence.

http-cors/src/main/scala/org/apache/pekko/http/cors/scaladsl/model/HttpHeaderRange.scala
```diff
@@ -0,0 +1,39 @@
...
+abstract class HttpHeaderRange extends javadsl.model.HttpHeaderRange
+
+object HttpHeaderRange {
+  case object `*` extends HttpHeaderRange {
+    def matches(header: String) = true
+  }
+
+  final case class Default(headers: Seq[String]) extends HttpHeaderRange {
+    val lowercaseHeaders: Seq[String] = headers.map(_.toLowerCase(Locale.ROOT))
+    def matches(header: String): Boolean = lowercaseHeaders contains header.toLowerCase(Locale.ROOT)
```
lomigmegard, line 35: suggestion: use `org.apache.pekko.util.Helpers.toRootLowerCase` instead. Not sure there is a big advantage, just consistency (I would have not created this helper in the first place)

lomigmegard, line 35: suggestion: remove the infix notation, that was old me thinking it was cool.

http-cors/src/main/scala/org/apache/pekko/http/cors/scaladsl/model/HttpOriginMatcher.scala
```diff
@@ -0,0 +1,90 @@
...
+        if (hasWildcard(origin)) new WildcardHostMatcher(origin) else new StrictHostMatcher(origin)
+      }
+    }
+
+    override def matches(origin: HttpOrigin): Boolean = matchers.exists(_.apply(origin))
+    override def toString: String = origins.mkString(" ")
+  }
+
+  final case class Strict(origins: Seq[HttpOrigin]) extends HttpOriginMatcher {
+    override def matches(origin: HttpOrigin): Boolean = origins contains origin
```
lomigmegard, line 67: nit: remove infix notation here too

http-cors/src/main/scala/org/apache/pekko/http/cors/scaladsl/settings/CorsSettings.scala
```diff
@@ -0,0 +1,226 @@
...
+  override def getAllowCredentials: Boolean = this.allowCredentials
+  override def getAllowedOrigins: jmcorsmodel.HttpOriginMatcher = this.allowedOrigins
+  override def getAllowedHeaders: jmcorsmodel.HttpHeaderRange = this.allowedHeaders
+  override def getAllowedMethods: java.lang.Iterable[jmmodel.HttpMethod] =
+    (this.allowedMethods: Seq[pekko.http.javadsl.model.HttpMethod]).asJava
+  override def getExposedHeaders: java.lang.Iterable[String] = this.exposedHeaders.asJava
+  override def getMaxAge: OptionalLong = this.maxAge.toJavaPrimitive
+
+  // Currently the easiest way to go from Java models to their Scala equivalent is to cast.
+  // See https://github.com/akka/akka-http/issues/661 for a potential opening of the JavaMapping API.
```
lomigmegard, line 148: chore: this comment can be removed now.

lomigmegard (approved):
LGTM, just a few minor comments.

http-cors/src/main/scala/org/apache/pekko/http/cors/scaladsl/model/HttpHeaderRange.scala
```diff
@@ -0,0 +1,39 @@
...
+
+abstract class HttpHeaderRange extends javadsl.model.HttpHeaderRange
+
+object HttpHeaderRange {
+  case object `*` extends HttpHeaderRange {
+    def matches(header: String) = true
+  }
+
+  final case class Default(headers: Seq[String]) extends HttpHeaderRange {
+    val lowercaseHeaders: Seq[String] = headers.map(_.toLowerCase(Locale.ROOT))
```
mdedetrich, line 34: @lomigmegard Can this be private or is it expected to be used publicly?

http-cors/src/main/scala/org/apache/pekko/http/cors/javadsl/CorsDirectives.scala
mdedetrich: Fixed.

http-cors/src/main/scala/org/apache/pekko/http/cors/scaladsl/model/HttpHeaderRange.scala
mdedetrich: Fixed.

mdedetrich: Fixed.

http-cors/src/main/scala/org/apache/pekko/http/cors/scaladsl/model/HttpOriginMatcher.scala
mdedetrich: Fixed.

http-cors/src/main/scala/org/apache/pekko/http/cors/scaladsl/settings/CorsSettings.scala
mdedetrich: Fixed

http-cors/src/main/scala/org/apache/pekko/http/cors/scaladsl/model/HttpHeaderRange.scala
```diff
@@ -0,0 +1,39 @@
...
+
+abstract class HttpHeaderRange extends javadsl.model.HttpHeaderRange
+
+object HttpHeaderRange {
+  case object `*` extends HttpHeaderRange {
+    def matches(header: String) = true
+  }
+
+  final case class Default(headers: Seq[String]) extends HttpHeaderRange {
+    val lowercaseHeaders: Seq[String] = headers.map(Helpers.toRootLowerCase)
```
mdedetrich, line 34: @lomigmegard Can this be private or is it expected to be used publicly?

lomigmegard: It can be private, it's an implementation detail.

mdedetrich:
Another thing I noticed is that we happen to have public `final case class`, this should be converted to a normal `final class` (with manually implemented `toString`/`equals` etc etc) because adding fields to a `case class` breaks binary compatibility and at least from some quick searching pekko-http doesn't any public `final case class`.

@jrudolph @lomigmegard wdyt?

http-cors/src/main/scala/org/apache/pekko/http/cors/scaladsl/model/HttpHeaderRange.scala
mdedetrich: Thanks, updated.

pjfanning:
> Another thing I noticed is that we happen to have public `final case class`, this should be converted to a normal `final class` (with manually implemented `toString`/`equals`,`apply` etc etc) because adding fields to a `case class` breaks binary compatibility and at least from some quick searching pekko-http doesn't any public `final case class`.
> 
> @jrudolph @lomigmegard wdyt?

I wouldn't worry about binary compatibility yet - we have no releases to be binary compatible with. I would favour using case classes over classes with custom toString/hashCode/equals/etc.

mdedetrich:
> I wouldn't worry about binary compatibility yet - we have no releases to be binary compatible with. I would favour using case classes over classes with custom toString/hashCode/equals/etc.

Its not about what people are currently using, its about any potential future changes. If you have a `case class` like this

```scala
final case class MyCaseClass(string: String)
```

And you add a field some point in the future, i.e.
```scala
final case class MyCaseClass(string: String, int: Int)
```

This will break binary compatibility for `MyCaseClass because of `unapply`/`copy` (which `case class` automatically implements). This is why for both pekko and pekko-http no public classes happen to be a `case class`. The same is of other Scala ecosystems/projefts (i.e. cats) which actually follow bincompat properly. If we want to follow semantic versioning, unless we can always guarantee that we will never add a new field it should be converted to a class.

This is a known problem and there is a massive thread on Scala contributors about it https://contributors.scala-lang.org/t/pre-sip-structural-data-structures-that-can-evolve-in-a-binary-compatible-way/5684

lomigmegard:
I agree about to be careful on the binary compatibility. However in the case of this library, all the occurrences of `case class` are wrappers around a single value, which semantically would make no sense to extend to more fields. For example the rejection, as all other rejections in pekko-http are a `case class`: 

```
final case class CorsRejection(cause: CorsRejection.Cause)
```

`CorsSettingsImpl` is an exception but it's private and we could/should make use of the workaround of the private constructor as shown by M. Odersky [here](https://github.com/scala/improvement-proposals/pull/50#issuecomment-1346210753). We could also use this trick in other places, limiting the boilerplate code (I don't think it's worth implementing equals/hashCode by hand).

mdedetrich:
> I agree about to be careful on the binary compatibility. However in the case of this library, all the occurrences of `case class` are wrappers around a single value, which semantically would make no sense to extend to more fields. For example the rejection, as all other rejections in pekko-http are a `case class`:
> 
> ```scala
> final case class CorsRejection(cause: CorsRejection.Cause)
> ```

Ah yes, I completely forgot that in this case pekko/akka-http uses `case class`es more like a new type in order for its routing dsl rather than a data structure

> `CorsSettingsImpl` is an exception but it's private and we could/should make use of the workaround of the private constructor as shown by M. Odersky [here](https://github.com/scala/improvement-proposals/pull/50#issuecomment-1346210753).

Since its `private` I don't think its necessary in this case unless I am missing something

> We could also use this trick in other places, limiting the boilerplate code (I don't think it's worth implementing equals/hashCode by hand).

Yes, I completely forgot about this technique and there are definitely places where this needs to be done, have a look at https://github.com/apache/incubator-pekko-connectors/blob/main/s3/src/main/scala/org/apache/pekko/stream/connectors/s3/model.scala !

mdedetrich:
@lomigmegard FYI on this topic of `case class` binary compatibility, the technique you mentioned (i.e. https://github.com/scala/improvement-proposals/pull/50#issuecomment-1346210753) only works with Scala 3. With Scala 2 even if you have a `private` constructor in the `case class` the generated `copy` method is still public (which is one of the problematic methods that breaks bincompat when more fields are added). It seems that this when changed as an improvement in Scala 3, see https://github.com/apache/incubator-pekko-connectors/issues/190#issuecomment-1616514860 for more info.

mdedetrich:
Im going to go ahead and merge this now. Many thanks @lomigmegard for your help on the review!

PR #210, opened 2023-06-30
upgrade paradox plugin

pjfanning:
* this update includes a fix to stop us using googlefonts - which ASF wants us to avoid due to Google's habit of tracking users


He-Pin (approved):
lgtm

Issue #213, opened 2023-06-30
fix links to pekko-http snapshot javadoc/scaladoc


pjfanning:
The current nightly builds publish docs but the links to the javadoc/scaladoc from the reference docs uses the full snapshot version. We do not publish to https://nightlies.apache.org/pekko/docs/pekko-http/main-snapshot/ with the full snapshot version.

Example: See `HttpRequest` link in https://pekko.apache.org/docs/pekko-http/current/introduction.html#low-level-http-server-apis

Today that links to https://pekko.apache.org/japi/pekko-http/0.0.0+4453-9a48d9fc-SNAPSHOT/org/apache/pekko/http/scaladsl/model/HttpRequest.html

For snapshot builds, can we change the link just to have `snapshot` instead of the full snapshot version?

So the link above would be https://pekko.apache.org/japi/pekko-http/snapshot/org/apache/pekko/http/scaladsl/model/HttpRequest.html - which works

pjfanning:
fyi @jrudolph @mdedetrich

mdedetrich:
While its true that in Akka's case, for the latest release branch it does maintain different patch versions (i.e. https://doc.akka.io/docs/akka/2.8.1/ vs https://doc.akka.io/docs/akka/2.8.2/) there is also https://doc.akka.io/docs/akka/2.8/ which points to the latest version (i.e. https://doc.akka.io/docs/akka/2.8.2 with my previous example), note that this is **NOT** the same as the docs pointing to the latest point of the 2.8 branch.

So there are 2 ways to solve this problem, one is to find a way so that we can point `1.0` to the latest released version (which will be `1.0.0`) or we do as you suggested and change to using patch versions.

pjfanning:
> While its true that in Akka's case, for the latest release branch it does maintain different patch versions (i.e. https://doc.akka.io/docs/akka/2.8.1/ vs https://doc.akka.io/docs/akka/2.8.2/) there is also https://doc.akka.io/docs/akka/2.8/ which points to the latest version (i.e. https://doc.akka.io/docs/akka/2.8.2 with my previous example), note that this is **NOT** the same as the docs pointing to the latest point of the 2.8 branch.
> 
> So there are 2 ways to solve this problem, one is to find a way so that we can point `1.0` to the latest released version (which will be `1.0.0`) or we do as you suggested and change to using patch versions.

@mdedetrich was this comment intended for https://github.com/apache/incubator-pekko-http/issues/212 ?

[...]

PR #214, opened 2023-06-30
use generic snapshot link for snapshot scaladoc/javadoc

pjfanning:
relates to #213

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -429,11 +429,11 @@ lazy val docs = project("docs")
       "scaladoc.org.apache.pekko.base_url" -> s"https://pekko.apache.org/api/pekko/${PekkoDependency.docs.link}",
       "scaladoc.org.apache.pekko.link_style" -> "direct",
       "javadoc.org.apache.pekko.http.base_url" -> {
-        val v = if (!isSnapshot.value) "current" else version.value
+        val v = if (!isSnapshot.value) "current" else "snapshot"
         s"https://pekko.apache.org/api/pekko-http/$v"
       },
       "scaladoc.org.apache.pekko.http.base_url" -> {
-        val v = if (!isSnapshot.value) "current" else version.value
+        val v = if (!isSnapshot.value) "current" else "snapshot"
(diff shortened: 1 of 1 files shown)


mdedetrich (changes requested):
Would it be possible to do this properly and use `projectInfoVersion` so that all versions referenced in docs are properly aligned, i.e. what was done in this PR https://github.com/apache/incubator-pekko/pull/392/files#diff-82e8a219706e01e038685e70db53a3e5c7de3d6648730ca5b632f91004bb0154R136 ?

In other words, set `projectInfoVersion` to `if (!isSnapshot.value) "current" else "snapshot"` and then use `projectInfoVersion` elsewhere for doc links since `projectInfoVersion` is also used within paradox/project-info.conf

pjfanning:
> Would it be possible to do this properly and use `projectInfoVersion` so that all versions referenced in docs are properly aligned, i.e. what was done in this PR https://github.com/apache/incubator-pekko/pull/392/files#diff-82e8a219706e01e038685e70db53a3e5c7de3d6648730ca5b632f91004bb0154R136 ?
> 
> In other words, set `projectInfoVersion` to `if (!isSnapshot.value) "current" else "snapshot"` and then use `projectInfoVersion` elsewhere for doc links since `projectInfoVersion` is also used within paradox/project-info.conf

I've made a change like the one you suggested

build.sbt
```diff
@@ -415,6 +415,7 @@ lazy val docs = project("docs")
       10.seconds
     },
     paradoxGroups := Map("Language" -> Seq("Scala", "Java")),
+    projectInfoVersion := (if (isSnapshot.value) "snapshot" else version.value),
```
mdedetrich, line 418: This should be placed in `inThisBuild` so that it applies (and sets) every sbt module version correctly. Around https://github.com/apache/incubator-pekko-http/blob/9e8e1e79cf27eca794220eb8d53152e30aebec1f/build.sbt#L45 is the best spot.

mdedetrich (approved):
lgtm

Issue #212, opened 2023-06-30
fix links to pekko javadoc/scaladoc


pjfanning:
pekko-http build uses `.link` to get the version number to use scaladoc/javadoc links 
https://github.com/apache/incubator-pekko-http/blob/main/build.sbt#L429

`.link` provides the `x.y` version as opposed to `x.y.z` version

At the moment, for Pekko, we only publish `1.0.0` but these links point to `1.0`.

Example: see the `Flow[HttpRequest, HttpResponse, _]` link in https://pekko.apache.org/docs/pekko-http/current/introduction.html#low-level-http-server-apis

* we could change the pekko-http build to use the `x.y.z` version in these links
* we could update `.htaccess` to redirect `1.0` to `1.0.0` (but this has a big maintenance overhead, we'd have to change the `.htaccess` after every release.

mdedetrich:
While its true that in Akka's case, for the latest release branch it does maintain different patch versions (i.e. https://doc.akka.io/docs/akka/2.8.1/ vs https://doc.akka.io/docs/akka/2.8.2/) there is also https://doc.akka.io/docs/akka/2.8/ which points to the latest version (i.e. https://doc.akka.io/docs/akka/2.8.2 with my previous example), note that this is **NOT** the same as the docs pointing to the latest point of the 2.8 branch.

So there are 2 ways to solve this problem, one is to find a way so that we can point `1.0` to the latest released version (which will be `1.0.0`) or we do as you suggested and change to using patch versions. I have a preference for the former.

mdedetrich:
@pjfanning I think this can be closed?

PR #215, opened 2023-07-01
use x.y.z version in scaladoc/javadoc links

pjfanning:
relates to #212


mdedetrich:
@pjfanning I just created https://github.com/apache/incubator-pekko/pull/461, I think this is a better overall solution?

pjfanning:
sure - we'll go with the 461 solution

PR #226, opened 2023-07-12
use standard license on new files

pjfanning:
relates to https://github.com/apache/incubator-pekko/pull/474

### DIFF
--- a/project/AddMetaInfLicenseFiles.scala
+++ b/project/AddMetaInfLicenseFiles.scala
@@ -1,10 +1,18 @@
 /*
  * Licensed to the Apache Software Foundation (ASF) under one or more
- * license agreements; and to You under the Apache License, version 2.0:
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
  *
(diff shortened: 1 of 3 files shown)


nvollmar (approved):
lgtm

project/AddMetaInfLicenseFiles.scala
```diff
@@ -1,10 +1,18 @@
...
  * Licensed to the Apache Software Foundation (ASF) under one or more
- * license agreements; and to You under the Apache License, version 2.0:
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
  *
- *   https://www.apache.org/licenses/LICENSE-2.0
+ *    http://www.apache.org/licenses/LICENSE-2.0
```
jrudolph, line 9: Was that intended?

pjfanning: * the standard ASL header uses http
* see https://www.apache.org/legal/src-headers.html

mdedetrich: Also note that it redirects to https, but yes we should be using the exact same text that is references on the official ASF website (https or not).

jrudolph: :+1: thanks for clarifying.

PR #228, opened 2023-07-13
add 1.0.0 release notes

pjfanning:
may not be the final version - but I'd like to link to these notes instead of the [cwiki draft notes](https://cwiki.apache.org/confluence/display/PEKKO/Pekko+HTTP+Release+Notes) in our emails about RC votes and release announcements


jrudolph (approved):
Great, good to have it here as well! lgtm

Issue #216, opened 2023-07-01
scaladoc links to java runtime classes are broken


pjfanning:
see https://github.com/apache/incubator-pekko-http/actions/runs/5427033437/jobs/9878610786

has issues like:
```
`[https://docs.oracle.com/en/java/javase/11/docs/api/java/util/Map.html`](https://docs.oracle.com/en/java/javase/11/docs/api/java/util/Map.html) status 404 Not Found
 - ../../../../../target/scala-2.13/unidoc/org/apache/pekko/http/javadsl/model/headers/HttpCredentials.html
 - ../../../../../target/scala-2.13/unidoc/org/apache/pekko/http/javadsl/server/AllDirectives.html
 - ... 3 more
```

should be https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Map.html

The sbt settings closely match those in incubator-pekko but the scaladoc is generating ok in incubator-pekko

mdedetrich:
@pjfanning So I think the core issue is this https://github.com/ThoughtWorksInc/sbt-api-mappings/blob/5b1e4f11647421d287d3a26ed7fb077474ea7a3e/src/main/scala/com/thoughtworks/sbtApiMappings/BootstrapApiMappings.scala#L37-L43 although I am still getting the issue when running with JDK 11, investigating further

mdedetrich:
Re-opening since its still broken

pjfanning:
@mdedetrich is this safe to close? - we have the link-validator job running in CI

mdedetrich:
Yes

jrudolph:
No, the link validator is still broken.

mdedetrich:
@jrudolph Huh? https://github.com/apache/incubator-pekko-http/actions/runs/5643524801

jrudolph:
Last time it ran, it failed, and then removed its settings accidentally...

https://github.com/apache/incubator-pekko-http/actions/runs/5643524801/job/15285543470#step:8:9

```
can't read [scripts/link-validator.conf]
```

It seems to fail silently.

mdedetrich:
Oh I see, let me rename issue so it's clearer then

jrudolph:
This issue is probably still existing, and the other is another issue which masks this.

mdedetrich:
> This issue is probably still existing, and the other is another issue which masks this.

I remember fixing/verifying the broken links, so unless some change was introduced which broke the links it should be fine.

[...]

PR #231, opened 2023-07-17
Fix broken generated java unidoc mappings

mdedetrich:
Resolves: https://github.com/apache/incubator-pekko-http/issues/216

The code which generates the api mappings for documentation originates from sbt-api-mappings. sbt-api-mappings tries to calculate the JDK version and if the calculated version is less than 11 it uses the broken reference as described in the issue, i.e.

https://github.com/ThoughtWorksInc/sbt-api-mappings/blob/5b1e4f11647421d287d3a26ed7fb077474ea7a3e/src/main/scala/com/thoughtworks/sbtApiMappings/BootstrapApiMappings.scala#L30-L44

Setting JDK 11 for publishing docs should be enough and its likely this is one of the reason why pekko-http core requires JDK 11 specifically for their documentation generation as well.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -45,11 +45,11 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Set up JDK 8
+      - name: Set up JDK 11
         uses: actions/setup-java@v3
         with:
           distribution: temurin
-          java-version: 8
+          java-version: 11
 
       - name: Cache Coursier cache
         uses: coursier/cache-action@v6


.github/workflows/publish.yml
```diff
@@ -45,11 +45,11 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Set up JDK 8
+      - name: Set up JDK 11
         uses: actions/setup-java@v3
         with:
           distribution: temurin
-          java-version: 8
+          java-version: 11
```
pjfanning, line 52: should we change the java version for the jar publish too?

mdedetrich: It should be safe due to https://github.com/apache/incubator-pekko-http/pull/187 but lets do this in a separate PR because I want to validate on my machine that the bytecode is exactly the same.

pjfanning (approved):
lgtm

PR #234, opened 2023-07-18
Update latest pekko-sbt-paradox

mdedetrich:
This will bring in the newly updated incubator logo which we were asked to use


pjfanning (approved):
lgtm

PR #235, opened 2023-07-18
[NOT FOR MERGE] test with 1.0.1-RC1

pjfanning:


build.sbt
```diff
@@ -30,6 +30,7 @@ inThisBuild(Def.settings(
   scmInfo := Some(
     ScmInfo(url("https://github.com/apache/incubator-pekko-http"), "git@github.com:apache/incubator-pekko-http.git")),
   description := "Apache Pekko Http: Modern, fast, asynchronous, streaming-first HTTP server and client.",
+  resolvers += "Apache Pekko Staging".at("https://repository.apache.org/content/groups/staging"),
```
He-Pin, line 33: seems like we will need this always during the RC progress?

pjfanning: that is where we put the staging jar files for - the jars are only there during the staging process - when they are released or dropped, they are no longer in staging

PR #236, opened 2023-07-18
update build details

pjfanning:
* turns out we need to use java 8 to do the doc build
* mention `sbt testQuick` like incubator-pekko docs do


mdedetrich (approved):
lgtm

PR #238, opened 2023-07-19
build: remove autobahn-jenkins scripts

jrudolph:
Those are not relevant any more. If we want to run these on Github Actions, it will need a completely new setup.


mdedetrich:
> Those are not relevant any more. If we want to run these on Github Actions, it will need a completely new setup.

Can we make an issue on re-implementing them on Github Actions so we don't forget?

mdedetrich (approved):
lgtm

jrudolph:
> > Those are not relevant any more. If we want to run these on Github Actions, it will need a completely new setup.
> 
> Can we make an issue on re-implementing them on Github Actions so we don't forget?

Not sure we still need them, the WS implementation is mature since years. Created #247 to track the idea to add something back.

PR #239, opened 2023-07-19
build: explain the existence of extra-libs.zip

jrudolph:


http-tests/src/test/extra-libs.zip.README
```diff
@@ -0,0 +1,24 @@
...
+
+Archive:  i have späces.jar
+  Length      Date    Time    Name
+---------  ---------- -----   ----
+        0  2015-07-02 15:11   META-INF/
+       68  2015-07-02 15:11   META-INF/MANIFEST.MF
+       19  2015-07-02 15:11   test-resource.txt
+---------                     -------
+       87                     3 files
+```
```
mdedetrich, line 24: nit: newline

mdedetrich (approved):
lgtm

pjfanning:
Are we sure that will keep everyone happy? I have my doubts. Can we consider removing the zip and building it at test run time?

jrudolph:
I would like to avoid building it at build time because then we might not uncover the encoding issues. The whole point is to have an example that we know has been built correctly and then test that this works while testing/building.

pjfanning:
I'm happy to have the file but I'm not the one blocking the release

justinmclean:
You can leave it there but there is a chance some IPMC member might vote -1 as they consider it a binary file in a source release. It doesn't contain compiled code so I think it is OK - but you need to look inside it to find that out.

mdedetrich:
> You can leave it there but there is a chance some IPMC member might vote -1 as they consider it a binary file in a source release. It doesn't contain compiled code so I think it is OK - but you need to look inside it to find that out.

There is no binary code contained in `extra-spaces.zip`. It contains a single file called `i have späces.jar` and that jar itself only contains metadata + a text file, i.e.
```
<@incubator-pekko-http/h/s/test>-<⎇ update-latest-pekko-sbt-paradox>-<*>-> unzip i\ have\ späces.jar 
Archive:  i have späces.jar
   creating: META-INF/
  inflating: META-INF/MANIFEST.MF    
  inflating: test-resource.txt   
  ```

PR #240, opened 2023-07-19
fix license headers for boilerplate templates

jrudolph:


mdedetrich (approved):
lgtm

mdedetrich:
So we probably should modify sbt-header so that it works with `*.template` files, will make an issue on this

PR #241, opened 2023-07-19
build: remove old build scripts

jrudolph:
They are not used anywhere afaics and were missing appropriate license headers.


mdedetrich (approved):
lgtm

PR #242, opened 2023-07-19
docs: remove left-over javascript file

jrudolph:
It was not used any more, but more importantly it does not seem to be necessary any more, as the underlying bug in Firefox seems to be fixed by now.

See https://github.com/lightbend/paradox/issues/443 for a history of the problem.


mdedetrich (approved):
lgtm

PR #245, opened 2023-07-19
build: also aggregate httpScalafix, so headerCheck will also work on those files

jrudolph:
Should fail until #243 is merged


mdedetrich (approved):
lgtm

jrudolph:
Also discovered a bug in the test itself and also a bug in sbt-scalafix...

https://github.com/scalacenter/scalafix/issues/1828

He-Pin (approved):
lgtm

jrudolph:
fyi. this is still failing because the `validatePullRequest` task has problems aggregating correctly over modules that have no Scala 3 support and thus fails when encountering the scalafix modules.

mdedetrich:
This is something that's been on the top of my head and probably should be discussed in general, but I honestly think that validatePullRequest should be removed because its overly complex and most of the problems that it catches are already solved by other mechanisms (i.e. strict github checks) which likely didn't even exist when it was created.

jrudolph:
It's not about the problems it solves but about how much time is saved by preventing to execute extra tests.

But I agree about the complexity, so if we can simplify without losing the ability to run a reduced set of tests, than that would be nice. Might be that most complexity is for the extra reporting which is basically related to reporting a good summary on Github while we were still running on Jenkins.

pjfanning:
@jrudolph @mdedetrich do we think this is needed for 1.0.0-RC2?

Other than upgrading to Pekko (Core) 1.0.1 when it is released, I don't think we need any more changes for Pekko HTTP 1.0.0-RC2.

jrudolph:
> @jrudolph @mdedetrich do we think this is needed for 1.0.0-RC2?

No, this is just an improvement to avoid overlooking issues in those modules in the future.

PR #246, opened 2023-07-19
add extra license header to DateTime.scala corresponding to note in LICENSE

jrudolph:


mdedetrich (approved):
lgtm

justinmclean:
I would assume as this is 3rd party code it's not copyright Lightbend Inc or the ASF??

mdedetrich:
> I would assume as this is 3rd party code it's not copyright Lightbend Inc or the ASF??

Its code that was written by Lightbend that was based off of ASF code

mdedetrich:
This was also clearly decided upon in https://issues.apache.org/jira/browse/LEGAL-626, we are not allowed to remove the existing Lightbend copyright header from code that was taken from Akka (which is the case here).

justinmclean:
Without permission no, so probably best to leave it there, it's unfortunate they copies someone else code.

PR #252, opened 2023-07-19
test: avoid frequent timeouts in ResponseParserSpec

jrudolph:
Seems to have happened quite often recently (mostly seen on Scala 3.3 Java 8 runs, not sure if that's something to investigate but likely not).


mdedetrich (approved):
lgtm

Issue #250, opened 2023-07-19
acknowledge Heiko Seeburger license


pjfanning:
ASL headers with Heiko Seeburger copyright

```
29. ./http-tests/src/test/java/org/apache/pekko/http/javadsl/marshalling/sse/EventStreamMarshallingTest.java
30. ./http-tests/src/test/java/org/apache/pekko/http/javadsl/unmarshalling/sse/EventStreamUnmarshallingTest.java
31. ./http-core/src/test/java/org/apache/pekko/http/javadsl/model/sse/ServerSentEventTest.java
32. ./http-core/src/main/java/org/apache/pekko/http/javadsl/model/sse/ServerSentEvent.java
33. ./http-core/src/main/java/org/apache/pekko/http/javadsl/model/headers/LastEventId.java
```

pjfanning:
@jrudolph any idea where these files originate? Unfortunately, we need to track down the original files so that we can see if the project has a LICENSE or NOTICE that we need to copy too.

pjfanning:
Could they be from https://github.com/hseeberger/akka-http-json ? The copyright matches but I'm not sure if all the files come from there.

pjfanning:
I think this is where they are from - https://github.com/hseeberger/akka-sse

jrudolph:
This has probably been an oversight when akka-sse has been merged in. In any case, it was @hseeberger himself who proposed the merge (in https://github.com/akka/akka-http/pull/1116), so IMO we can take it as is. Let's add a section to LICENSE to acknowledge that some code files have his copyright note on top.

hseeberger:
We have not changed the headers? D'oh!
Of course I am cool with changing the headers in akka-http and elsewhere, no need to mention my copyright/license/anything.

pjfanning:
> We have not changed the headers? D'oh! Of course I am cool with changing the headers in akka-http and elsewhere, no need to mention my copyright/license/anything.

Thanks @hseeberger - it's easier for us to just keep the headers as is. What we are missing is adding something to our LICENSE. I am working on that now.

Changing the headers would require a whole extra level of ASF bureaucracy.

hseeberger:
Hehe, then my name shall not be forgotten ;-)

PR #253, opened 2023-07-19
update license and notice with akka-see license/notice

pjfanning:
* relates to #250 
* http-tests module is not published as a jar, so no need to worry about getting a tailored LICENSE/NOTICE included in its jar

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -279,6 +279,19 @@ pekko-http-core contains code based on code by Tim Kienzle with his permission t
 
 ---------------
 
+pekko-http-core contains code based on akka-sse <https://github.com/hseeberger/akka-sse>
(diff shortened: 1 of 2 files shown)


jrudolph:
Do we have to change/add our license headers as well?

He-Pin:
@hseeberger Ping~

pjfanning:
@kerr this is discussed in https://github.com/apache/incubator-pekko-http/issues/250 -- Heiko is aware of it

People may not trust me and you can ignore what I say - changing a license header is going to be more complicated than the change I have in this PR.

mdedetrich:
I agree with @pjfanning here, changing an existing header is far more problematic

mdedetrich (approved):
lgtm

PR #256, opened 2023-07-19
add CORS specific notice to pekko-http-cors jar

pjfanning:
* build currently adds the short PekkoNotice to the pekko-http-cors and it would be better to include the acknowledgement of Lomi Mégard's code donation 
* the big NOTICE in project root can still serve as the pekko-http-core jar because almost the entries relate to that jar - not that much point having a slightly truncated copy to put in pekko-http-core jar


project/AddMetaInfLicenseFiles.scala
```diff
@@ -41,6 +41,13 @@ object AddMetaInfLicenseFiles extends AutoPlugin {
     apacheSonatypeLicenseFile := baseDir.value / "LICENSE",
     apacheSonatypeNoticeFile := baseDir.value / "NOTICE")
 
+  /**
+   * Settings specific for Pekko http-cors subproject which require a different license file.
+   */
+  lazy val httpCorsSettings = Seq(
+    apacheSonatypeLicenseFile := baseDir.value / "legal" / "StandardLicense.txt",
+    apacheSonatypeNoticeFile := baseDir.value / "legal" / "CorsNotice.txt")
```
He-Pin, line 49: does it need a `DISCLAIMER` too?

pjfanning: this is inherited from the projectSetting - that adds the DISCLAIMER to all jars - it is the same DISCLAIMER in all jars but the LICENSE and NOTICE can be specific to a jar

mdedetrich: Yes, by default https://github.com/mdedetrich/sbt-apache-sonatype will add the DISCLAIMER into `projectSettings` since its a mandatory for ASF projects.

mdedetrich (approved):
lgtm

PR #257, opened 2023-07-21
add parboiled2 license

pjfanning:
Base64Parsing is from parboiled2.

https://github.com/apache/incubator-pekko-http/blob/38ff8037a5677a47a652330af407e96cc78f98e7/http-core/src/main/scala/org/apache/pekko/http/impl/model/parser/Base64Parsing.scala

We have removed a lot of parboiled2 code from pekko-http but Base64Parsing is still there - although only some of its code is used.

[parboiled2](https://github.com/sirthias/parboiled2) has no NOTICE file.

relates to #249

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -279,6 +279,16 @@ pekko-http-core contains code based on code by Tim Kienzle with his permission t
 
 ---------------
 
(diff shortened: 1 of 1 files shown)


He-Pin (approved):
lgtm

PR #259, opened 2023-07-23
Remove Base64Parsing

pjfanning:
* use parboiled class instead (we have removed all the other parboiled code, now that we have a jar dependency for it)
* the code that I moved to Http2SettingsHeader did not come from the parboiled Base64Parsing class
* relates to #257


http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/Http2SettingsHeader.scala
```diff
@@ -42,4 +48,5 @@ private[pekko] object Http2SettingsHeader {
     val reader = new io.ByteStringParser.ByteReader(bytes)
     Http2FrameParsing.readSettings(reader, log)
   }
+
```
He-Pin, line 51: remove this new line

LICENSE
```diff
@@ -279,16 +279,6 @@ pekko-http-core contains code based on code by Tim Kienzle with his permission t
...
 
-pekko-http-core contain code from parboiled2 <https://github.com/sirthias/parboiled2>
-distributed under the Apache 2.0 license.
-
-  - http-core/src/main/scala/org/apache/pekko/http/impl/model/parser/Base64Parsing.scala
-
-Copyright © 2009-2013 Mathias Doenitz <http://parboiled2.org>
-Copyright © 2013 Alexander Myltsev
-
----------------
```
pjfanning, line 290: only added yesterday - but not needed if we remove Base64Parsing.scala

http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/Http2SettingsHeader.scala
He-Pin: Does the removed method been used anywhere ,eg in grpc module?

pjfanning: it's package private code so it would be really bad practice to use it outside of this module

pjfanning: I've just checked incubator-pekko-grpc, incubator-pekko-connectors and incububator-pekko-management git repos and can't find any uses of the class that I am removing in this PR.

pjfanning: I've committed a change that removes the new line

He-Pin (approved):
LGTM

PR #261, opened 2023-07-23
update list of 3rd party jars

pjfanning:
* pekko-http-cors is now part of Apache Pekko


He-Pin (approved):
lgtm ,but better grouped with different catageories.

mdedetrich (approved):
lgtm

Issue #262, opened 2023-07-23
Update h2spec


lomigmegard:
Currently tests are using version `1.5.0`, which is from 2016. 

My current machine (Apple M1) Seg Fault on it (see https://github.com/apache/incubator-pekko-http/issues/218#issuecomment-1646844106), making the `H2SpecIntegrationSpec` test suite fail.

```
./http2-tests/target/h2spec_darwin_amd64/h2spec --dryrun
[1]    3328 segmentation fault  ./http2-tests/target/h2spec_darwin_amd64/h2spec --dryrun
```

Updating h2spec to its [latest version](https://github.com/summerwind/h2spec/releases) `2.6.0` would hopefully fix the issue. 

Note this is non-critical as the current version is working well in the CI environment.

pjfanning:
I'm not sure if h2spec supports Mac Silicon.

There only appear to be downloads for Mac AMD64 - https://github.com/summerwind/h2spec/releases/

lomigmegard:
Indeed, no binaries for arm64/aarch64. 

I still think this issue makes sense to eventually update to a more recent version.

lomigmegard:
The update to 2.6.0 makes the new `h2spec` binary work on Apple Silicon, at least transparently using Rosetta (https://github.com/apache/incubator-pekko-http/issues/218#issuecomment-1648476214)

jrudolph:
> The update to 2.6.0 makes the new `h2spec` binary work on Apple Silicon, at least transparently using Rosetta ([#218 (comment)](https://github.com/apache/incubator-pekko-http/issues/218#issuecomment-1648476214))

Great to know. I guess Rosetta hit a corner case in the old release.

PR #263, opened 2023-07-23
try h2spec 2.6.0

pjfanning:
* relates to #262 
* the h2spec v2 GitHub release files are now `.tar.gz` -- except the Windows one which is `.zip`
* the h2spec v2 GitHub release artifacts have no directories embedded in them
* the `h2spec` command no longer supports the `-s` option so that causes some test failures
    * needs investigation

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -11,7 +11,7 @@ import org.apache.pekko
 import pekko._
 import pekko.ValidatePullRequest._
 import PekkoDependency._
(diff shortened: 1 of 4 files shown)


jrudolph:
Fixed the test problems in #266.

The problem was, that with the latest version of h2spec, it seems selecting a group executed all examples even in subgroups while the previous version seemed to either select fewer examples or new examples were added in the meantime.

PR #265, opened 2023-07-24
job to run tests on JDK20

pjfanning:
* also set CI permissions on jobs that have no setting


jrudolph (changes requested):
I'd prefer to use the existing nightly build for that purpose and just add 20 to the matrix.

pjfanning:
> I'd prefer to use the existing nightly build for that purpose and just add 20 to the matrix.

Idea would be to use this until we get JDK20 build working consistently

jrudolph:
Ok, I see. We should make sure only to run this in nightlies or the build matrix will get really large which is inefficient and might have repercussions on the Github CI capacity.

jrudolph:
(But let's test it here with "Validate and test" to see the current state)

jrudolph:
We are already running 17 in nightlies but only there for the same reasons.

pjfanning:
> We are already running 17 in nightlies but only there for the same reasons.

Good point - nightly job is a better place to test with the extra JDK

.github/workflows/publish.yml
```diff
@@ -1,5 +1,8 @@
 name: Publish nightly snapshots and documentation
 
+permissions:
+  contents: read
```
jrudolph, line 4: Afaiu, giving no permissions is equivalent to setting all permissions to `none`. I.e. we are giving more permissions here than before.

.github/workflows/nightly.yml
```diff
@@ -1,5 +1,7 @@
 name: Nightly Builds
 
+permissions: {}
```
jrudolph, line 3: Isn't this just syntactically a no-op, because not specifying the element is the same as setting it as empty. According to https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs#defining-access-for-the-github_token-scopes:

> all of those that are not specified are set to `none`.

so not sure, if should or will change anything?

jrudolph (approved):
Adding 20 lgtm, not sure whether we should or need to change permissions.

.github/workflows/nightly.yml
pjfanning: If you don't set this, all permissions are available. As an example, we don't set permissions in publish.yml (before this PR set them) and that job reads secrets.

pjfanning: I think you're right. I tried it out in one of my projects and the permissions did not seem to affect secrets access.

PR #267, opened 2023-07-24
Revert old link-validator.conf accidentally deleted in #241

jrudolph:
Refs #216


mdedetrich (approved):
lgtm, the check passed so all links seem to work

jrudolph:
Yep, also getting some output this time. There's also an overflow error, so not sure if everything is tested comprehensively but let's assume for now.

PR #277, opened 2023-07-28
Update sbt-reproducible-builds and set check resolver to staging

mdedetrich:
So its easier to verify future builds


pjfanning (approved):
lgtm

PR #295, opened 2023-08-07
fix snapshot check

pjfanning:
problems with #294 
* I added a stray `"` into the regex
* the Group match has to handle 3 extra subgroups that the regex change brings in
  * for `1.1.0-M0+22-4f570ea2-SNAPSHOT`, the 3 extra subgroups are `-M0`, `-M`, `0`
  * if there is no `-M0` (or `-RC0`), the extra subgroups are all null
* the sorting will do for now but it doesn't take into account whether the tag is M or RC and there are other problems
* I did play around with sbt.librarymanagement.VersionNumber but it doesn't have a comparator (or scala.Ordering implementation).
  * I will try this again to see if I can create my own comparator for sbt.librarymanagement.VersionNumber
  * using this PR to get the build working for now
* I have tested these changes on my laptop

### DIFF
--- a/project/PekkoDependency.scala
+++ b/project/PekkoDependency.scala
@@ -99,7 +99,7 @@ object PekkoDependency {
     import scala.concurrent.Await
     import scala.concurrent.duration._
 
-    val snapshotVersionR = """href=".*/((\d+)\.(\d+)\.(\d+)(-(M|RC)(\d+))?\+(\d+)-[0-9a-f]+-SNAPSHOT)/""""".r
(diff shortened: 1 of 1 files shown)


mdedetrich (approved):
lgtm

mdedetrich:
> I will try this again to see if I can create my own comparator for sbt.librarymanagement.VersionNumber

Having an ordering number for a `VersionNumber` doesn't make sense in a library like sbt because its completely bespoke, different orgs have the option of making their own release strategies.

In any case I would recommend how we approach this problem in general because there are issues in just using the latest snapshot as is done now, i.e. the builds are not deterministic which can be especially problematic if you are checking out past versions of pekko-http (i.e. trying to disect/figure out a bug) and instead of resolving the pekko version at that time you are instead resolving a completely different modern version.

PR #297, opened 2023-08-07
!sbt Use multiJvmPlugin from pekko.

He-Pin:
Just copy code from pekko, better to submit this upstream, will do later.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -11,7 +11,7 @@ import ValidatePullRequest._
 import net.bzzt.reproduciblebuilds.ReproducibleBuildsPlugin.reproducibleBuildsCheckResolver
 import PekkoDependency._
 import Dependencies.{ h2specExe, h2specName }
-import com.typesafe.sbt.SbtMultiJvm.MultiJvmKeys.MultiJvm
+import MultiJvmPlugin.MultiJvmKeys.MultiJvm
 import java.nio.file.Files
 import java.nio.file.attribute.{ PosixFileAttributeView, PosixFilePermission }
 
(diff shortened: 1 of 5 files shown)


pjfanning (approved):
approved as temporary solution - if it doesn't help, we should probably revert this - if it does work, we should be looking at creating a shared sbt plugin with this change

jrudolph:
What problem does it solve?

He-Pin:
> What problem does it solve?
> 

It start the Player only after The Controller can be connected.

mdedetrich:
> approved as temporary solution - if it doesn't help, we should probably revert this - if it does work, we should be looking at creating a shared sbt plugin with this change

@pjfanning This is probably a good candidate to include in the `incubator-pekko-sbt-build` I was talking about

He-Pin:
That's true, the sbt one seems unmaintained at least for now.

He-Pin:
@mdedetrich 
 1. We need wait the `Controller` transited to connectable before start the `Player`.
 2. Need submit a PR to `sbt-multi-jvm` and waiting a new snapshot, which may take sometimes and is not a *quick fix*.
 3. I did not diff the code with sbt-multi-jvm line by line, but there must be some reason for Akka was keep it separately

mdedetrich:
@He-Pin Lets discuss in https://github.com/apache/incubator-pekko/issues/548

jrudolph:
Where did the problem occur and how was it fixed? Which kubernetes cluster
are we connecting against here?

Matthew de Detrich ***@***.***> schrieb am Di., 8. Aug. 2023,
09:12:

He-Pin:
@jrudolph After migrate the `netty3` to `netty4`, the `Contoller` part which started at `index==0` seems not bind faster enough. The thread model has been changed in netty 4, or the max attept in the `Player` can be changed to a larger value, does that's what you expected?

He-Pin:
Let's me submit a fix and then this code will not be needed.

PR #305, opened 2023-08-23
Use c.abort in macro error cases

mdedetrich:
I started working on macros recently and when browsing through pekko-http I noticed that we are throwing exceptions rather than `c.abort` which is what you are intended to use in unhandled macro cases (`c.abort` contains information about macro expansion such as the current macro position which is helpful for the error message).


nvollmar (approved):
lgtm

He-Pin:
I know little about macro but Great!

PR #311, opened 2023-08-30
=core Make use of statefulMap instead of statefulMapConcat.

He-Pin:
Motivation:
Use the `statefulMap` instead when concating is not needed.

```
[info] Benchmark                                                Mode  Cnt         Score         Error  Units
[info] MessageToFrameRendererBenchmark.benchNewStreamedFrames  thrpt   10  25333318.747 锟斤拷 1293108.754  ops/s
[info] MessageToFrameRendererBenchmark.benchOldStreamedFrames  thrpt   10  20352150.287 锟斤拷 1563364.567  ops/s
```

PR #312, opened 2023-08-31
mima check on pull requests

pjfanning:
* relates to #311 
* I don't know for definite but that PR looks like the removed method should cause a mima failure

### DIFF
--- a/.github/workflows/validate-and-test.yml
+++ b/.github/workflows/validate-and-test.yml
@@ -88,7 +88,7 @@ jobs:
       # Quick testing for PR validation
       - name: Validate pull request for JDK ${{ matrix.JDK }}, Scala ${{ matrix.SCALA_VERSION }}
         if: ${{ github.event_name == 'pull_request' }}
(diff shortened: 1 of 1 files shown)


He-Pin (changes requested):
better to have a dedicated `binary-compatibility-checks` workflow like pekko

mdedetrich:
Can we also please make a decision on https://lists.apache.org/thread/4oymvkv6t415pocdtc6l1vxssoflbw9f and https://github.com/apache/incubator-pekko/pull/532 ? The whole point of that thread is to actually implement a proper handling of MiMa that sticks to semver and actually automatically deals with changes where people add functionality (doing MiMa checks against a hardcoded Pekko 1.0.0/1.1.0 is not enough, we should actually be doing MiMa checks against the latest release of the software against every PR).

I understand the solution may not be the cleanest (although tbh we kind of have our hands tied due to ASF process) but what I am suggesting is standard and normal in Scala OS projects, its not some crazy out there idea (otherwise this feature in dynver wouldn't even exist).

He-Pin (approved):
After another think, I think we can take this in first after then we can consider if we want to do some kind of spliting.

PR #313, opened 2023-09-05
Use fetch tags from latest checkout action

mdedetrich:
The latest version of the checkout github actions now supports a `fetch-tags` flag which does as described, i.e. it will fetch all of our tags which removes the various workarounds that was used before (i.e. manually running `git fetch` command) etc etc, see https://github.com/actions/checkout/pull/579

This will also help in the erroneous errors that we get in CI, i.e. 

```
[error] Failed to derive version from git tags. Maybe run `git fetch --unshallow` or `git fetch upstream` on a fresh git clone from a fork? Derived version: 0.0.0+1-337943bd-SNAPSHOT`
```

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -13,9 +13,10 @@ jobs:
     if: github.repository == 'apache/incubator-pekko-http'
     steps:
       - name: Checkout current branch (full)
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           fetch-depth: 0
+          fetch-tags: 0
           persist-credentials: false
 
       - name: Check project is formatted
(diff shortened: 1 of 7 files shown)


.github/workflows/format.yml
```diff
@@ -13,9 +13,10 @@ jobs:
     if: github.repository == 'apache/incubator-pekko-http'
     steps:
       - name: Checkout current branch (full)
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
```
He-Pin, line 16: Can we add a dependency-bot, is that allowed?

He-Pin (approved):
lgtm

PR #333, opened 2023-10-18
allow test with pekko 1.0.x snapshot

pjfanning:
* adjusts the tagNumber logic to treat build numbers without an RC or M tag as newer than ones that have them
* example
```
sbt -Dpekko.http.build.pekko.version=1.0.x package
```

### DIFF
--- a/project/PekkoDependency.scala
+++ b/project/PekkoDependency.scala
@@ -44,6 +44,7 @@ object PekkoDependency {
       case None =>
         Option(System.getProperty("pekko.http.build.pekko.version")) match {
           case Some("main")           => mainSnapshot
+          case Some("snapshot-1.0.x") => snapshot10x
           case Some("default") | None => Artifact(defaultVersion)
           case Some(other)            => Artifact(other, true)
         }
@@ -55,6 +56,7 @@ object PekkoDependency {
   val default = pekkoDependency(defaultVersion = minimumExpectedPekkoVersion)
   def docs = default
 
+  lazy val snapshot10x = Artifact(determineLatestSnapshot("1.0", false), true)
   lazy val mainSnapshot = Artifact(determineLatestSnapshot(), true)
 
   val pekkoVersion: String = default match {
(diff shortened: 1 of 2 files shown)


project/PekkoDependency.scala
```diff
@@ -44,6 +44,7 @@ object PekkoDependency {
       case None =>
         Option(System.getProperty("pekko.http.build.pekko.version")) match {
           case Some("main")           => mainSnapshot
+          case Some("snapshot-1.0.x") => snapshot10x
```
mdedetrich, line 47: Can we change the naming to `main-1.0.x` so its consistent with `main`?

mdedetrich (changes requested):
I am already testing this locally (and yes there is a regression in 1.0.x, I will make a revert commit), have one comment.

project/PekkoDependency.scala
pjfanning: `main` is the name of the branch we are using for 1.1.x 
so calling the option to use a 1.0.x snapshot something with the word `main` in it would be very confusing

mdedetrich: > `main` is the name of the branch we are using for 1.1.x so calling the option to use a 1.0.x snapshot something with the word `main` in it would be very confusing

No its not, `main` points to the latest `main` branch which are snapshots. There is an argument the naming is confusing but if we want to change that lets do it in another PR

mdedetrich:
@pjfanning Do you want to get this over the finish line? I am happy with the change, just need want to change from `snapshot-1.0.x` to just `1.0.x`.

I can also update the PR branch myself if you have time constraints.

pjfanning:
> @pjfanning Do you want to get this over the finish line? I am happy with the change, just need want to change from `snapshot-1.0.x` to just `1.0.x`.
> 
> I can also update the PR branch myself if you have time constraints.

I've committed a change to call it `1.0.x`.

mdedetrich:
@pjfanning Also need to add it to the matrix at https://github.com/apache/incubator-pekko-http/blob/17da94039e5b63d2be8f73cfac673e2a6f8517ca/.github/workflows/nightly.yml#L18 if you want the tests to actually run it.

project/PekkoDependency.scala
```diff
@@ -114,11 +116,12 @@ object PekkoDependency {
               ep.toInt,
               maj.toInt,
               min.toInt,
-              Option(tagNumber).map(_.toInt),
+              Option(tagNumber).map(_.toInt).getOrElse(Integer.MAX_VALUE),
```
mdedetrich, line 119: Why was this change needed?

mdedetrich:
I found something else, i.e. https://github.com/apache/incubator-pekko-http/pull/333#discussion_r1375408032 . Also PR needs to be rebased/merge conflict against main since there are conflicts.

project/PekkoDependency.scala
pjfanning: yes - this is related to `adjusts the tagNumber logic to treat build numbers without an RC or M tag as newer than ones that have them`

mdedetrich: Okay, but how is `Integer.MAX_VALUE` going to provide a correct value here, or am I missing something? Is it just some sentinel value that will get ignored later (if so add a comment because logic/reasoning behind it is unclear).

mdedetrich: Oh nvm, its for ordering and not the actual value. @pjfanning Can you just add a comment explaining that its there to fix ordering and then ill approve PR.

pjfanning: The values are used only in sorting.

* `-RC1-` will eval as 1
* no `-RC` or `-M` will now eval as MAX-INT
* MAX-INT > 1 so sorting will treat non-RC/non-milestone releases as newer

Old logic:
* `-RC1-` will eval as `Some(1)`
* no `-RC` or `-M` will eval as `None`
* `None` < `Some(1)` so sorting will treat non-RC/non-milestone releases as older

mdedetrich: Understood, just add a comment in the code quickly explaining this because on first glance its misleading.

mdedetrich (approved):
lgtm, lets see how this runs

mdedetrich:
@pjfanning Good news its working, i.e. you can check the latest run at https://github.com/apache/incubator-pekko-http/actions/runs/6687481949 and to further confirm if you check the job at https://github.com/apache/incubator-pekko-http/actions/runs/6687481949/job/18168181064 you can see its picking up the latest `1.0.x` snapshot, i.e.

```
[info] Building Pekko HTTP 1.1.0-M0+18-a7676ef1+20231030-0213-SNAPSHOT against Pekko 1.0.1+49-f1f6285b-SNAPSHOT on Scala 2.13.11
```

Issue #332, opened 2023-10-12
CVE-2023-44487


He-Pin:
Description
A client might overload the server by issue frequent RST frames. This can cause a massive amount of load on the remote system and so cause a DDOS attack.

References
https://www.cve.org/CVERecord?id=CVE-2023-44487
https://blog.cloudflare.com/technical-breakdown-http2-rapid-reset-ddos-attack/
https://cloud.google.com/blog/products/identity-security/google-cloud-mitigated-largest-ddos-attack-peaking-above-398-million-rps/

I just read this, not sure if pekko-http is affected.
Netty's fix is https://github.com/netty/netty/commit/58f75f665aa81a8cbcf6ffa74820042a285c5e61

@jrudolph cc

jrudolph:
In my tests on my machine, you can load a core with about 8-10MB/s of network input using either just RST frames or HEADERS/RST. How bad is that? It means you already have to almost saturate a 100GBit link for this to become relevant. 

On the other hand, the real cost might come from actually running requests that are already cancelled (see also #330). These are run in a stream with configurable parallelism, so there's no unlimited forking of work going on.

In perspective, this is probably not the best targeted attack you could do if you want to target a specific site (because you only target HTTP/2 infrastructure and not all the rest of the backend). On the other hand, it is unfortunate if you get hit by an easy drive-by attack.

One thing to consider is that it is not recommended to run Pekko HTTP on the public internet directly (as it is not recommended to run any web server on the public internet directly without the necessary precautions, you would much more likely put a more well-established web proxy like nginx etc in front of any real loads, especially if you have exposure). So, the real impact of this problem is probably negligible right now.

A solution will probably want to implement some kind of rate limiting. Unfortunately, this will introduce both a small performance penalty but also another knob that you have to tune, especially for high loads. This makes for an awkward position when choosing the default because there's little reason to enable the rate limit for a non-exposed load (= almost all) while when you are running it on the public internet you will have to tweak settings to enable the feature.

atennapel:
I noticed that Pekko-http is not listed on the vulnerability page here: https://nvd.nist.gov/vuln/detail/CVE-2023-44487 (but Akka-http is). Do you know why that is?

He-Pin:
@atennapel Maybe pekko-http is new, and @jrudolph is working on a PR for this.

pjfanning:
[...]

PR #344, opened 2023-10-29
HTTP/2 rapid reset mitigation

pjfanning:
* #332
* adds a throttle and the connection is closed if the reset frame rate is too high
* code is largely based on suggestions from @jrudolph 
* the mima filter had to be added due to the new fields in Http2Settings
* the mima check did not work so I copied over the working MiMa.scala from incubator-pekko repo


jrudolph:
Thanks for setting that up and sorry for dropping the ball here.

pjfanning:
@jrudolph @raboof @nvollmar @kerr @gmethvin @mdedetrich would any of you have time to review this? This is a reaction to the Akka team issuing a security warning. I would favour not getting a CVE or attaching our name to the existing CVE-2023-44487 but I'm not against it either. My priority is just to get out a release with some change.

project/MiMa.scala
```diff
@@ -11,57 +11,54 @@
...
 import sbt.Keys._
-import com.typesafe.tools.mima.core.ProblemFilter
 import com.typesafe.tools.mima.plugin.MimaPlugin
 import com.typesafe.tools.mima.plugin.MimaPlugin.autoImport._
 
-import scala.util.Try
-
 object MiMa extends AutoPlugin {
 
+  private val latestPatchOf10 = 0
```
He-Pin, line 22: Split this change to a seprate PR?

http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/Http2Blueprint.scala
```diff
@@ -198,6 +199,20 @@ private[http] object Http2Blueprint {
       Flow[FrameEvent].map(FrameRenderer.render).prepend(Source.single(Http2Protocol.ClientConnectionPreface)),
       Flow[ByteString].via(new Http2FrameParsing(shouldReadPreface = false, log)))
 
+  private def rapidResetMitigation(
+      settings: Http2ServerSettings): BidiFlow[FrameEvent, FrameEvent, FrameEvent, FrameEvent, NotUsed] = {
+    def frameCost(event: FrameEvent): Int = event match {
+      case _: FrameEvent.DataFrame         => 0
+      case _: FrameEvent.WindowUpdateFrame => 0 // TODO: should we throttle these?
+      case _                               => 1
+    }
```
He-Pin, line 208: this can be contantFunc

project/MiMa.scala
pjfanning: One advantage to having it here is this PR proves it works. I could move it in a separate PR but the build will pass regardless.

http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/Http2Blueprint.scala
```diff
@@ -198,6 +199,20 @@ private[http] object Http2Blueprint {
...
+
+    BidiFlow.fromFlows(
+      Flow[FrameEvent],
+      Flow[FrameEvent].throttle(settings.resetsThrottleCost, settings.resetsThrottleInterval,
+        settings.resetsThrottleBurst, frameCost, ThrottleMode.Enforcing))
```
He-Pin, line 213: 1. As the default `throttle` implementation will do a scheduling, but we will always close/fails on the first rapid reset detected, maybe a simpler copy/impl can be better?

http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/Http2Blueprint.scala
pjfanning: I don't know what `constantFunc` means. Do you mean making this a `val` instead of a `def`?

http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/Http2Blueprint.scala
```diff
@@ -127,6 +127,7 @@ private[http] object Http2Blueprint {
       serverDemux(settings.http2Settings, initialDemuxerSettings, upgraded) atop
       FrameLogger.logFramesIfEnabled(settings.http2Settings.logFrames) atop // enable for debugging
       hpackCoding(masterHttpHeaderParser, settings.parserSettings) atop
+      rapidResetMitigation(settings.http2Settings) atop
```
He-Pin, line 130: Better with an option to not include this stage into flow, if the endpoint is not public but internal.

pjfanning: makes sense - there are 3 new configs - one is an interval - what about me updating the code so that the interval is 0 then the feature is just disabled?

http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/Http2Blueprint.scala
pjfanning: Could you suggest an implementation?

http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/Http2Blueprint.scala
He-Pin: Yes, make just like the `Keep.left/right`

http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/Http2Blueprint.scala
He-Pin: I think we can leverage the current TokenBucket, IIRC, the Netty fix  https://github.com/netty/netty/commit/58f75f665aa81a8cbcf6ffa74820042a285c5e61 does no scheduling.

http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/Http2Blueprint.scala
pjfanning: made a change

http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/Http2Blueprint.scala
pjfanning: where is this TokenBucket? No class with that name in this repo. The Netty PR does not seem to have a TokenBucket either.

He-Pin: `org.apache.pekko.util.TokenBucket` from `pekko-actor`

pjfanning: The throttle code already uses a TokenBucket under the hood. See https://github.com/search?q=repo%3Aapache%2Fincubator-pekko%20tokenbucket&type=code

This issue is time sensitive. Pekko HTTP has no current release that has a solution for CVE-2023-44487

@jrudolph would you have any thoughts on this? Do you think the throttle adds excessive overhead?

pjfanning: @He-Pin The throttle in 'Enforcing' mode like this PR uses does not seem to schedule - see this code in Throttle.scala.

```
      override def onPush(): Unit = {
        val elem = grab(in)
        val cost = costCalculation(elem)
        val delayNanos = tokenBucket.offer(cost)

        if (delayNanos == 0L) push(out, elem)
        else {
          if (mode eq Enforcing) failStage(new RateExceededException("Maximum throttle throughput exceeded."))
          else {
            currentElement = elem
            scheduleOnce(Throttle.TimerKey, delayNanos.nanos)
          }
        }
```

http-core/src/main/resources/reference.conf
```diff
@@ -307,6 +307,12 @@ pekko.http {
       # Fail the connection if a sent ping is not acknowledged within this timeout.
       # When zero the ping-interval is used, if set the value must be evenly divisible by less than or equal to the ping-interval.
       ping-timeout = 0s
+
+      # Configure the throttle for Reset Frames (https://github.com/apache/incubator-pekko-http/issues/332)
+      resets-throttle-cost = 100
+      resets-throttle-burst = 100
+      # setting resets-throttle-interval to 0s will disable the throttle
+      resets-throttle-interval = 1s
```
mdedetrich, line 315: Did we make an official decision about whether this mitigation should be enabled by default or not? My memory may be fuzzy but I thought we decided to not have it enabled by default but have the option so people can enable it (in the case they expose pekko-http directly to the internet).

Also not sure if a formal discussion on the mailing list for this is excessive.

http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/Http2Blueprint.scala
mdedetrich: Maybe @He-Pin is talking about the `scheduleOnce` here?

pjfanning: The `scheduleOnce` in this code snippet does not happen because `mode eq Enforcing`.

project/MiMa.scala
mdedetrich: I would prefer moving this to a separate PR as well, its not directly related to this change and its also something I would like to review in isolation (its also scratching this pet peeve of mine that I have echoed elsewhere which is that that we shouldn't even have this code, i.e. it should be automated via https://github.com/sbt/sbt-dynver?tab=readme-ov-file#previous-version-detection.

mdedetrich (changes requested):
In terms of implementation I don't have too many issues however @He-Pin if you have time it would be great to address @pjfanning 's outstanding issues.

The MiMa changes definitely should go into a separate PR (which can be merged before this one at which point this PR should be rebased).

Also a the risk of bikeshedding, I would prefer if the `resets-throttle-cost` config was changed to `reset-throttle-cost` (i.e. de-pluralize the resets). To me as an english speaker "resets" sounds slightly off.

http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/Http2Blueprint.scala
mdedetrich: Thanks for clarifying, in this case it seems to be a false alarm

pjfanning:
I think I've covered most if the review items

http-core/src/main/resources/reference.conf
pjfanning: renamed to `reset-frame-throttle...`

pjfanning:
I've just noticed that this throttle breaks some of the microbenchmarks. I think the throttle might be applied to too many frame types.

Running sbt from the incubator-pekko-http dir and executing

```
http-bench-jmh/Jmh/run -i 3 -wi 3 -f1 -t1 org.apache.pekko.http.impl.engine.http2.H2ServerProcessingBenchmark.
```

This will fail with the changes in this PR but runs ok without them. The new throttle is set to fail if there are too many frames. This benchmark does not test rapid resets directly - so it appears that the new throttle is affecting too many frame types.

pjfanning:
I have changed the cost function on the frame throttle to only throttle RST frames. It may be useful to also throttle some other frame types - possibly also:
* Continuation Frames
* GoAway Frames
* WindowUpdate Frames?

pjfanning:
I changed the code so that the rapid reset check is disabled by default. Would it be possible to have this reviewed again?

In the benchmark, the check does not seem to affect performance.

He-Pin:
> I changed the code so that the rapid reset check is disabled by default. Would it be possible to have this reviewed again?
> 
> In the benchmark, the check does not seem to affect performance.

How about with a configuration?

pjfanning:
These are the results of running the H2ClientServerBenchmark on my laptop (with the change in this PR, resetFrameThrottleInterval=0s means no throttle applied). As you may see there is a small impact on my laptop, 0s runs are a little faster than 1s runs.

```
[info] Benchmark                                       (minStrictEntitySize)  (requestbody)  (resetFrameThrottleInterval)  (responsetype)   Mode  Cnt      Score      Error  Units
[info] H2ClientServerBenchmark.benchRequestProcessing                      1          empty                            1s          strict  thrpt    5  34987.925 ±  911.714  ops/s
[info] H2ClientServerBenchmark.benchRequestProcessing                      1          empty                            1s  closedelimited  thrpt    5  29428.729 ± 2196.091  ops/s
[info] H2ClientServerBenchmark.benchRequestProcessing                      1          empty                            0s          strict  thrpt    5  36503.237 ± 2168.155  ops/s
[info] H2ClientServerBenchmark.benchRequestProcessing                      1          empty                            0s  closedelimited  thrpt    5  31649.809 ± 1056.275  ops/s
[info] H2ClientServerBenchmark.benchRequestProcessing                      1    singleframe                            1s          strict  thrpt    5  34068.047 ± 1009.244  ops/s
[info] H2ClientServerBenchmark.benchRequestProcessing                      1    singleframe                            1s  closedelimited  thrpt    5  29889.829 ± 2166.454  ops/s
[info] H2ClientServerBenchmark.benchRequestProcessing                      1    singleframe                            0s          strict  thrpt    5  35286.571 ± 2411.328  ops/s
[info] H2ClientServerBenchmark.benchRequestProcessing                      1    singleframe                            0s  closedelimited  thrpt    5  29821.582 ± 2270.763  ops/s
```

pjfanning:
without my PR changes (ie latest main branch code)

```
[info] Benchmark                                       (minStrictEntitySize)  (requestbody)  (responsetype)   Mode  Cnt      Score      Error  Units
[info] H2ClientServerBenchmark.benchRequestProcessing                      1          empty          strict  thrpt    5  35597.887 ± 2219.899  ops/s
[info] H2ClientServerBenchmark.benchRequestProcessing                      1          empty  closedelimited  thrpt    5  30468.592 ± 1722.882  ops/s
[info] H2ClientServerBenchmark.benchRequestProcessing                      1    singleframe          strict  thrpt    5  34457.236 ± 2100.593  ops/s
[info] H2ClientServerBenchmark.benchRequestProcessing                      1    singleframe  closedelimited  thrpt    5  29022.231 ± 5096.366  ops/s
```

mdedetrich (approved):
lgtm, would be good to get an approval from someone else.

http2-tests/src/test/scala/org/apache/pekko/http/impl/engine/http2/Http2SpecWithMaterializer.scala
```diff
@@ -0,0 +1,138 @@
...
+
+        // force connection to shutdown (in case it is an invalid state)
+        setup.network.fromNet.sendError(new RuntimeException)
+        setup.network.toNet.cancel()
+
+        // and then assert that all stages, substreams in particular, are stopped
+      }
+  }
+
+  protected /* To make ByteFlag warnings go away */ abstract class TestSetupWithoutHandshake {
```
laglangyue, line 52: The comment after `protected` doesn't seem elegant

pjfanning: https://github.com/apache/incubator-pekko-http/pull/344/files/a58637af2969f314cd900b44cb87523dc60e61fd#diff-1bf6a966e29f3f599b7126edc971a2eb45087a0f37ff3b578389f38f257424a5L1809

laglangyue: yeah, Should we keep it orgin? If it's not necessary, I still hope to be able to modify it.

pjfanning:
closing due to #394

PR #346, opened 2023-10-29
avoid deprecated getFirst function

### LINKED ISSUE #345: fix use of deprecated code
```
[warn] /Users/pj.fanning/code/incubator-pekko-http/http-core/src/main/scala/org/apache/pekko/http/scaladsl/model/HttpEntity.scala:666:23: method getFirst in class Attributes is deprecated (since Akka 2.5.7): Attributes should always be most specific, use get[T]
[warn]           _attributes.getFirst[SizeLimit] match {
```

pjfanning:
fixes #345


pjfanning:
breaks too many tests

PR #351, opened 2023-10-30
no need to cast to ActorMaterializer

pjfanning:
* these changes were taken from #347 but I don't want to backport all of that
* see also https://github.com/apache/incubator-pekko-http/pull/352
* ActorMaterializer is deprecated and it would be good not to assume all Materializers are ActorMaterializer


nvollmar (approved):
lgtm

PR #354, opened 2023-10-30
fix more akka naming

pjfanning:
* raised https://github.com/apache/incubator-pekko-http/issues/355 for follow up
* some issue links were incorrectly updated to point to pekko-http issues but the pekko links don't work


nvollmar (approved):
lgtm

PR #363, opened 2023-11-11
remove more uses of ActorMaterializer

pjfanning:
there are still some left but these are the easy ones to remove


He-Pin (approved):
lgtm

PR #365, opened 2023-11-19
[DRAFT] uptake sbt-pekko-build

pjfanning:
WIP towards centralising some build logic in https://github.com/pjfanning/sbt-pekko-build

early days


pjfanning:
@mdedetrich is it possible to merge this as at least a short term fix for the mima-filters issue (found in #344). A better solution can be added later but right now the mima-filters logic doesn't work in pekko-http build - and copying the Mima.scala from incubator-pekko (where mima-filters work) via the sbt-pekko-build helps.

PR #369, opened 2023-12-01
standardise on specs 4.20

pjfanning:
https://mvnrepository.com/artifact/org.specs2/specs2-core

* most recent 4.x release
* simplifies the build - avoids having different versions depending on scala version


nvollmar (approved):
lgtm

PR #371, opened 2023-12-08
Make .splitWhen explicit with SubstreamCancelStrategy.drain

mdedetrich:
The context of this change is https://github.com/apache/incubator-pekko/pull/252. I want to try and get that PR over the finish line for Pekko 1.1.0 and while that PR is not a binary breaking change it is a behaviour breaking change.

What this PR does is it makes any usage of `splitWhen` explicitly use `SubstreamCancelStrategy.drain` (which is the current default) so that if someone uses pekko-http with a future release of pekko 1.1.0 there won't be any behaviour change. Note that I am not entirely sure if this change makes sense or even if its necessary i.e. it may be the case that whatever `SubstreamCancelStrategy` you use it doesn't matter because the various `SubFlow`'s cannot error (in which case this PR is pointless and can be closed) or it could also be that we really should be explicit about which `SubstreamCancelStrategy` to use for correctness reasons (i.e. if hypothetically an error in a `SubFlow` can occur then it should always be either `SubstreamCancelStrategy.drain` or `SubstreamCancelStrategy.propagate`)

@jrudolph @raboof @kerr Pinging you here since you have most of the domain knowledge.

### DIFF
--- a/http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/WebSocket.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/WebSocket.scala
@@ -177,7 +177,7 @@ private[http] object WebSocket {
     def prepareMessages: Flow[MessagePart, Message, NotUsed] =
       Flow[MessagePart]
         .via(PrepareForUserHandler)
-        .splitAfter(_.isMessageEnd)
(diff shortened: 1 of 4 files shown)


mdedetrich:
Im going to close this since the premise is premature, should tackle each of the cases individually if necessary.

PR #376, opened 2023-12-21
Add UnsupportedContentTypeException javadsl

mdedetrich:
This PR adds a `javadsl` equivalent of `org.apache.pekko.http.scaladsl.unmarshalling.Unmarshaller.UnsupportedContentTypeException` without breaking binary compatibility (you can confirm this by running `sbt mimaReportBinaryIssues`)

While the changes aren't as trivial as what would be considered ideal, they are necessary in order to not break MiMa. Of note is that the core data structure for the exception is being changed from Scala `Set`/`Option` to Java `Set`/`Optional`. Additionally the javadsl version of `UnsupportedContentTypeException` doesn't support Scala specific language features (i.e. `Product`/`Serializable`/`canEqual`).

### DIFF
--- a/http/src/main/scala/org/apache/pekko/http/javadsl/unmarshalling/Unmarshaller.scala
+++ b/http/src/main/scala/org/apache/pekko/http/javadsl/unmarshalling/Unmarshaller.scala
@@ -14,6 +14,7 @@
 package org.apache.pekko.http.javadsl.unmarshalling
 
 import java.util.concurrent.CompletionStage
+import java.util.Optional
 
 import org.apache.pekko
 import pekko.actor.ClassicActorSystemProvider
@@ -21,14 +22,12 @@ import pekko.annotation.InternalApi
 import pekko.http.impl.model.JavaQuery
 import pekko.http.impl.util.JavaMapping
 import pekko.http.impl.util.JavaMapping.Implicits._
-import pekko.http.javadsl.model._
+import pekko.http.{ javadsl => jm }
+import jm.model._
 import pekko.http.scaladsl.model.{ ContentTypeRange, ContentTypes }
 import pekko.http.scaladsl.unmarshalling
 import pekko.http.scaladsl.unmarshalling.FromEntityUnmarshaller
-import pekko.http.scaladsl.unmarshalling.Unmarshaller.{
-  EnhancedFromEntityUnmarshaller,
-  UnsupportedContentTypeException
-}
+import pekko.http.scaladsl.unmarshalling.Unmarshaller.EnhancedFromEntityUnmarshaller
 import pekko.http.scaladsl.util.FastFuture
 import pekko.stream.{ Materializer, SystemMaterializer }
 import pekko.util.ByteString
@@ -85,8 +84,9 @@ object Unmarshaller extends pekko.http.javadsl.unmarshalling.Unmarshallers {
         val mediaType = t.asScala
         if (entity.contentType == ContentTypes.NoContentType || mediaType.matches(entity.contentType.mediaType)) {
           um.asScala(entity)
-        } else FastFuture.failed(UnsupportedContentTypeException(Some(entity.contentType),
-          ContentTypeRange(t.toRange.asScala)))
+        } else FastFuture.failed(
+          pekko.http.scaladsl.unmarshalling.Unmarshaller.UnsupportedContentTypeException(Some(entity.contentType),
+            ContentTypeRange(t.toRange.asScala)))
       }
     }
   }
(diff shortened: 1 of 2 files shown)


http/src/main/scala/org/apache/pekko/http/javadsl/unmarshalling/Unmarshaller.scala
```diff
@@ -124,6 +124,40 @@ object Unmarshaller extends pekko.http.javadsl.unmarshalling.Unmarshallers {
       implicit mi: JavaMapping[JI, SI]): unmarshalling.Unmarshaller[JI, O] =
     um.asInstanceOf[unmarshalling.Unmarshaller[JI, O]] // since guarantee provided by existence of `mi`
 
+  class UnsupportedContentTypeException(
+      private val _supported: java.util.Set[jm.model.ContentTypeRange],
+      private val _actualContentType: Optional[jm.model.ContentType])
+      extends RuntimeException(_supported.asScala.mkString(
```
mdedetrich, line 130: Converting the java data-structures to Scala is done here so that the runtime exception message is the same as it is currently. This is done just incase currently existing code is matching against the exception message although one can argue that you shouldn't be doing this, furthermore if users are doing this its likely they will just be matching against `Unsupported Content-Type` in which case this will still work.

http/src/main/scala/org/apache/pekko/http/scaladsl/unmarshalling/Unmarshaller.scala
```diff
@@ -190,7 +202,7 @@ object Unmarshaller
 
     override def equals(that: Any): Boolean = that match {
       case that: UnsupportedContentTypeException =>
-        that.canEqual(this) && that.supported == this.supported && that.actualContentType == this.actualContentType
+        that.canEqual(this) && super.equals(this)
```
mdedetrich, line 205: This will point to `org.apache.pekko.http.javadsl.unmarshalling.Unmarshaller.UnsupportedContentTypeException` `equals` method which does the proper structural equality check.

http/src/main/scala/org/apache/pekko/http/javadsl/unmarshalling/Unmarshaller.scala
```diff
@@ -124,6 +124,40 @@ object Unmarshaller extends pekko.http.javadsl.unmarshalling.Unmarshallers {
...
+        _supported.asScala.toSet.asInstanceOf[Set[pekko.http.scaladsl.model.ContentTypeRange]],
+        _actualContentType.asScala)
+
+    def getSupported(): java.util.Set[jm.model.ContentTypeRange] = _supported
+
+    def getActualContentType(): Optional[jm.model.ContentType] = _actualContentType
+
+    override def equals(that: Any): Boolean = that match {
+      case that: UnsupportedContentTypeException =>
+        that._supported == this._supported && that._actualContentType == this._actualContentType
```
mdedetrich, line 156: Since it may not be immediately obvious, both `java.util.Set` and `java.util.Optional` equals methods perform structural equality, not reference (I tested this in Scala repl).

http/src/main/scala/org/apache/pekko/http/javadsl/unmarshalling/Unmarshaller.scala
```diff
@@ -14,21 +14,20 @@
...
-import pekko.http.javadsl.model._
+import pekko.http.{ javadsl => jm }
+import jm.model._
 import pekko.http.scaladsl.model.{ ContentTypeRange, ContentTypes }
 import pekko.http.scaladsl.unmarshalling
 import pekko.http.scaladsl.unmarshalling.FromEntityUnmarshaller
-import pekko.http.scaladsl.unmarshalling.Unmarshaller.{
-  EnhancedFromEntityUnmarshaller,
-  UnsupportedContentTypeException
-}
```
He-Pin, line 31: This will break user's code

http/src/main/scala/org/apache/pekko/http/javadsl/unmarshalling/Unmarshaller.scala
```diff
@@ -85,8 +84,9 @@ object Unmarshaller extends pekko.http.javadsl.unmarshalling.Unmarshallers {
         val mediaType = t.asScala
         if (entity.contentType == ContentTypes.NoContentType || mediaType.matches(entity.contentType.mediaType)) {
           um.asScala(entity)
-        } else FastFuture.failed(UnsupportedContentTypeException(Some(entity.contentType),
-          ContentTypeRange(t.toRange.asScala)))
+        } else FastFuture.failed(
+          pekko.http.scaladsl.unmarshalling.Unmarshaller.UnsupportedContentTypeException(Some(entity.contentType),
```
He-Pin, line 88: split with a single import line ? this line is too long

http/src/main/scala/org/apache/pekko/http/javadsl/unmarshalling/Unmarshaller.scala
mdedetrich: How, it passes MiMa? The exception being thrown here is the same.

http/src/main/scala/org/apache/pekko/http/javadsl/unmarshalling/Unmarshaller.scala
```diff
@@ -14,21 +14,20 @@
...
+import java.util.Optional
 
 import org.apache.pekko
 import pekko.actor.ClassicActorSystemProvider
 import pekko.annotation.InternalApi
 import pekko.http.impl.model.JavaQuery
 import pekko.http.impl.util.JavaMapping
 import pekko.http.impl.util.JavaMapping.Implicits._
-import pekko.http.javadsl.model._
+import pekko.http.{ javadsl => jm }
```
laglangyue, line 25: Do you think of using `javadsl` instead of jm?

mdedetrich: `jm` is whats used in the rest of the codebase so I kept the same convention.

http/src/main/scala/org/apache/pekko/http/javadsl/unmarshalling/Unmarshaller.scala
mdedetrich: Fixed and pushed.

mdedetrich: So I had to undo this change because using a single inline import makes the code fail to compile on Scala 3 due to

```
[error] -- [E049] Reference Error: /home/runner/work/incubator-pekko-http/incubator-pekko-http/http/src/main/scala/org/apache/pekko/http/javadsl/unmarshalling/Unmarshaller.scala:88:33 
[error] 88 |        } else FastFuture.failed(UnsupportedContentTypeException(Some(entity.contentType),
```

A better alternative might be to throw the `javadsl` variant however this is technically a behavioural change (hypothetically users may be relying on the scala specific features of scaladsl when catching the exception).

pjfanning:
added to 1.1.0 milestone - seems ok to me if targeted just at that milestone

I'm not fully convinced that we can't have some binary incompatibility in a 1.1.0 release but the changes here to maintain bin compatibility are not very complicated and look maintainable.

mdedetrich:
> added to 1.1.0 milestone - seems ok to me if targeted just at that milestone

fine with me

> I'm not fully convinced that we can't have some binary incompatibility in a 1.1.0 release but the changes here to maintain bin compatibility are not very complicated and look maintainable.

We did decide that both pekko and pekko-http strictly follow semver due to being critical components which is why https://github.com/apache/incubator-pekko-http/blob/38aa25e7c20232a4226c4ea5767813b99bca870a/build.sbt#L59 is set.

In any the point is moot since this PR goes out of its way to **not** break binary compatibility.

pjfanning (approved):
lgtm

mdedetrich:
Going to go ahead and merge this

PR #377, opened 2023-12-21
Update Scala to 3.3.2

mdedetrich:
Scala 3.3.2-RC1 just came out so this PR is testing it out to see if the codebase compiles. It will remain in draft until full release of Scala 3.3.2 is out.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -35,7 +35,7 @@ object Dependencies {
 
   val scala212Version = "2.12.18"
   val scala213Version = "2.13.12"
-  val scala3Version = "3.3.1"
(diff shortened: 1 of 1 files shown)


pjfanning:
It is my understanding that pekko and pekko-http are part of the Scala community build. Isn't that enough?

mdedetrich:
> It is my understanding that pekko and pekko-http are part of the Scala community build. Isn't that enough?

Possibly, afaik the community build only runs compile (and it may only do compile on the main scope and not `test`) where as this PR makes sure that both `main` and `test` scope are compiled as well as running tests.

pjfanning:
closed due to #502

Issue #380, opened 2023-12-25
remove minExpectedVersion check


pjfanning:
See https://github.com/apache/incubator-pekko-http/pull/328

@mdedetrich has suggested that we should be testing with pekko 1.0.2 or latest pekko 1.0.x snapshot (in Slack - link is private). 

This check is currently what ties up to pekko 1.0.0.

Alternatively, we need to change the build so we can set the min expected value to 1.0.0 but test with a different version of pekko. I still think the min check is unnecessary. 1.0.0 is our first release so there are no releases that we need to ban. We can add this check back when we need to ban pekko releases (if ever).

He-Pin:
We can do a time based release bom, eg 2023-03-release

He-Pin:
I think we can adjust that check than remove it.

mdedetrich:
So I am not sure what the "check" part of `minExpectedVersion` check is meant to be, but at least starting off with Pekko (i.e. when we forked Akka), pekko-http should be compatible with **all** versions of Pekko 1.0.x, after all this is what semver means and why both pekko and pekko-http follow semver strictly.

This may be an artifact from older versions of Akka where some functionality was added to Akka that Akka-Http relied on and hence Akka-Http required a minimum version of Akka in order to function.

Until thats the case, I am all for removing this `minExpectedVersion` check and ideally (as discussed on slack) pekko-http (and the other modules) should be building against the latest of Pekko that is semver compatible (which in this case will be Pekko 1.1.x) with nightly checks that link with older minor versions of Pekko (i.e. 1.0.x)

mdedetrich:
@pjfanning On a related note I made a PR against sbt-pekko-build so that we at least are using the latest version of Pekko 1.0.x. In the meantime we can work on removing the check as well.

pjfanning:
#383 and #382  merged

PR #382, opened 2023-12-26
build with pekko 1.0.2

pjfanning:
relates to #380 

* sbt-pekko-build 0.2.0 uses pekko 1.0.2 as the default
* VersionGenerator uses a minPekkoVersion value that is set to 1.0.0
  * this means that the validation mentioned in #380 will allow all versions from Pekko 1.0.0 and up


He-Pin (approved):
Lgtm

laglangyue:
lgtm

PR #390, opened 2023-12-31
Replace ExecutionContexts.sameThreadExecutionContext with parasitic

mdedetrich:
`ExecutionContexts.sameThreadExecutionContext` is deprecated and `ExecutionContexts.parasitic` should be used instead (one is an alias for the other).


He-Pin (approved):
Lgtm

gmethvin (approved):

PR #393, opened 2024-01-01
Use designated Scala methods for converter utils

mdedetrich:
This PR is a response to https://github.com/apache/incubator-pekko/pull/894, specifically rather than using the Java designed converter methods within Scala sources we use the Scala ones so that when the project is compiled with Scala 3 it gets inlined.

Note that in the case of `BasicDirectives`, the reason why pekko-http was using the designated Java methods in the first place is due to ambiguous implicit imports which is why `CompletionStageOps` is being used rather than the typical use of extension methods within `implicit class`.


laglangyue:
lgtm

He-Pin (approved):
Lgtm

PR #394, opened 2024-01-02
Http2 frame type throttle

pjfanning:
* replace #344 - see #332
* reworked configs - we can specify a list of frame types to apply the throttle to
* defaults to an empty list so no throttle is applied at all in the default case


http-core/src/main/scala/org/apache/pekko/http/scaladsl/settings/Http2ServerSettings.scala
```diff
@@ -124,6 +137,10 @@ object Http2ServerSettings extends SettingsCompanion[Http2ServerSettings] {
       logFrames: Boolean,
       pingInterval: FiniteDuration,
       pingTimeout: FiniteDuration,
+      frameTypeThrottleFrameTypes: Seq[String],
```
mdedetrich, line 140: This should be a `Set`

http-core/src/main/resources/reference.conf
```diff
@@ -307,6 +307,17 @@ pekko.http {
       # Fail the connection if a sent ping is not acknowledged within this timeout.
       # When zero the ping-interval is used, if set the value must be evenly divisible by less than or equal to the ping-interval.
       ping-timeout = 0s
+
+      frame-type-throttle {
+        # Configure the throttle for non-data frame types (https://github.com/apache/incubator-pekko-http/issues/332)
+        # supported frame-types for throttlng are:
+        # reset, headers, continuation, go-away, priority, ping, push-promise, window-update
```
mdedetrich, line 314: While we are here, should we documented a suggested list of frame types and/or should we create an alias for all of them?

mdedetrich (changes requested):
Have a couple of comments otherwise lgtm

http-core/src/main/scala/org/apache/pekko/http/scaladsl/settings/Http2ServerSettings.scala
pjfanning: thanks - changed

http-core/src/main/resources/reference.conf
pjfanning: I'd prefer not to provide too much guidance. If users want to protect against CVE-2023-44487, they just need `frame-types = ["reset"]`. In @jrudolph's original solution, almost the full list in line 314 was throttled (the exception was `window-update` wasn't throttled but there was a TODO to work out whether it should be throttled).

If we do make a recommendation, I would suggest `frame-types = ["reset"] and suggest to users that they might want to expand the list.

pjfanning: @mdedetrich wdyt?

pjfanning:
@mdedetrich would it be possible for you to re-review this?

mdedetrich (approved):
lgtm

mdedetrich:
@He-Pin Did you have any last minute comments for this PR?

http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/Http2Blueprint.scala
```diff
@@ -122,12 +122,18 @@ private[http] object Http2Blueprint {
...
-      hpackCoding(masterHttpHeaderParser, settings.parserSettings) atop
-      framing(log) atop
+      hpackCoding(masterHttpHeaderParser, settings.parserSettings)
+
+    val frameTypesForThrottle = getFrameTypesForThrottle(settings.http2Settings)
+    val flow1 = if (frameTypesForThrottle.nonEmpty) {
+      flow0 atop rapidResetMitigation(settings.http2Settings, frameTypesForThrottle) atopKeepLeft framing(log)
+    } else flow0 atop framing(log)
+
+    flow1 atop
```
He-Pin, line 136: Can `flow0` and `flow1` be replaced with a more meaningful name?

http-core/src/main/scala/org/apache/pekko/http/scaladsl/settings/Http2ServerSettings.scala
```diff
@@ -102,6 +103,18 @@ trait Http2ServerSettings extends javadsl.settings.Http2ServerSettings with Http
...
+  def withFrameTypeThrottleFrameTypes(frameTypes: Set[String]) = copy(frameTypeThrottleFrameTypes = frameTypes)
+
+  def frameTypeThrottleCost: Int
+  def withFrameTypeThrottleCost(cost: Int) = copy(frameTypeThrottleCost = cost)
+
+  def frameTypeThrottleBurst: Int
+  def withFrameTypeThrottleBurst(burst: Int) = copy(frameTypeThrottleBurst = burst)
+
+  def frameTypeThrottleInterval: FiniteDuration
+  def withFrameTypeThrottleInterval(interval: FiniteDuration) = copy(frameTypeThrottleInterval = interval)
```
He-Pin, line 116: use `java.time.Duration`

http-core/src/test/scala/org/apache/pekko/http/impl/util/WithLogCapturing.scala
```diff
@@ -86,6 +92,19 @@ trait WithLogCapturing extends SuiteMixin { this: TestSuite =>
       Failed(new AssertionError(
         s"No severe log messages should be emitted during test run but got [${stats(
             Logging.WarningLevel)}] warnings and [${stats(Logging.ErrorLevel)}] errors (see marked lines above)"))
+    } else if (expectSevereLogsOnlyToMatch.nonEmpty) {
+      val severeEvents = events.filter(isSevere(_))
+      val matchingEvents = severeEvents.filter(_.message.toString.contains(expectSevereLogsOnlyToMatch.get))
+      if (severeEvents.isEmpty || matchingEvents != severeEvents) {
+        val stats = events.groupBy(_.level).mapValues(_.size).toMap.withDefaultValue(0)
```
He-Pin, line 99: IIRC, there is a `groupMap`.

He-Pin (approved):
Great, overall lgtm

http-core/src/main/resources/reference.conf
He-Pin: I think we can just keep this simple, a user can add what they want.

http-core/src/main/scala/org/apache/pekko/http/scaladsl/settings/Http2ServerSettings.scala
```diff
@@ -136,6 +153,7 @@ object Http2ServerSettings extends SettingsCompanion[Http2ServerSettings] {
     require(minCollectStrictEntitySize <= (incomingConnectionLevelBufferSize / maxConcurrentStreams),
       "min-collect-strict-entity-size <= incoming-connection-level-buffer-size / max-concurrent-streams")
     require(outgoingControlFrameBufferSize > 0, "outgoing-control-frame-buffer-size must be > 0")
+    require(frameTypeThrottleInterval.toMillis > 0, "frame-type-throttle.interval must be a positive duration")
```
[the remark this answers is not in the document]
pjfanning: I still think this is better to not allow bad values - the default value is ok and we have a warning in the reference.conf.

And it is more efficient to do 1 check instead of 2 checks.

http-core/src/test/scala/org/apache/pekko/http/impl/util/WithLogCapturing.scala
pjfanning: this matches the code used in a couple of other places - if we want to try groupMap, I think that should be in a separate PR that addresses all similar code in pekko-http too.

http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/Http2Blueprint.scala
pjfanning: is cdfca2d098eacae695fb4c09cc37276e2d4f2d2c ok?

http-core/src/main/scala/org/apache/pekko/http/scaladsl/settings/Http2ServerSettings.scala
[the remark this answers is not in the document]
He-Pin: OK's let's ship it.

PR #398, opened 2024-01-07
use specs2 v5

pjfanning:
currently tied to v4


pjfanning:
specs2 v5 only supports scala 3

PR #414, opened 2024-01-07
Add and fix sbt paradox update ignore

mdedetrich:
Incase we ever end up using scala steward on 1.0.x so we don't create unnecessary PR's


pjfanning (approved):
lgtm

PR #418, opened 2024-01-14
enable compile inline (upgrade sbt-pekko-build)

pjfanning:
sbt-pekko-build 0.3.0 enables Scala compile time inlining 

Do not backport this as is to 1.0.x branch. 1.0.x branch will need to have this flag changed to false:

```
ThisBuild / pekkoInlineEnabled := false
```


mdedetrich (approved):
lgtm, only one comment.

If you want I can handle the cherry pick to `1.0.x` while making sure that the inlining is disabled.

build.sbt
```diff
@@ -28,6 +28,9 @@ commands := commands.value.filterNot { command =>
   }
 }
 
+// pekkoInlineEnabled must be set to false when this is backported to 1.0.x branch
+ThisBuild / pekkoInlineEnabled := true
```
mdedetrich, line 32: Technically this part is not needed since its already enabled by default, I was planning to add

```scala
// pekkoInlineEnabled must be set to false when this is backported to 1.0.x branch
ThisBuild / pekkoInlineEnabled := false
```

When doing the cherry pick backport to `1.0.x`.

I can handle the backports if you want, this is not hard to do.

pjfanning: I prefer to keep this explicit. There are a few contributors and they are not all up to date with the subtleties of this sbt plugin.

mdedetrich: Fair enough, I won't fight over this. We can always change it later once the storm of the 1.1.x releases are over

mdedetrich:
> What is the bytecode diff here and how does the performance change?

I can create a bytecode diff using jardiff, let me do this now

> how does the performance change?

I did some performance testing a while back but its on my other machine (currently on holidays in Aus). At the time I didn't see anything incredibly significant, however its still free performance with no downside. If you want I can redo the benchmarks albeit its going to be on an M1 laptop

jrudolph:
> At the time I didn't see anything incredibly significant

Ok, but what's the rationale for a potential intrusive compilation change if it does not change anything significantly (and we don't track performance changes diligently)?

mdedetrich:
> Ok, but what's the rationale for a potential intrusive compilation change if it does not change anything significantly (and we don't track performance changes diligently)?

You were probably unavailable at the time when it was being discussed, I think that significantly may not have been the wisest choice of words as it depends on the JDK version (the whole point of the inliner is that it removes the type of Scala bytecode that the JVM doesn't optimize which is more prelevant in earlier JDK's such as 1.8 but from what I gathered most real hotspots have already been handled).

Where I disagree is about the intrusive part, the inliner has been used in the most critical parts of Scala ecosystem including scalac itself for many years now (scalac wouldn't use the inliner for no reason), even parboiled2 is using the inliner and most importantly there isn't any regression.

Tracking performance diligently is definitely an issue, unfortunately we don't have the best setup for this right now. But ultimately even minor performance gains is still performance.

jrudolph:
> the inliner has been used in the most critical parts of Scala ecosystem including scalac itself for many years now

scalac is a different case as scalac is an application with only few dependencies and not a library, so it can benefit in different ways from inlining than a library.

> even parboiled2 is using the inliner

This is somewhat irrelevant since much of the performance critical parts of parboiled2 are only generated in the user code at a different compilation time with different settings, and as you know no performance benchmarks have been run before merging that change, especially not in the downstream libraries...

In general, for best performance you will have to evolve and continuously benchmark your codebase including all your settings and the environment. This is hard, especially since we do not have any control over e.g. the choice of JDK which can have a huge impact. Incidentally, in my tests, JDK 8 is often still the fastest option in many cases.

The [inliner documentation](https://docs.scala-lang.org/overviews/compiler-options/optimizer.html#inliner-heuristics-and-inline) shows this warning right at the front:

> Don’t enable the optimizer during development: it breaks incremental compilation, and it makes the compiler slower. Only enable it for testing, on CI, and to build releases.

How do we deal with it? How is developer efficiency impacted by this change?

> But ultimately even minor performance gains is still performance.

That's my point, right now, in my tests, there's zero performance difference (but I cannot test it comprehensively) but we added complexity and don't fully understand the consequences.

The only thing I ask for is that we don't just do changes because we think they might be useful but that we collect evidence first and carefully evaluate pros and cons, especially if the evidence is so scarce.

Right now the balance is:
 - added complexity
 - potential productivity and correctness issues during development
 - untested performance impact in comprehensive tests
 +/- little/zero negative/positive performance change in random testing on my machine

mdedetrich:
> How do we deal with it? How is developer efficiency impacted by this change?

There is a toggle/knob to disable inlining for local development, see https://github.com/pjfanning/sbt-pekko-build/blob/main/src/main/scala/com/github/pjfanning/pekkobuild/PekkoInlinePlugin.scala#L56. In practice this has been a none issue. Once we are able to release in CI I do plan on having the inliner disabled by default and only enabled in CI

> That's my point, right now, in my tests, there's zero performance difference (but I cannot test it comprehensively) but we added complexity and don't fully understand the consequences.

Im running benchmarks now on JDK 1.8

> The only thing I ask for is that we don't just do changes because we think they might be useful but that we collect evidence first and carefully evaluate pros and cons, especially if the evidence is so scarce.

Of course, and this is exactly the reason why we both enabled the inliner sooner and are doing a M1 release, its so that people can test more often. The primary reason to enable the inliner now is that we are doing a minor bump change (1.0.x to 1.1.x) which gives plenty of people a signal of "please test, and test a lot".

Regarding benchmarks, I should be done in 40 min. I am doing the whole suite while keeping the CPU degrees constant on my M1

He-Pin:
@mdedetrich You should have a fan beside your laptop, otherwise it may degree its performance because of heat.

mdedetrich:
@jrudolph Sorry for the delay, o I have some good news and some bad news

 the inliner did make some noticeable performance improvements when running against Pekko Core 1.0.2, you can see the results below

- Pekko Http with core 1.0.2 without inline https://gist.github.com/mdedetrich/b078ed767341b5bde8cc687c91345305
- Pekko Http with core 1.0.2 with inline https://gist.github.com/mdedetrich/1e9f27a33316761359cd52b4c164ce99

Here is an [inline diff](https://www.diffchecker.com/iH0WbWNZ/) for easy comparison

In some cases there are noticeable improvements i.e. `org.apache.pekko.http.impl.engine.StreamServerProcessingBenchmark.benchRequestProcessing` which is ~12% faster

The bad news is that for some reason, Pekko Core 1.1.0 appears to have a performance regression (this is unrelated to inline). I will make an issue for this in Pekko core after some more initial investigation.

mdedetrich:
> The bad news is that for some reason, Pekko Core 1.1.0 appears to have a performance regression (this is unrelated to inline). I will make an issue for this in Pekko core after some more initial investigation.

So this seems to have been a false alarm, I reran the benchmark from a clean slate and got the results that I expected. You did raise a point about being diligent with benchmarks so its something we have to look into.

mdedetrich:
So I made a thread about this (plus more general topic of dedicate HW for benchmarking) at https://lists.apache.org/thread/kbv10ngvjdhffc80jkf82o98ldo08ngw

I am also attaching the JVM bytecode diffs. Note that the diffs are represented in git format, so just extract the compressed folders and open up git diff to see the diff.

[http-caching.zip](https://github.com/apache/incubator-pekko-http/files/13957447/http-caching.zip)
[http-core.zip](https://github.com/apache/incubator-pekko-http/files/13957448/http-core.zip)
[http-cors.zip](https://github.com/apache/incubator-pekko-http/files/13957457/http-cors.zip)
[http-spray-json.zip](https://github.com/apache/incubator-pekko-http/files/13957459/http-spray-json.zip)
[http-xml.zip](https://github.com/apache/incubator-pekko-http/files/13957461/http-xml.zip)

PR #424, opened 2024-01-20
use ByteString.toArrayUnsafe where it is safe to do so and improves performance

pjfanning:
only safe to use toArrayUnsafe when you know that you are not mutating the array

the advantage is that you safe on having to clone the inner array that the ByteString has

there might be 1 or 2 more places that we could update in pekko-http too

### DIFF
--- a/http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/hpack/ByteStringInputStream.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/hpack/ByteStringInputStream.scala
@@ -28,7 +28,7 @@ private[http2] object ByteStringInputStream {
     bs match {
       case cs: ByteString1C =>
         // TODO optimise, ByteString needs to expose InputStream (esp if array backed, nice!)
-        new ByteArrayInputStream(cs.toArray)
+        new ByteArrayInputStream(cs.toArrayUnsafe())
       case _ =>
         // NOTE: We actually measured recently, and compact + use array was pretty good usually
         apply(bs.compact)
--- a/http2-tests/src/test/scala/org/apache/pekko/http/impl/engine/http2/Http2FrameHpackSupport.scala
+++ b/http2-tests/src/test/scala/org/apache/pekko/http/impl/engine/http2/Http2FrameHpackSupport.scala
@@ -58,7 +58,7 @@ trait Http2FrameHpackSupport extends Http2FrameProbeDelegator with Http2FrameSen
   val decoder = new Decoder(Http2Protocol.InitialMaxHeaderListSize, Http2Protocol.InitialMaxHeaderTableSize)
 
   def decodeHeaders(bytes: ByteString): Seq[(String, String)] = {
-    val bis = new ByteArrayInputStream(bytes.toArray)
+    val bis = new ByteArrayInputStream(bytes.toArrayUnsafe())
     val hs = new VectorBuilder[(String, String)]()
 
     decoder.decode(bis,
--- a/http/src/main/scala/org/apache/pekko/http/scaladsl/coding/DeflateCompressor.scala
+++ b/http/src/main/scala/org/apache/pekko/http/scaladsl/coding/DeflateCompressor.scala
@@ -51,7 +51,7 @@ class DeflateCompressor private[coding] (compressionLevel: Int) extends Compress
 
   protected def compressWithBuffer(input: ByteString, buffer: Array[Byte]): ByteString = {
     require(deflater.needsInput())
-    deflater.setInput(input.toArray)
+    deflater.setInput(input.toArrayUnsafe())
     drainDeflater(deflater, buffer)
   }
   protected def flushWithBuffer(buffer: Array[Byte]): ByteString = {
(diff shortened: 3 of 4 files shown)


pjfanning:
@mdedetrich what do you think? It seems to me like this is safe and to be honest, I'm a little surprised that it doesn't work this way already.

He-Pin:
How about with a jmh number?

pjfanning:
I can add a benchmark but first, I'd like to know whether the changes are valid.

He-Pin:
@pjfanning If the raw data will not be changed and the endIndx - startIndx == length,   seems fine,otherwise better to do a copy.

mdedetrich:
> @mdedetrich what do you think? It seems to me like this is safe and to be honest, I'm a little surprised that it doesn't work this way already.

I am also getting the same impression @jrudolph @raboof Do you have any comments on this?

He-Pin:
It will throw if the byteString is not a Bytestring1c/1
image

pjfanning:
> It will throw if the byteString is not a Bytestring1c/1 And another problem is when a slice changed the underling bytes, this will cause problem too. ![image](https://private-user-images.githubusercontent.com/501740/298313240-df4f4442-f306-47c9-8072-fff25eeb0d19.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MDU3OTUzODAsIm5iZiI6MTcwNTc5NTA4MCwicGF0aCI6Ii81MDE3NDAvMjk4MzEzMjQwLWRmNGY0NDQyLWYzMDYtNDdjOS04MDcyLWZmZjI1ZWViMGQxOS5wbmc_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjQwMTIwJTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI0MDEyMFQyMzU4MDBaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT00ZTdkOTdjMWJkYTg0YWE5OWNiZWIxMWI4NmFhZGZiNzY0MTg5NmJjNzkzZGRhZTQ1ODgyMDA0YThiZDk0OGI1JlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZhY3Rvcl9pZD0wJmtleV9pZD0wJnJlcG9faWQ9MCJ9.cg2KE4LEbyh1ShQ9ixeB7UFBi3mH4tYBUyKU0QmNb4M

what will throw? the pattern matching is completly unchanged from what is was before - all I have changed is toArray to toArrayUnsafe

He-Pin:
@pjfanning hah, my bad, I was thinking the `copyToArray`, it's good to go for me.

He-Pin (approved):
lgtm.

http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/hpack/ByteStringInputStream.scala
```diff
@@ -28,7 +28,7 @@ private[http2] object ByteStringInputStream {
     bs match {
       case cs: ByteString1C =>
         // TODO optimise, ByteString needs to expose InputStream (esp if array backed, nice!)
```
laglangyue, line 30: Here is a todo, considering optimization here, can it be deleted.
If not relevant,  ignore it.
LGTM

nvollmar (approved):
lgtm

http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/hpack/ByteStringInputStream.scala
He-Pin: refs: https://github.com/apache/incubator-pekko/issues/995

jrudolph: Weird that looks like a optimization only done half-way, this whole class looks somewhat useless in its current form, I wonder how it ended up that way.

The reason this does not turn up in profiles here is that the amount of data that is pushed through here is relatively low (not least because HPACK2 is a compression algorithm that limits the amount data necessary).

http2-tests/src/test/scala/org/apache/pekko/http/impl/engine/http2/Http2FrameHpackSupport.scala
```diff
@@ -58,7 +58,7 @@ trait Http2FrameHpackSupport extends Http2FrameProbeDelegator with Http2FrameSen
   val decoder = new Decoder(Http2Protocol.InitialMaxHeaderListSize, Http2Protocol.InitialMaxHeaderTableSize)
 
   def decodeHeaders(bytes: ByteString): Seq[(String, String)] = {
-    val bis = new ByteArrayInputStream(bytes.toArray)
+    val bis = new ByteArrayInputStream(bytes.toArrayUnsafe())
```
jrudolph, line 61: Does not matter to optimize tests.

PR #427, opened 2024-01-21
Update multi-jvm

mdedetrich:
This PR updates multi-jvm as fixing code that is now redundant due to improvements in multi-jvm (i.e. https://github.com/sbt/sbt-multi-jvm/pull/42)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -243,7 +243,6 @@ lazy val httpTests = project("http-tests")
   .enablePlugins(NoPublish) // don't release tests
   .enablePlugins(MultiNode)
   .disablePlugins(MimaPlugin) // this is only tests
-  .configs(MultiJvm)
   .settings(headerSettings(MultiJvm))
   .addPekkoModuleDependency("pekko-stream", "provided", PekkoCoreDependency.default)
   .addPekkoModuleDependency("pekko-multi-node-testkit", "test", PekkoCoreDependency.default)
--- a/project/MultiNode.scala
+++ b/project/MultiNode.scala
@@ -11,7 +11,7 @@
  * Copyright (C) 2009-2020 Lightbend Inc. <https://www.lightbend.com>
  */
 
-import com.typesafe.sbt.SbtMultiJvm
+import com.typesafe.sbt.{ MultiJvmPlugin, SbtMultiJvm }
 import com.typesafe.sbt.SbtMultiJvm.MultiJvmKeys._
 import sbt._
 import sbt.Keys._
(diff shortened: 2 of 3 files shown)


build.sbt
```diff
@@ -243,7 +243,6 @@ lazy val httpTests = project("http-tests")
   .enablePlugins(NoPublish) // don't release tests
   .enablePlugins(MultiNode)
   .disablePlugins(MimaPlugin) // this is only tests
-  .configs(MultiJvm)
```
mdedetrich, line 246: This is no longer needed due to https://github.com/sbt/sbt-multi-jvm/issues/41

project/MultiNode.scala
```diff
@@ -33,7 +33,7 @@ object MultiNode extends AutoPlugin {
   val multiTest = CliOptions.multiNode.ifTrue(MultiJvm / multiNodeTest).getOrElse(MultiJvm / test)
 
   override def trigger = noTrigger
-  override def requires = plugins.JvmPlugin
+  override def requires = plugins.JvmPlugin && MultiJvmPlugin
```
mdedetrich, line 36: This is now required because beforehand the  `.configs(MultiJvm)`  was inadvertently causing the `MultiJvmPlugin` to load before the `MultiNode` plugin.

Now that `.configs(MultiJvm)` is removed from `build.sbt` we need to mark `MultiJvmPlugin` as an explicit dependency of `MultiNodePlugin`. It should have been like this anyways since `MultiNode` does use `MultiJvm`.

He-Pin: The multiNode testing need be tested explicitly.
`sbt cluster/MultiJvm/test`

mdedetrich: > The multiNode testing need be tested explicitly. `sbt cluster/MultiJvm/test`

I don't understand the context of this comment. Are you saying that this is now a problem (i.e. the PR has changed the behaviour) or something else?

He-Pin: I think you can verify it locally, the ci is not testing it.

He-Pin: I will check it later.

mdedetrich: @He-Pin I think you have the wrong project, this is pekko-http and `cluster` is for pekko

mdedetrich: I also just ran the nightly tests manually and its behaving the same way as it currently does, see https://gist.github.com/mdedetrich/4e10d79f3687c49fd0886f83ec97d42a

He-Pin (approved):
Lgtm

mdedetrich:
Ill merge now so we can see a nightly tests run

PR #437, opened 2024-01-23
pekko-http2-support is part of pekko-http-core

pjfanning:
In older versions of Akka, there was a separate jar. This comment makes no sense in Pekko.


He-Pin (approved):
lgtm

jrudolph (approved):
True, lgtm

Issue #444, opened 2024-01-26
need to fix issues in source tgz with the linked-dir


pjfanning:
We had issues in the last pekko-http release with the linked dir used in testing. Seemed to only affect Windows users then but now, it's affecting Mac users.

```
x apache-pekko-http-1.0.1-RC1-incubating-src-20240126/http-tests/src/test/resources/dirWithLink/linked-dir: Truncated tar archive
tar: Error exit delayed from previous errors.
```

same dir that causes the issue in https://github.com/apache/incubator-pekko-http/issues/275

pjfanning:
I looked at sbt-source-dist and it might be possible to get it some symlinks - but I think we are better off rewriting the FileAndResourceDirectivesSpec so that it creates a temp dir that has all the files and symlinks set up by the test code before we run the main part of the test - i.e. remove the need for the symlink being checked into git and being reproduced in the tgz.

Issue #275, opened 2023-07-27
FileAndResourceDirectivesSpec fails when run from source distribution


pjfanning:
The source distributions are part of the release process (tgz and zip files).

See https://lists.apache.org/thread/smj8xz3q8mpmcrqy7bg34637m7qc4sn1

The test depends on a symbolic link that is checked into our git repo (http-tests/src/test/resources/dirWithLink).

Options include:
* disable the test if the symbolic link is missing
* try to get [sbt-source-dist](https://github.com/pjfanning/sbt-source-dist) to store symbolic links properly

pjfanning:
latest plan is https://github.com/apache/incubator-pekko-http/issues/444#issuecomment-1912087476

PR #445, opened 2024-01-26
rework symlink test

pjfanning:
* rework existing symlink test so that it creates the link itself (and removes it) - in /tmp folder for duration of test
* aim is to get 1.0.x ready for 1.0.1-RC2 - will forward fit after it is merged

#275, #444

### DIFF
--- a/http-tests/src/test/scala/org/apache/pekko/http/scaladsl/server/directives/FileAndResourceDirectivesSymlinkSpec.scala
+++ b/http-tests/src/test/scala/org/apache/pekko/http/scaladsl/server/directives/FileAndResourceDirectivesSymlinkSpec.scala
@@ -39,7 +39,7 @@ class FileAndResourceDirectivesSymlinkSpec extends RoutingSpec
   dirWithLink.mkdir()
   val symlink = Files.createSymbolicLink(
     Paths.get(dirWithLink.getAbsolutePath, "linked-dir"),
-    new File(testRoot, "subDirectory").toPath)
+    new File(testRoot, "subDirectory").toPath.toAbsolutePath)
 
   override def afterAll(): Unit = {
     super.afterAll()
@@ -59,6 +59,7 @@ class FileAndResourceDirectivesSymlinkSpec extends RoutingSpec
     def _getFromDirectory() = getFromDirectory(dirWithLink.getCanonicalPath)
 
     "not follow symbolic links to find a file" in {
+      Files.isSymbolicLink(symlink) shouldBe true
       EventFilter.warning(pattern = ".* points to a location that is not part of .*", occurrences = 1).intercept {
         Get("linked-dir/empty.pdf") ~> _getFromDirectory() ~> check {
           handled shouldBe false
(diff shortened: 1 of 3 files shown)


mdedetrich (approved):
Nothing immediate jumps out at me but I would get @jrudolph to check just to make sure nothing is missed

lgtm

http-tests/src/test/scala/org/apache/pekko/http/scaladsl/server/directives/FileAndResourceDirectivesSymlinkSpec.scala
```diff
@@ -0,0 +1,73 @@
...
+import pekko.testkit._
+
+import org.scalatest.{ BeforeAndAfterAll, Inside, Inspectors }
+
+class FileAndResourceDirectivesSymlinkSpec extends RoutingSpec
+    with Inspectors with Inside with BeforeAndAfterAll {
+
+  // need to serve from the src directory, when sbt copies the resource directory over to the
+  // target directory it will resolve symlinks in the process
+  val testRoot = new File("http-tests/src/test/resources")
```
He-Pin, line 33: Better with Files.path, and yes Winodws is using different separator

mdedetrich: So TIL that Windows has symlinks but only for [Windows 10/11](https://www.howtogeek.com/16226/complete-guide-to-symbolic-links-symlinks-on-windows-or-linux/). I don't know how this factors into everything, my initial inclination is to disable the test for Windows since this seems to only really be testing behaviour specific to Posix systems.

@He-Pin Can you confirm if the test passes on your machine (assuming all of the Windows related path separator issues have been solved)?

pjfanning: This line is copied from an existing test. The test either works or fails on Windows - either way noone is complaining about it.

pjfanning: My understanding is that Windows Java installs know how to apply the file path even if it doesn't use Windows idioms.

pjfanning: @He-Pin could you try running `sbt http-tests/test`? It doesn't take that long compared to similar commands in main pekko repo.

pjfanning: @He-Pin I created https://github.com/apache/incubator-pekko-http/issues/446 - is ok to say that you are not blocking this? I would like to get this merged so that I can do an RC2 for the Pekko HTTP 1.0.1 release.

PR #450, opened 2024-01-28
Updates sbt-pekko-build

mdedetrich:
(cherry picked from commit 56928146b469edfd5d2841943c2f059e56e8ddfc)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -22,12 +22,6 @@ import com.lightbend.paradox.apidoc.ApidocPlugin.autoImport.apidocRootPackage
 sourceDistName := "apache-pekko-http"
 sourceDistIncubating := true
 
-commands := commands.value.filterNot { command =>
-  command.nameOption.exists { name =>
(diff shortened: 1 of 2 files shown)


He-Pin (approved):
Lgtm

He-Pin:
Would you like to write a bot for this kind of cherry picking @mdedetrich there are so many projects....

mdedetrich:
> Would you like to write a bot for this kind of cherry picking @mdedetrich there are so many projects....

We need to be particular about what kind of commits we are cherry picking, i.e. for the 1.0.x branch it will only be build related commits.

Since we are doing a lot of build improvements via sbt-pekko-build its a given that we will be cherry picking a bit but once the builds for pekko projects stabilize more (i.e. sbt-pekko-build is almost finished) then it should reduce a lot.

The ultimate goal is to automate as much as possible via sbt-pekko-build so its much easier to maintain the pekko core projects

PR #453, opened 2024-01-28
Upgrade dependencies to move away from repo.scala-sbt.org

mkurz:
This PR is more a reminder, I don't expect it to be merged.
See 
- https://github.com/apache/incubator-pekko/pull/1061#issuecomment-1913598078


mkurz:
Outdated.

PR #456, opened 2024-01-29
disable inlining for now as it is not save nor efficient in development usage

jrudolph:
As the official documentation prominently states, the inliner is unsafe to use with incremental compilation (see https://github.com/sbt/zinc/issues/537). This leads to hard to diagnose issues during development.

It also massively inflates compile times (e.g. 16s -> 44s for http-core) making development much more painful than necessary.

Real world demonstration of correctness issues:

 * Add `private def doPull2(): ResponseOutput = MessageEnd` below `onPull` in HttpResponseParser.scala
 * `clean` and `compile`
 * Run tests in `HttpsProxyGraphStageSpec`
 * change `onPull` in `HttpResponseParser` to call `doPull2` instead of `doPull` (now doing incremental compilation)
 * Run tests in `HttpsProxyGraphStageSpec`, despite the broken implementation the test still completes fine

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -24,6 +24,8 @@ sourceDistIncubating := true
 
 ThisBuild / reproducibleBuildsCheckResolver := Resolver.ApacheMavenStagingRepo
 
+ThisBuild / pekkoInlineEnabled := false
+
 addCommandAlias("verifyCodeStyle", "scalafmtCheckAll; scalafmtSbtCheck; +headerCheckAll; javafmtCheckAll")
 addCommandAlias("applyCodeStyle", "+headerCreateAll; scalafmtAll; scalafmtSbt; javafmtAll")
 


mdedetrich:
Note that there is https://github.com/pjfanning/sbt-pekko-build/issues/9 which means it can be setup so the inliner only happens in CI (which is always clean environment) and not on local dev

He-Pin:
@mdedetrich Can you attach a framegraph to compare the difference? eg, using the https://github.com/LesnyRumcajs/grpc_bench as a demo.

mdedetrich:
> @mdedetrich Can you attach a framegraph to compare the difference? eg, using the https://github.com/LesnyRumcajs/grpc_bench as a demo.

I can't do this now because I don't have an appropriate machine to do this on (I am still overseas). If you have a desktop machine with locked + lower clocks and you are motivated yourself then feel free to go ahead.

Issue #395, opened 2024-01-03
Replace `statefulMapConcat` with `statefulMap` + `mapConcat` in codebase


mdedetrich:
As the title suggests, `statefulMapConcat` is deprecated so it should replaced with `statefulMap` + `mapConcat` so that the 1.1.0 version of pekko-http isn't relying on deprecated pekko methods.

laglangyue:
also found in pekko-connectors

mdedetrich:
> also found in pekko-connectors

Thanks, issue created at https://github.com/apache/incubator-pekko-connectors/issues/307

pjfanning:
https://github.com/apache/incubator-pekko-http/pull/311 was merged already but a quick search indicates that we might still have 1 or 2 remaining uses.

He-Pin:
I did a check seems there are no usage now. @pjfanning , the only lefts are in scaladoc and benchmarks

jrudolph:
This might have performance implications (mapConcat had performance issues in the past) so make sure to benchmark before doing these changes.

pjfanning:
This was deprecated in https://github.com/apache/incubator-pekko/issues/601 - if we think deprecating the code was a mistake then we can reopen that discussion. If we think it was the right decision to deprecate statefulMapConcat, then we should start removing its usage.

jrudolph:
I like the change from an API orthogonality perspective, just saying that it might not come for free. In the worst case, it could be possible to keep a copy in StreamUtils, but better if that is not needed.

He-Pin:
I think this issue can be closed as the associated change has been one. in https://github.com/apache/incubator-pekko-http/pull/311. 

Or we can do some dedicated implementation just for pekko-http usage in `StreamUtils`, without any additional allocations.

He-Pin:
The current `statefulMap` or `mapAccumlate` implementation will need a `(S, A)` , which is a kind  of allocation, in JEP https://openjdk.org/jeps/461 @viktorklang @viktorklang-ora is using an object which reducing this allocation.

He-Pin:
@jrudolph I will review the changes and try to come up with an allocation-free implementation in pekko-http.

pjfanning:
I'm going to close this because we undeprecated statefulMapConcat. Feel free to reopen.

PR #462, opened 2024-01-30
chore: Add a specialized `statefulMap` internal implement.

He-Pin:
Motivation:
In https://github.com/apache/incubator-pekko-http/issues/395 , @jrudolph said the current change may hurt performance, additional allocation for `(State, A)`, So I submit this, a simpler but still with the drawback for internal usage.

Result:
Performance keeped.

### DIFF
--- a/http-core/src/main/scala/org/apache/pekko/http/impl/util/StreamUtils.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/impl/util/StreamUtils.scala
@@ -14,7 +14,6 @@
 package org.apache.pekko.http.impl.util
 
 import org.apache.pekko
-import org.apache.pekko.stream.ActorAttributes.SupervisionStrategy
 import pekko.NotUsed
 import pekko.actor.Cancellable
 import pekko.annotation.InternalApi
@@ -22,6 +21,7 @@ import pekko.dispatch.ExecutionContexts
 import pekko.http.scaladsl.model.HttpEntity
 import pekko.http.scaladsl.util.FastFuture
 import pekko.stream._
+import pekko.stream.ActorAttributes.SupervisionStrategy
 import pekko.stream.impl.fusing.GraphInterpreter
 import pekko.stream.impl.fusing.GraphStages.SimpleLinearGraphStage
 import pekko.stream.scaladsl._
--- a/http-core/src/main/scala/org/apache/pekko/http/impl/util/StreamUtils.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/impl/util/StreamUtils.scala
@@ -436,7 +436,7 @@ private[http] class EnhancedByteStringSource[Mat](val byteStringStream: Source[B
               case Supervision.Restart =>
                 statefulFunction = statefulFunctionConstructor()
                 tryPull(in)
-              case _ => tryPull(in)
+              case Supervision.Resume => tryPull(in)
             }
         }
       }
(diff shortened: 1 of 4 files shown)


http-core/src/main/scala/org/apache/pekko/http/impl/util/StreamUtils.scala
```diff
@@ -396,3 +406,43 @@ private[http] class EnhancedByteStringSource[Mat](val byteStringStream: Source[B
       setHandlers(in, out, this)
     }
 }
+
+/**
+ * INTERNAL API
+ */
+@InternalApi private[http] final class SimpleStatefulMap[T, U](
```
pjfanning, line 413: I think we will need positive and negative tests for this code. It is one thing using a widely used lib function but when you create a new customised equivalent, you need to duplicate a good fraction of the tests from the original lib function.

He-Pin: Seems fine, but I will wait @jrudolph for some input before I put more time on this.

mdedetrich: At the risk of derailing, both this and the inlining discussion really demonstrate the need to have dedicated hardware for benchmarking and a setup so that we can easily compare performance inbetween changes

He-Pin: I will write some test this weekend when I back my hometown, wdyt about this approach @jrudolph ?

@mdedetrich Yes, have something like grpc-bench and test the performance continuously would be nice.

He-Pin: - [ ]  I have added the test.

He-Pin:
@Roiocam would you like give this a look,thanks.

pjfanning:
do we need this now that statefulMap has not been deprected?

He-Pin:
I think we can close this

http-core/src/main/scala/org/apache/pekko/http/impl/util/StreamUtils.scala
```diff
@@ -260,6 +261,15 @@ private[http] object StreamUtils {
       }
   }
 
+  /**
+   * Similar idea as [[FlowOps.statefulMap]] but simpler,
+   * which the state is captured instead of a passing in parameter, when
+   * the Flow completes, the state is lost, if you prefer additional emitting on complete,
+   * use [[FlowOps.statefulMap]] instead.
+   */
+  def statefulMap[T, U](functionConstructor: () => T => U): Flow[T, U, NotUsed] =
```
He-Pin, line 270: @pjfanning @jrudolph @mdedetrich Do you think we can put this in pekko-core too? we just need another name.

He-Pin: `lazyMap`?

pjfanning: please raise an issue in pekko-core

I have little objections to new methods. I just get worried when existing methods are changed and if that could have negative consequences.

He-Pin: I don' t have a better name too...

PR #470, opened 2024-02-05
Replace getFirst with get

mdedetrich:
So `getFirst` for getting an attribute is deprecated and you are instead meant to use `get` however this does have a different behaviour change (i.e. `getFirst` will get the first found attribute where as `get` will get the most specific attribute).

Since we are now in the 1.1.x branch these kinds of behaviour changes are allowed

### DIFF
--- a/http-core/src/main/scala/org/apache/pekko/http/scaladsl/model/HttpEntity.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/scaladsl/model/HttpEntity.scala
@@ -663,7 +663,7 @@ object HttpEntity {
         private var bytesLeft = Long.MaxValue
 
         override def preStart(): Unit = {
-          _attributes.getFirst[SizeLimit] match {
+          _attributes.get[SizeLimit] match {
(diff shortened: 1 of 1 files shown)


mdedetrich:
Okay so as I expected, we are actually relying on the behaviour of `getFirst` since the tests have started failing, will do some digging here.

mdedetrich:
@jrudolph @raboof So I had a quick look at this and I don't know what to do here, it seems like making `getFirst` deprecated was a mistake or maybe it should only be deprecated for public and not private inter-pekko use cases?

The whole reason why `getFirst` is necessary is so that the `withSizeLimit` directive works correctly with [nested routes](https://github.com/apache/incubator-pekko-http/blob/c5eda6a5d6182c2f6432e6c891ce263ae2e5d8aa/http-tests/src/test/scala/org/apache/pekko/http/scaladsl/server/directives/MiscDirectivesSpec.scala#L148-L185) and while it is possible to remove the deprecation of `getFirst` without using `get` in doing so it would just involve copying the implementation of it which means its not really solving a problem?

PR #478, opened 2024-02-09
upgrade paradox plugins

pjfanning:
* might fail due to java 8 build issues
* passed on my laptop - which surprised me

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -28,9 +28,9 @@ addSbtPlugin(("org.apache.pekko" % "pekko-sbt-paradox" % "1.0.0").excludeAll(
   "com.lightbend.paradox", "sbt-paradox",
   "com.lightbend.paradox" % "sbt-paradox-apidoc",
   "com.lightbend.paradox" % "sbt-paradox-project-info"))
-addSbtPlugin(("com.lightbend.paradox" % "sbt-paradox" % "0.9.2").force())
-addSbtPlugin(("com.lightbend.paradox" % "sbt-paradox-apidoc" % "0.10.1").force())
-addSbtPlugin(("com.lightbend.paradox" % "sbt-paradox-project-info" % "2.0.0").force())
+addSbtPlugin(("com.lightbend.paradox" % "sbt-paradox" % "0.10.6").force())
+addSbtPlugin(("com.lightbend.paradox" % "sbt-paradox-apidoc" % "1.1.0").force())
+addSbtPlugin(("com.lightbend.paradox" % "sbt-paradox-project-info" % "3.0.1").force())
 
 addSbtPlugin("de.heikoseeberger" % "sbt-header" % "5.10.0")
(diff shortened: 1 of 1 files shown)


mdedetrich:
> * might fail due to java 8 build issues
> * passed on my laptop - which surprised me

We need to be careful with this. The core issue is not that it would fail to compile but rather it links against the wrong JDK which means the generated api docs would t be correct since the lowest JDK version we support is 8.

I am trying to suggest that sbt-paradox maintains maintenance of its 0.9.x branch which supports JDK 8 with backports of from future paradox versions.

pjfanning:
* today we are dependent on a non-standard maven repo to get these old versions of plugins
* this is bad
* we need to use maven central
* we need a proper control over our builds to avoid supply chain injection attacks

mdedetrich:
> * today we are dependent on a non-standard maven repo to get these old versions of plugins
> * this is bad
> * we need to use maven central
> * we need a proper control over our builds to avoid supply chain injection attacks

Yes and the whole point of that ticket is to re-release current sbt-paradox 0.9.2 on maven

mkurz:
> the whole point of that ticket is to re-release current sbt-paradox 0.9.2 on maven

@mdedetrich Looking at https://repo1.maven.org/maven2/com/lightbend/paradox/sbt-paradox_2.12_1.0/ I think this is the case also for the 0.9.x line already? Or are you talking about a different library?

mdedetrich:
> > the whole point of that ticket is to re-release current sbt-paradox 0.9.2 on maven
> 
> @mdedetrich Looking at https://repo1.maven.org/maven2/com/lightbend/paradox/sbt-paradox_2.12_1.0/ I think this is the case also for the 0.9.x line already? Or are you talking about a different library?

Oh so it seems that sbt-paradox 0.9.2 has already been published on maven in which case 

> * today we are dependent on a non-standard maven repo to get these old versions of plugins

Is no longer an issue

pjfanning:
@mdedetrich @raboof https://github.com/apache/incubator-pekko-http/actions/runs/7871851115/job/21475933085?pr=478 just built web site ok with Java 8.

My plugins.sbt changes mean we use only artifacts that are on Maven Central but we still retain Java 8 support.

Is this ok?

raboof (approved):
This introduces a little risk since it's mixing paradox versions that aren't intended to work together, but since this is only used during the build and tested on CI I think that risk is manageable.

Aside from that I think it's an improvement so I'm :+1: on the change.

mdedetrich:
> This introduces a little risk since it's mixing paradox versions that aren't intended to work together, but since this is only used during the build and tested on CI I think that risk is manageable.

We have been doing this for a while and the API between different versions paradox/paradox-theme is completely stable from 0.9.x onwards.

This is one of the reasons why building the docs with JDK 11 hasn't been such a high priority, the JDK breakage between paradox 0.9.x and 0.10.x+ is not due paradox itself but the scala-xml/parboiled dependency that it brings in.

mdedetrich (approved):
lgtm

PR #479, opened 2024-02-10
mem settings for link-validator.yml

pjfanning:
Job is running into GC issues because it is running with constrained heap memory


mdedetrich (approved):
lgtm

PR #486, opened 2024-02-21
Use sbt-web 1.5.x despite staying on paradox 0.9.2

mkurz:
I am pretty sure this is safe since I am basically the on who maintains sbt-web and v1.5 is pretty much compatible with v1.4 API wise.
I use the same workaround in https://github.com/sbt/sbt-paradox-material-theme/blob/v0.7.0/project/plugins.sbt#L1-L3 actually

With this change I can `+Test/compile` pekko-http and I am not hitting https://repo.scala-sbt.org/ anymore (since I have blocked it on my machine).

@mdedetrich I am pretty sure this is ok to merge.

See
- https://github.com/apache/incubator-pekko/pull/1135


pjfanning (approved):
lgtm

PR #487, opened 2024-02-22
Remove apidoc from ignore and pin paradox versions

mdedetrich:
Pins the sbt-paradox/sbt-paradox-project-info versions since we can use the entire 0.9.x series (it being compiled against JDK 8).

Also removes `sbt-paradox-apidoc` from the ignore list since we are free to update this (note that `sbt-paradox-apidoc` was recently updated to latest version without any issues)


nvollmar (approved):
lgtm

Issue #493, opened 2024-02-27
API doc links broken


mpartel:
Both "API Documentation" links on https://pekko.apache.org/docs/pekko-http/current/reference.html are broken.

pjfanning:
We'll need to fix that. In the meantime, see https://pekko.apache.org/modules.html. It's links work.

pjfanning:
the links are fixed now in https://pekko.apache.org/docs/pekko-http/snapshot/reference.html#api-documentation

Will need to backport to 1.0.x branch to get current and 1.0 docs fixed.

PR #495, opened 2024-02-28
there is no index.html to link to

pjfanning:
relates to #493 

we will also need to changes to our .htaccess because the snapshot part of the URL is not currently supported either - see https://github.com/apache/incubator-pekko-site/pull/92 for that bit

### DIFF
--- a/docs/src/main/paradox/reference.md
+++ b/docs/src/main/paradox/reference.md
@@ -2,7 +2,7 @@
 
 ## API Documentation
 
- * @scaladoc:[Scaladoc](org.apache.pekko.http.scaladsl.index)
+ * @scaladoc:[Scaladoc](org.apache.pekko.http.scaladsl.package-summary)
  * @javadoc:[Javadoc](org.apache.pekko.http.javadsl.package-summary)
 
 ## Directives


docs/src/main/paradox/reference.md
```diff
@@ -2,7 +2,7 @@
 
 ## API Documentation
 
- * @scaladoc:[Scaladoc](org.apache.pekko.http.scaladsl.index)
+ * @scaladoc:[Scaladoc](org.apache.pekko.http.scaladsl.package-summary)
```
Roiocam, line 5: We have two api docs paths:

- `https://pekko.apache.org/api/`: has index.html, not package-summary.html
- `https://pekko.apache.org/japi/`: has package-summary.html, not index.html

I think there are some problems with the path under [this page](https://pekko.apache.org/docs/pekko-http/current/reference.html). One links to api and the other is japi. Why?

Roiocam: After looking at the Pekko core page, I think it is no problem. Scaladoc links to api (has index page), while javadoc points to japi (no index page)

pjfanning: I raised https://github.com/apache/incubator-pekko-http/pull/496

pjfanning:
will close this in favour of https://github.com/apache/incubator-pekko-http/pull/496

PR #496, opened 2024-02-29
fix api/japi links in build.sbt

pjfanning:
japi is for javadoc - api is for scaladoc - we have them the wrong way around in build.sbt

relates to https://github.com/apache/incubator-pekko-http/issues/493

### DIFF
--- a/scripts/link-validator.conf
+++ b/scripts/link-validator.conf
@@ -64,6 +64,7 @@ site-link-validator {
     "http://localhost:8080"
     # broken URL in license report
     "http://pholser.github.com/"
+    "http://specs2.org/"
   ]
 
   non-https-whitelist = [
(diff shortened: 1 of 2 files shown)


Roiocam (approved):
LGTM

scripts/link-validator.conf
```diff
@@ -64,6 +64,7 @@ site-link-validator {
     "http://localhost:8080"
     # broken URL in license report
     "http://pholser.github.com/"
+    "http://specs2.org/"
```
laglangyue, line 67: Ah...I can't visit this website. Why add this? 
maybe
https://etorreborre.github.io/specs2/

pjfanning: the license-report.html is generated - we don't really get to control what goes in it

that specs.org is what it is linking to and it now seems to be broken

so that means I need to get the validator to ignore it - that is what this setting is for

He-Pin (approved):
Lgtm

scripts/link-validator.conf
laglangyue: Thank you for your explanation. It may be an issue with specs2, and there may be an issue with their domain DNS.

LGTM

PR #508, opened 2024-03-03
Update sbt-pekko-paradox

mdedetrich:
(cherry picked from commit 687ee6114525ca2db309dd3a98a4b3cbfeaa36e0)

Backporting the latest pekko-sbt-paradox fixes. This is necessary due to the github api rate limiting issue so it also needs to be fixed for 1.0.x


pjfanning:
@mdedetrich the real concern with api.github.com is privacy. The ASF does not want us to call out to external sites because they could track our users.

pjfanning (approved):
lgtm

PR #519, opened 2024-03-16
support jackson buffer recycler

pjfanning:
see https://github.com/apache/pekko/pull/1192

### DIFF
--- a/http-marshallers-java/http-jackson/src/main/java/org/apache/pekko/http/javadsl/marshallers/jackson/Jackson.java
+++ b/http-marshallers-java/http-jackson/src/main/java/org/apache/pekko/http/javadsl/marshallers/jackson/Jackson.java
@@ -30,6 +30,9 @@ import com.fasterxml.jackson.core.JsonFactory;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.StreamReadConstraints;
 import com.fasterxml.jackson.core.StreamWriteConstraints;
+import com.fasterxml.jackson.core.util.BufferRecycler;
+import com.fasterxml.jackson.core.util.JsonRecyclerPools;
+import com.fasterxml.jackson.core.util.RecyclerPool;
 import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.json.JsonMapper;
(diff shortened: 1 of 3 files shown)


http-marshallers-java/http-jackson/src/main/java/org/apache/pekko/http/javadsl/marshallers/jackson/Jackson.java
```diff
@@ -113,7 +116,28 @@ static ObjectMapper createMapper(final Config config) {
         JsonFactory.builder()
             .streamReadConstraints(streamReadConstraints)
             .streamWriteConstraints(streamWriteConstraints)
+            .recyclerPool(getBufferRecyclerPool(config))
             .build();
     return new JsonMapper(jsonFactory);
   }
+
+  private static RecyclerPool<BufferRecycler> getBufferRecyclerPool(final Config cfg) {
+    switch (cfg.getString("buffer-recycler.pool-instance")) {
```
He-Pin, line 125: As it's been used in the exceptions, how about extract to a local val？

He-Pin (approved):
Lgtm, thanks

http-marshallers-java/http-jackson/src/main/java/org/apache/pekko/http/javadsl/marshallers/jackson/Jackson.java
He-Pin: And maybe we can assert it should be null too.

pjfanning: it is defaulted in reference.conf - virtually every config would blow up if we didn't have defaults - nowhere do we need to check for them not being set - because we have defaults in reference.conf

http-marshallers-java/http-jackson/src/main/java/org/apache/pekko/http/javadsl/marshallers/jackson/Jackson.java
```diff
@@ -113,7 +116,28 @@ static ObjectMapper createMapper(final Config config) {
...
+        return JsonRecyclerPools.sharedLockFreePool();
+      case "concurrent-deque":
+        return JsonRecyclerPools.newConcurrentDequePool();
+      case "shared-concurrent-deque":
+        return JsonRecyclerPools.sharedConcurrentDequePool();
+      case "bounded":
+        return JsonRecyclerPools.newBoundedPool(cfg.getInt("buffer-recycler.bounded-pool-size"));
+      default:
+        throw new IllegalArgumentException(
+            "Unknown recycler-pool: " + cfg.getString("buffer-recycler.pool-instance"));
```
He-Pin, line 140: I mean it is been used here too

He-Pin (approved):
Lgtm

Issue #396, opened 2024-01-03
More descriptive failure messages in HttpMessageParser


SkyTrix:
I'm currently investigating an issue where we're suddenly running into the following error: `HTTP chunk size exceeds the configured limit of 1048576 bytes`.

This is possibly due to an infrastructure change we made which results in a change of behavior when chunked transfer encoding is used. I did not find a way to determine the actual size of the incoming chunks and it would be really helpful if it was part of that error message.

My proposal here would be to extend all relevant errors in `HttpMessageParser` with extra parameter values. e.g. for the max chunk size, the full hex size could be read before validating against `maxChunkSize` and then including the `size` parameter in the error message:

https://github.com/apache/incubator-pekko-http/blob/4f80b1ef7cad2833382a659e7dd0b1a54078d126/http-core/src/main/scala/org/apache/pekko/http/impl/engine/parsing/HttpMessageParser.scala#L309-L320

Given that on line 318 the actual encountered illegal char is part of the error message, I think having the size in the other error is warranted. Apart from max chunk size there are a few other messages that could be improved such as max chunk extensions and amount of chunk trailer headers.

Any thoughts on this proposal?

pjfanning:
probably should have been closed after https://github.com/apache/pekko-http/pull/528 was merged

PR #528, opened 2024-03-28
Parse entire HTTP chunk size #396

SkyTrix:
Continue parsing HTTP chunk size up until Int.MaxValue so the actual size of the chunk can be logged if it exceeds the configured max chunk size.

### DIFF
--- a/http-core/src/main/scala/org/apache/pekko/http/impl/engine/parsing/HttpMessageParser.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/impl/engine/parsing/HttpMessageParser.scala
@@ -320,7 +320,7 @@ private[http] trait HttpMessageParser[Output >: MessageOutput <: ParserOutput] {
           case c if CharacterClasses.WSP(c) => parseSize(cursor + 1, size) // illegal according to the spec but can happen, see issue #1812
           case c                            => failEntityStream(s"Illegal character '${escape(c)}' in chunk start")
         }
-      } else failEntityStream(s"HTTP chunk size exceeds ${Int.MaxValue} bytes")
+      } else failEntityStream(s"HTTP chunk size exceeds Integer.MAX_VALUE (${Int.MaxValue}) bytes")
 
     try parseSize(offset, 0)
     catch {
--- a/http-core/src/main/scala/org/apache/pekko/http/impl/engine/parsing/HttpMessageParser.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/impl/engine/parsing/HttpMessageParser.scala
@@ -307,17 +307,20 @@ private[http] trait HttpMessageParser[Output >: MessageOutput <: ParserOutput] {
         s"HTTP chunk extension length exceeds configured limit of ${settings.maxChunkExtLength} characters")
 
     @tailrec def parseSize(cursor: Int, size: Long): StateResult =
-      if (size <= settings.maxChunkSize) {
+      if (size <= Int.MaxValue) {
         byteChar(input, cursor) match {
           case c if CharacterClasses.HEXDIG(c) => parseSize(cursor + 1, size * 16 + CharUtils.hexValue(c))
-          case ';' if cursor > offset          => parseChunkExtensions(size.toInt, cursor + 1)()
+          case c if size > settings.maxChunkSize =>
+            failEntityStream(
+              s"HTTP chunk of $size bytes exceeds the configured limit of ${settings.maxChunkSize} bytes")
+          case ';' if cursor > offset => parseChunkExtensions(size.toInt, cursor + 1)()
           case '\r' if cursor > offset && byteChar(input, cursor + 1) == '\n' =>
             parseChunkBody(size.toInt, "", cursor + 2)
           case '\n' if cursor > offset      => parseChunkBody(size.toInt, "", cursor + 1)
           case c if CharacterClasses.WSP(c) => parseSize(cursor + 1, size) // illegal according to the spec but can happen, see issue #1812
           case c                            => failEntityStream(s"Illegal character '${escape(c)}' in chunk start")
         }
-      } else failEntityStream(s"HTTP chunk size exceeds the configured limit of ${settings.maxChunkSize} bytes")
+      } else failEntityStream(s"HTTP chunk size exceeds ${Int.MaxValue} bytes")
 
     try parseSize(offset, 0)
     catch {
--- a/http-core/src/test/scala/org/apache/pekko/http/impl/engine/parsing/RequestParserSpec.scala
+++ b/http-core/src/test/scala/org/apache/pekko/http/impl/engine/parsing/RequestParserSpec.scala
@@ -547,7 +547,8 @@ abstract class RequestParserSpec(mode: String, newLine: String) extends AnyFreeS
           """400000
             |""") should generalMultiParseTo(
           Right(baseRequest),
-          Left(EntityStreamError(ErrorInfo("HTTP chunk of 4194304 bytes exceeds the configured limit of 1048576 bytes"))))
+          Left(
+            EntityStreamError(ErrorInfo("HTTP chunk of 4194304 bytes exceeds the configured limit of 1048576 bytes"))))
         closeAfterResponseCompletion shouldEqual Seq(false)
       }
 
--- a/http-core/src/test/scala/org/apache/pekko/http/impl/engine/parsing/RequestParserSpec.scala
+++ b/http-core/src/test/scala/org/apache/pekko/http/impl/engine/parsing/RequestParserSpec.scala
@@ -537,7 +537,7 @@ abstract class RequestParserSpec(mode: String, newLine: String) extends AnyFreeS
           """1a2b3c4d5e
             |""") should generalMultiParseTo(
           Right(baseRequest),
-          Left(EntityStreamError(ErrorInfo("HTTP chunk size exceeds 2147483647 bytes"))))
+          Left(EntityStreamError(ErrorInfo("HTTP chunk size exceeds Integer.MAX_VALUE (2147483647) bytes"))))
         closeAfterResponseCompletion shouldEqual Seq(false)
       }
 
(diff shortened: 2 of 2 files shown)


http-core/src/main/scala/org/apache/pekko/http/impl/engine/parsing/HttpMessageParser.scala
```diff
@@ -307,17 +307,20 @@ private[http] trait HttpMessageParser[Output >: MessageOutput <: ParserOutput] {
...
+              s"HTTP chunk of $size bytes exceeds the configured limit of ${settings.maxChunkSize} bytes")
+          case ';' if cursor > offset => parseChunkExtensions(size.toInt, cursor + 1)()
           case '\r' if cursor > offset && byteChar(input, cursor + 1) == '\n' =>
             parseChunkBody(size.toInt, "", cursor + 2)
           case '\n' if cursor > offset      => parseChunkBody(size.toInt, "", cursor + 1)
           case c if CharacterClasses.WSP(c) => parseSize(cursor + 1, size) // illegal according to the spec but can happen, see issue #1812
           case c                            => failEntityStream(s"Illegal character '${escape(c)}' in chunk start")
         }
-      } else failEntityStream(s"HTTP chunk size exceeds the configured limit of ${settings.maxChunkSize} bytes")
+      } else failEntityStream(s"HTTP chunk size exceeds ${Int.MaxValue} bytes")
```
pjfanning, line 323: Can you include the text `Int.MAX_VALUE` in this message too? Some users may not recognise the numeric value and what it means - still possibly useful to have the actual number too.

SkyTrix: Sure. I believe it should either be `Int.MaxValue` or `Integer.MAX_VALUE` though. Probably the latter due to Java interoperability?

pjfanning: Integer.MAX_VALUE might be better

http-core/src/main/scala/org/apache/pekko/http/impl/engine/parsing/HttpMessageParser.scala
```diff
@@ -307,17 +307,20 @@ private[http] trait HttpMessageParser[Output >: MessageOutput <: ParserOutput] {
...
+              s"HTTP chunk of $size bytes exceeds the configured limit of ${settings.maxChunkSize} bytes")
+          case ';' if cursor > offset => parseChunkExtensions(size.toInt, cursor + 1)()
           case '\r' if cursor > offset && byteChar(input, cursor + 1) == '\n' =>
             parseChunkBody(size.toInt, "", cursor + 2)
           case '\n' if cursor > offset      => parseChunkBody(size.toInt, "", cursor + 1)
           case c if CharacterClasses.WSP(c) => parseSize(cursor + 1, size) // illegal according to the spec but can happen, see issue #1812
           case c                            => failEntityStream(s"Illegal character '${escape(c)}' in chunk start")
         }
-      } else failEntityStream(s"HTTP chunk size exceeds the configured limit of ${settings.maxChunkSize} bytes")
+      } else failEntityStream(s"HTTP chunk size exceeds Integer.MAX_VALUE (${Int.MaxValue}) bytes")
```
He-Pin, line 323: I don't think we can have such a big chunk in memory, and if we do want to know how many bytes still there in the chunk, we can just logging:` there is still $n bytes remaining maybe maybe more, increase `settings.maxChunkSize` if you want to parse the full chunk`.

Parse it until the `Int.MaxValue` seems some kind of ddos for me, @jrudolph cc

SkyTrix: Note that this does not start parsing the actual chunk body. It only reads the size value that's written in the chunk header and bails if that's larger than `Int.MaxValue`. It will never attempt to parse anything larger than `settings.maxChunkSize`.

He-Pin: And eg, In Netty, there is a `maxContentLength` to be configured and user can config how to response the when oversized with `handleOversizedMessage`.see : `HttpObjectAggregator`

He-Pin: I did another around of check locally,it's ok.

mdedetrich (approved):
This seems good to me

pjfanning (approved):
lgtm

http-core/src/main/scala/org/apache/pekko/http/impl/engine/parsing/HttpMessageParser.scala
```diff
@@ -307,17 +307,20 @@ private[http] trait HttpMessageParser[Output >: MessageOutput <: ParserOutput] {
...
 
     @tailrec def parseSize(cursor: Int, size: Long): StateResult =
-      if (size <= settings.maxChunkSize) {
+      if (size <= Int.MaxValue) {
         byteChar(input, cursor) match {
           case c if CharacterClasses.HEXDIG(c) => parseSize(cursor + 1, size * 16 + CharUtils.hexValue(c))
-          case ';' if cursor > offset          => parseChunkExtensions(size.toInt, cursor + 1)()
+          case c if size > settings.maxChunkSize =>
+            failEntityStream(
+              s"HTTP chunk of $size bytes exceeds the configured limit of ${settings.maxChunkSize} bytes")
```
He-Pin, line 315: I see now, It was try decoding how big the chunk size is, and it will stop parsing how if the size already exceed MaxValue.

He-Pin: How about `HTTP chunk size:$size exceeds the configured limit of ${settings.maxChunkSize} bytes`

http-core/src/main/scala/org/apache/pekko/http/impl/engine/parsing/HttpMessageParser.scala
He-Pin: In netty, when oversize, it invoke a protect method `handleOversizedMessage`, maybe we can introduce something like that too.

He-Pin (approved):
lgtm after check the code in an IDE

http-core/src/main/scala/org/apache/pekko/http/impl/engine/parsing/HttpMessageParser.scala
```diff
@@ -307,17 +307,20 @@ private[http] trait HttpMessageParser[Output >: MessageOutput <: ParserOutput] {
         s"HTTP chunk extension length exceeds configured limit of ${settings.maxChunkExtLength} characters")
 
     @tailrec def parseSize(cursor: Int, size: Long): StateResult =
-      if (size <= settings.maxChunkSize) {
+      if (size <= Int.MaxValue) {
```
samueleresca, line 310: My personal preference would be to have:
```
if (size > Int.MaxValue) {
// early exit here with failEntityStream
}

// everything else

```

But I don't feel is binding.

samueleresca (approved):
Looks good to me

http-core/src/main/scala/org/apache/pekko/http/impl/engine/parsing/HttpMessageParser.scala
Roiocam: guard clauses. 👍🏻

http-core/src/test/scala/org/apache/pekko/http/impl/engine/parsing/RequestParserSpec.scala
```diff
@@ -531,13 +531,24 @@ abstract class RequestParserSpec(mode: String, newLine: String) extends AnyFreeS
...
-          Left(EntityStreamError(ErrorInfo("HTTP chunk size exceeds the configured limit of 1048576 bytes"))))
+          Left(EntityStreamError(ErrorInfo("HTTP chunk size exceeds Integer.MAX_VALUE (2147483647) bytes"))))
+        closeAfterResponseCompletion shouldEqual Seq(false)
+      }
+
+      "too-large chunk size" in new Test {
+        Seq(
+          start,
+          """400000
+            |""") should generalMultiParseTo(
```
jxnu-liguobin, line 548:
```suggestion
          "400000") should generalMultiParseTo(
```

http-core/src/test/scala/org/apache/pekko/http/impl/engine/parsing/RequestParserSpec.scala
```diff
@@ -531,13 +531,24 @@ abstract class RequestParserSpec(mode: String, newLine: String) extends AnyFreeS
         closeAfterResponseCompletion shouldEqual Seq(false)
       }
 
-      "too-large chunk size" in new Test {
+      "too-large chunk size (> Int.MaxValue)" in new Test {
         Seq(
           start,
           """1a2b3c4d5e
```
laglangyue, line 537: "1a2b3c4d5e"

http-core/src/main/scala/org/apache/pekko/http/impl/engine/parsing/HttpMessageParser.scala
SkyTrix: I'm trying to keep the structure consistent with the rest of the file. I also think the above suggestion would be incorrect in the sense that `everything else` would still be executed after calling `failEntityStream`? An else block would still be required.

http-core/src/test/scala/org/apache/pekko/http/impl/engine/parsing/RequestParserSpec.scala
SkyTrix: I found that the trailing newline is significant in this and many other tests in the file.

SkyTrix (commented):
Could a maintainer please shed some light on what it would still take to get this merged? There are currently 6 approvals but also some comments. I don't want to invalidate approvals by making new changes if that wouldn't be required.

pjfanning:
Merged - thanks.

It can be refactored in new PRs.

PR #529, opened 2024-04-02
Improve too large chunk size test

SkyTrix:
This is a followup of #528

The current test goes over the configured max chunk size when parsing the last hex char. The spirit of the code is to continue parsing the size even if we've already gone over the max chunk size. To actually test this we need to make sure another char follows after we've gone over the max chunk size.


He-Pin (approved):
Lgtm

pjfanning:
@SkyTrix how long does this test take to run? A couple of seconds is ok but we ideally would not have long running tests.

SkyTrix:
> @SkyTrix how long does this test take to run? A couple of seconds is ok but we ideally would not have long running tests.

This test, like most others in this file, is pretty simple and runs in 5 milliseconds on my machine.

He-Pin:
Maybe the value can be adjusted with the configuration value read from the default configuration, better than hard cording.

pjfanning (approved):
lgtm - if the test still runs quickly, it is ok to upgrade

PR #534, opened 2024-04-11
fix deprecation warning in Untar.scala

pjfanning:
```
project/Untar.scala:34:25: method getNextTarEntry in class TarArchiveInputStream is deprecated
[warn]         var entry = tis.getNextTarEntry
```


He-Pin (approved):
Lgtm

PR #536, opened 2024-04-11
Add munit test framework

Philippus:
Applying this PR will add support for testing routes with MUnit.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -59,6 +59,7 @@ lazy val userProjects: Seq[ProjectReference] = List[ProjectReference](
   httpCaching,
   httpCors,
   httpTestkit,
+  httpTestkitMunit,
   httpMarshallersScala,
   httpMarshallersJava,
   httpSprayJson,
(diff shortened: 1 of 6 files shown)


He-Pin:
Thanks, but why a new test framework ?

pjfanning:
I'd prefer not to add more dependencies to the main testkit. Could this go into a new testkit extension module? Something like pekko-http-testkit-munit.

http-testkit/src/main/scala/org/apache/pekko/http/scaladsl/testkit/MunitRouteTest.scala
```diff
@@ -0,0 +1,16 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * license agreements; and to You under the Apache License, version 2.0:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
```
pjfanning, line 6: new code should get the standard ASF header - https://www.apache.org/legal/src-headers.html#headers

mdedetrich:
To me the issue is less about whether adding the dependency but more about whether the dependency itself is stable or not (talking about bincompat here).

If munit breaks fairly often, adding it to Pekko will create more problems then it solves

laglangyue:
Agree that this is an extension module

Philippus:
> To me the issue is less about whether adding the dependency but more about whether the dependency itself is stable or not (talking about bincompat here).
> 
> If munit breaks fairly often, adding it to Pekko will create more problems then it solves

The testkit doesn't guarantee binary compatibility: https://github.com/apache/pekko-http/blob/main/build.sbt#L225

Philippus:
> I'd prefer not to add more dependencies to the main testkit. Could this go into a new testkit extension module? Something like pekko-http-testkit-munit.

I'll rework the PR and do that.

Philippus:
> Thanks, but why a new test framework ?

This has been setting in our private repo for a while now, I thought it would be a nice addition as afaics MUnit is gaining some traction.

Philippus:
I separated it into a module, please let me know if it is ok like this.

pjfanning (approved):
lgtm

pjfanning:
Thanks @Philippus - merging

Would you have any interest in making a similar PR for main Pekko repo to add a pekko-testkit-munit? There are a few more testkits there too (streams, actor-typed, persistence).

Philippus:
> Thanks @Philippus - merging
> 
> Would you have any interest in making a similar PR for main Pekko repo to add a pekko-testkit-munit? There are a few more testkits there too (streams, actor-typed, persistence).

There are more testkits, but I think only actor-typed and pekko-connectors-kafka have equivalent scalatest-specific functionality.

pjfanning:
@Philippus It isn't strictly necessary but could you consider supplying an iCLA (if you haven't done so already)?

https://www.apache.org/licenses/contributor-agreements.html

pjfanning:
Thanks @Philippus - your iCLA was received by the ASF Secretary.

PR #539, opened 2024-04-19
use optimised ByteString.asInputStream

pjfanning:
* ByteString.asInputStream is only in Pekko 1.1 - so if you use this code with Pekko 1.0, the MethodHandle will not work and the code falls back to existing ByteArrayInputStream

### DIFF
--- a/http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/hpack/ByteStringInputStream.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/hpack/ByteStringInputStream.scala
@@ -27,7 +27,7 @@ import pekko.util.ByteString.ByteString1C
 @InternalApi
 private[http2] object ByteStringInputStream {
 
-  private lazy val byteStringInputStreamMethodTypeOpt = Try {
+  private val byteStringInputStreamMethodTypeOpt = Try {
     val lookup = MethodHandles.publicLookup()
     val inputStreamMethodType = MethodType.methodType(classOf[InputStream])
     lookup.findVirtual(classOf[ByteString], "asInputStream", inputStreamMethodType)
--- a/http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/hpack/ByteStringInputStream.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/hpack/ByteStringInputStream.scala
@@ -14,6 +14,9 @@
 package org.apache.pekko.http.impl.engine.http2.hpack
 
 import java.io.{ ByteArrayInputStream, InputStream }
+import java.lang.invoke.{ MethodHandles, MethodType }
+
+import scala.util.Try
 
 import org.apache.pekko
 import pekko.annotation.InternalApi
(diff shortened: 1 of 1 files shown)


pjfanning:
I benchmarked this using https://github.com/pjfanning/pekko-test/blob/main/bench/src/main/scala/org/example/pekko/ByteStringBench.scala

The new code (in this PR) was faster for byte strings that are made up of multiple byte strings but slower for simple byte strings.

```
[info] Benchmark                                 Mode  Cnt       Score        Error  Units
[info] ByteStringBench.newGetInputStream        thrpt    3     752.065 ±    787.928  ops/s
[info] ByteStringBench.newGetInputStreamSimple  thrpt    3  447966.746 ± 273762.965  ops/s
[info] ByteStringBench.oldGetInputStream        thrpt    3     535.979 ±   1080.417  ops/s
[info] ByteStringBench.oldGetInputStreamSimple  thrpt    3  486935.775 ± 222571.828  ops/s
```

While the speedup for complex byte strings is good and the slowdown for simple byte strings is proportionally small - it might not be worth the hit.

I presume that the MethodHandle adds an overhead.

pjfanning:
Now have after https://github.com/pjfanning/pekko-test/commit/a79e14f66cedce61d07275aa5639f13a9668af60

```
[info] Benchmark                                 Mode  Cnt       Score        Error  Units
[info] ByteStringBench.newGetInputStream        thrpt    3     816.413 ±    114.507  ops/s
[info] ByteStringBench.newGetInputStreamSimple  thrpt    3  492976.928 ± 145281.777  ops/s
[info] ByteStringBench.oldGetInputStream        thrpt    3     588.401 ±    117.303  ops/s
[info] ByteStringBench.oldGetInputStreamSimple  thrpt    3  490941.644 ± 119759.598  ops/s
```

http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/hpack/ByteStringInputStream.scala
```diff
@@ -24,13 +27,29 @@ import pekko.util.ByteString.ByteString1C
...
-        // TODO optimise, ByteString needs to expose InputStream (esp if array backed, nice!)
-        new ByteArrayInputStream(cs.toArrayUnsafe())
-      case _ =>
-        // NOTE: We actually measured recently, and compact + use array was pretty good usually
-        apply(bs.compact)
+  private lazy val byteStringInputStreamMethodTypeOpt = Try {
+    val lookup = MethodHandles.publicLookup()
+    val inputStreamMethodType = MethodType.methodType(classOf[InputStream])
+    lookup.findVirtual(classOf[ByteString], "asInputStream", inputStreamMethodType)
+  }.toOption
```
He-Pin, line 34: 1. I think we should drop the `lazy` and just make it a `val`.
2. the `toOption` will introduce an allocation, use `OptionVal`

pjfanning: this code is only called once in the whole lifetime of the JVM

http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/hpack/ByteStringInputStream.scala
```diff
@@ -24,13 +27,29 @@ import pekko.util.ByteString.ByteString1C
...
+
+  def apply(bs: ByteString): InputStream = bs match {
+    case cs: ByteString1C =>
+      new ByteArrayInputStream(cs.toArrayUnsafe())
```
He-Pin, line 38: Even the code is the same as the old, maybe a simple `instanceOf/asInstanceOf` will help the performance for this simple case.

http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/hpack/ByteStringInputStream.scala
pjfanning: I think it should be lazy because this code may need to run unless it is needed

He-Pin: But the `lazy` will introduce a `volatile`  iirc, I think this is a trade-off, just not sure if that will cause the problem in the native image.

He-Pin: The getorelse will introduce another closure creation, think we can just fallback to imperative style for performance reason

http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/hpack/ByteStringInputStream.scala
pjfanning: https://github.com/pjfanning/pekko-test/tree/instanceof -- this branch is slower than the main branch - the instanceof is slower than the pattern matching

http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/hpack/ByteStringInputStream.scala
pjfanning: I found that the getOrElse has a small cost so I made a change to remove it

laglangyue:
LGTM

He-Pin (approved):
Lgtm

PR #540, opened 2024-04-19
use ByteStringInputStream

pjfanning:
pekko-http already has a class that optimises getting an InputStream for a ByteString. We should use it for consistency.

I know this is test code but If anything, this improves our test coverage of ByteStringInputStream


He-Pin (approved):
lgtm

PR #542, opened 2024-04-23
use different article because spray.io is dead

pjfanning:
I'm not sure if https://www.baeldung.com/scala/magnet-pattern is ideal so if anyone has a better suggestion, pass it on.

### DIFF
--- a/docs/src/main/paradox/routing-dsl/directives/debugging-directives/logRequest.md
+++ b/docs/src/main/paradox/routing-dsl/directives/debugging-directives/logRequest.md
@@ -14,7 +14,7 @@ def logRequest(magnet: LoggingMagnet[HttpRequest => Unit]): Directive0
 
 The signature shown is simplified, the real signature uses magnets. <a id="^1" href="#1">[1]</a>
 
-> <a id="1" href="#^1">[1]</a> See [The Magnet Pattern](http://spray.io/blog/2012-12-13-the-magnet-pattern/) for an explanation of magnet-based overloading.
+> <a id="1" href="#^1">[1]</a> See [The Magnet Pattern](https://www.baeldung.com/scala/magnet-pattern) for an explanation of magnet-based overloading.
 
 @@@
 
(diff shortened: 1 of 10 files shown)


mdedetrich:
Is there an issue with using the original spray link but via wayback machine? I believe this is standard practice for such style references?

pjfanning:
I can see that using the Wayback link is ok in the short term but it does lead to users being asked for donations, etc.

mdedetrich:
> I can see that using the Wayback link is ok in the short term but it does lead to users being asked for donations, etc.

Seems like the least worst option to me, especially since the whole point of Wayback machine is to archive this content forever?

He-Pin (approved):
lgtm

He-Pin:
As the spray is there, I think this is fine, user would like to see the content, not the url

He-Pin:
As the spray is there, I think this is fine, user would like to see the content, not the url

laglangyue:
LGTM

sirthias:
I'm sorry that spray.io is dead now, but I was personally sitting on the domain and hosting cost for >10 years before eventually getting an offer for the domain from a nice someone from Korea who would like to use it for another not-for-profit project.
So I sold and transferred it two days ago.

Part of the deal is the keep the old artifact repo at http://repo.spray.io (which still sees a decent amount of traffic!) intact for some time longer, so that's not a problem.

The only thing of some value on the actual spray.io site was probably the post on the magnet pattern (from 2012!).
I'm completely fine with linking to the archive version of it, if that works for pekko.

An alternative could be to directly implant it into the pekko docs. [The source](https://github.com/spray/spray/blob/master/docs/blog/2012-12-13-the-magnet-pattern.rst) are still around and written in reStructuredText (my preferred format at the time), which shouldn't be too hard to transcribe to Markdown or similar.

pjfanning:
Thanks @sirthias for the quick and detailed response.

GitHub does a pretty decent job of rendering the `.rst` file too - so linking to that may be an option.

https://github.com/spray/spray/blob/master/docs/blog/2012-12-13-the-magnet-pattern.rst

I'm happy enough to use Wayback Machine link as this PR now does.

PR #543, opened 2024-04-25
use different article because spray.io is dead (#542)

pjfanning:
cherry pick f998dde534f7e06d75101418ec488f7f517c94ec #542


He-Pin (approved):
lgtm

PR #551, opened 2024-05-10
send notification email if CI job fails

pjfanning:
see https://issues.apache.org/jira/browse/INFRA-25782 and https://github.com/apache/infrastructure-asfyaml?tab=readme-ov-file#GHA_build_status


He-Pin (approved):
lgtm

PR #552, opened 2024-05-10
Update format.yml

pjfanning:
use scalafmt-native-action@v4 because it supports Node 20 (and Node 16 support is being withdrawn by GitHub CI)


He-Pin (approved):
Lgtm

Issue #554, opened 2024-05-11
fix compiler warnings about using Java `@Deprecated` annotation


pjfanning:
e.g.

```
[warn] /Users/pj.fanning/code/incubator-pekko-http/http-core/src/main/scala/org/apache/pekko/http/scaladsl/settings/ServerSettings.scala:142:4: Prefer the Scala annotation over Java's `@Deprecated` to provide a message and version: @deprecated("message", since = "MyLib 1.0")
[warn]   @Deprecated @deprecated("Kept for binary compatibility; Use websocketSettings.withRandomFactoryFactory instead",
```

pjfanning:
Aim is to change the compiler settings to quieten these warnings

PR #555, opened 2024-05-11
compiler warnings about Deprecated annotation

pjfanning:
see #554

### DIFF
--- a/http-core/src/main/scala/org/apache/pekko/http/javadsl/ConnectionContext.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/javadsl/ConnectionContext.scala
@@ -60,14 +60,14 @@ object ConnectionContext {
 
   // ConnectionContext
   /** Used to serve HTTPS traffic. */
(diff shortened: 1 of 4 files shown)


He-Pin:
Not sure Idea will support this, I think it was there for Java user？

He-Pin:
Lgtm，We can starting remove them.after.1.1.x

He-Pin (approved):

laglangyue:
lgtm,but ci failed

He-Pin (approved):

raboof (commented):
These are Java DSLs. I think they were originally there to make sure any Java tooling would also see the Deprecated state of these methods.

I seem to remember perhaps newer Scala compiler versions will automatically emit a Java `@Deprecated` annotation when you use the Scala `@deprecated` annotation. Have we checked that this is the case of all Scala compiler versions that we build with?

He-Pin:
The question is ,Will scala generate the `@Deprecated` annotation in classfile too ,and since Java 9, the java one can have a `since` filed too.

Issue #558, opened 2024-05-18
FileAndResourceDirectivesSpec fails when scala 2.12 used


pjfanning:
See https://github.com/apache/pekko-http/actions/runs/9136372804/job/25125085205

Fails regardless of Pekko or Java version.

```
  [info] FileAndResourceDirectivesSpec:
  [info] org.apache.pekko.http.scaladsl.server.directives.FileAndResourceDirectivesSpec *** ABORTED *** (0 milliseconds)
  [info]   java.lang.IllegalArgumentException was thrown inside "listDirectoryContents" should, construction cannot continue: "URI is not hierarchical" (FileAndResourceDirectivesSpec.scala:327)
  [info]   org.scalatest.exceptions.NotAllowedException:
  [info]   at org.scalatest.wordspec.AnyWordSpecLike.org$scalatest$wordspec$AnyWordSpecLike$$registerBranch(AnyWordSpecLike.scala:226)
  [info]   at org.scalatest.wordspec.AnyWordSpecLike$$anon$1.apply(AnyWordSpecLike.scala:1158)
  [info]   at org.scalatest.verbs.ShouldVerb$StringShouldWrapperForVerb.should(ShouldVerb.scala:194)
  [info]   at org.scalatest.verbs.ShouldVerb$StringShouldWrapperForVerb.should$(ShouldVerb.scala:193)
  [info]   at org.scalatest.matchers.should.Matchers$StringShouldWrapper.should(Matchers.scala:8242)
  [info]   at org.apache.pekko.http.scaladsl.server.directives.FileAndResourceDirectivesSpec.<init>(FileAndResourceDirectivesSpec.scala:327)
  [info]   at java.base/jdk.internal.reflect.DirectConstructorHandleAccessor.newInstance(DirectConstructorHandleAccessor.java:62)
  [info]   at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:502)
  [info]   at java.base/java.lang.reflect.ReflectAccess.newInstance(ReflectAccess.java:128)
  [info]   at java.base/jdk.internal.reflect.ReflectionFactory.newInstance(ReflectionFactory.java:304)
  [info]   at java.base/java.lang.Class.newInstance(Class.java:725)
  [info]   at org.scalatest.tools.Framework$ScalaTestTask.execute(Framework.scala:454)
  [info]   at sbt.TestRunner.runTest$1(TestFramework.scala:153)
[...]

PR #559, opened 2024-05-18
fix issue in listDirectoryContents test

pjfanning:
see #558 

the issue may be down to classloading changes in sbt 1.10.0 - only recent change in this repo

still, the testRoot val seems tidier and works fine when tested on my laptop


He-Pin (approved):
Lgtm

PR #575, opened 2024-07-09
fix parse broken when header value is null

### LINKED ISSUE #574: Uncaught NPE when parsing a header value
I'm using Akka 10.2.9, but the issue (as in, broken error handling) doesn't seem to have been fixed in Pekko as well. Sorry I don't have a reproduction example at this moment, I'm creating this issue to have something to reference when rolling out a patch that will hopefully give me a clue. :smile: 

I have a piece of code like this:
```scala
`Content-Disposition`.parseFromValueString(headerValue)
```

I expect it to return an `Either`. Instead, for some input values it seems to throw the following exception:
```
java.lang.NullPointerException: Cannot invoke "String.split(String, int)" because "message" is null
	at akka.http.scaladsl.model.ErrorInfo$.fromCompoundString(ErrorInfo.scala:65)
	at akka.http.impl.model.parser.HeaderParser.failure(HeaderParser.scala:77)
	at akka.http.scaladsl.model.HttpHeader$.parse(HttpHeader.scala:92)
	at akka.http.scaladsl.model.headers.ModeledCompanion.parseFromValueString(headers.scala:40)
        <redacted>
```
It seems that there is an exception raised somewhere inside the parser that doesn't have its message set. [Here](https://github.com/akka/akka-http/blob/v10.2.9/akka-http-core/src/main/scala/akka/http/impl/model/parser/HeaderParser.scala#L77) we do `.getMessage`, passing `null` into `ErrorInfo.fromCompoundString`. It [tries to split it](https://github.com/akka/akka-http/blob/v10.2.9/akka-http-core/src/main/scala/akka/http/scaladsl/model/ErrorInfo.scala#L65), and we get an NPE.

laglangyue:
## motivation
fix #574
## test
by ci

### DIFF
--- a/http-core/src/main/scala/org/apache/pekko/http/impl/model/parser/HeaderParser.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/impl/model/parser/HeaderParser.scala
@@ -87,7 +87,7 @@ private[http] class HeaderParser(
     HeaderParser.Failure {
       error match {
         case IllegalUriException(info) => info
-        case NonFatal(e)               => ErrorInfo.fromCompoundString(if (e.getMessage == null) "" else e.getMessage)
+        case NonFatal(e)               => ErrorInfo.fromCompoundString(e.getMessage)
       }
     }
   def ruleNotFound(ruleName: String): Result = HeaderParser.RuleNotFound
--- a/http-core/src/main/scala/org/apache/pekko/http/impl/model/parser/HeaderParser.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/impl/model/parser/HeaderParser.scala
@@ -17,10 +17,7 @@ import org.apache.pekko
 import pekko.annotation.InternalApi
 import pekko.http.scaladsl.settings.ParserSettings
 import pekko.http.scaladsl.settings.ParserSettings.CookieParsingMode
-import pekko.http.scaladsl.settings.ParserSettings.{
-  IllegalResponseHeaderNameProcessingMode,
-  IllegalResponseHeaderValueProcessingMode
-}
+import pekko.http.scaladsl.settings.ParserSettings.{IllegalResponseHeaderNameProcessingMode, IllegalResponseHeaderValueProcessingMode}
 import pekko.http.scaladsl.model.headers.HttpCookiePair
 import pekko.util.ConstantFun
 
@@ -87,7 +84,7 @@ private[http] class HeaderParser(
     HeaderParser.Failure {
       error match {
         case IllegalUriException(info) => info
-        case NonFatal(e)               => ErrorInfo.fromCompoundString(e.getMessage)
+        case NonFatal(e)               => ErrorInfo.fromCompoundString(if(e.getMessage == null) "" else e.getMessage)
       }
     }
   def ruleNotFound(ruleName: String): Result = HeaderParser.RuleNotFound
--- a/http-core/src/main/scala/org/apache/pekko/http/impl/model/parser/HeaderParser.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/impl/model/parser/HeaderParser.scala
@@ -17,7 +17,10 @@ import org.apache.pekko
 import pekko.annotation.InternalApi
 import pekko.http.scaladsl.settings.ParserSettings
 import pekko.http.scaladsl.settings.ParserSettings.CookieParsingMode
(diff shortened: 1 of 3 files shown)


http-core/src/main/scala/org/apache/pekko/http/scaladsl/model/HttpHeader.scala
```diff
@@ -99,10 +99,11 @@ object HttpHeader {
...
           }
         case Failure(error) =>
-          val info = (error match {
-            case e: ParseError => parser.parseError(e)
-            case e             => parser.failure(e)
-          }).info
+          val info = error match {
+            case e: ParseError           => parser.parseError(e).info
+            case _: NullPointerException => ErrorInfo()
+            case e                       => parser.failure(e).info
```
pjfanning, line 105: how about `case e if e.getMessage() == null => ???` ?

This is useful too.

laglangyue: yeah, done

http-core/src/main/scala/org/apache/pekko/http/scaladsl/model/HttpHeader.scala
```diff
@@ -99,10 +99,11 @@ object HttpHeader {
             case HeaderParser.RuleNotFound => ParsingResult.Ok(RawHeader(name, preProcessedValue), Nil)
           }
         case Failure(error) =>
-          val info = (error match {
-            case e: ParseError => parser.parseError(e)
-            case e             => parser.failure(e)
-          }).info
+          val info = error match {
+            case e: ParseError             => parser.parseError(e).info
+            case e if e.getMessage == null => ErrorInfo()
```
pjfanning, line 104: I think this code should be reverted and the change should be in HeaderParser.scala instead. I have the change in #576. I created 576 to test if that approach works but I'd prefer to let you complete this PR. View 576 as a set of suggested changes for this PR.

http-core/src/test/scala/org/apache/pekko/http/impl/model/parser/HttpHeaderSpec.scala
```diff
@@ -876,6 +876,13 @@ class HttpHeaderSpec extends AnyFreeSpec with Matchers {
       parse("User-Agent", "(" * 10000).errors.head shouldEqual
       ErrorInfo("Illegal HTTP header 'User-Agent': Illegal header value", "Header comment nested too deeply")
     }
+
+    "should not broken when header-value is null" in {
+      val errors = parse("Content-Disposition", null).errors
+      errors.size shouldBe 1
```
pjfanning, line 882: `errors should have size 1` is a better way to do this. When this fails, scalatest will say something like 
`List(a1, a2) did not have size 1`

With `errors.size shouldBe 1` fails, you will get an error like `2 was not 1`. I much prefer getting the detailed message from my assertion. I would strongly recommend that you read up on scalatest. The assertions are carefully designed to give us lots of context when assertions fail. But if you go off and use the simplest assertions then you get very little info when things fail.

laglangyue: done

pjfanning (approved):
Lgtm

http-core/src/main/scala/org/apache/pekko/http/impl/model/parser/HeaderParser.scala
```diff
@@ -86,8 +86,9 @@ private[http] class HeaderParser(
   def failure(error: Throwable): HeaderParser.Failure =
     HeaderParser.Failure {
       error match {
-        case IllegalUriException(info) => info
-        case NonFatal(e)               => ErrorInfo.fromCompoundString(e.getMessage)
+        case IllegalUriException(info)           => info
+        case NonFatal(e) if e.getMessage == null => ErrorInfo.fromCompoundString(e.toString)
```
He-Pin, line 90: Can use eq

He-Pin (approved):

http-core/src/main/scala/org/apache/pekko/http/impl/model/parser/HeaderParser.scala
laglangyue: `e.getMessage eq null` ? I don't think it is  good. `eq` is used to replace the equals method, and when compared to null, it is usually to use `==`

pjfanning:
Thanks @laglangyue - could you create a new PR that cherry picks this commit to the 1.0.x branch?

PR #578, opened 2024-07-16
remove unnecessary interpolation

pjfanning:
* also tidy up test name from #574 - the 'should' is already picked up from code wrapping this test


Roiocam (approved):
LGTM

PR #581, opened 2024-07-19
fix parse broken when header value is null (#575)

pjfanning:
cherry pick 41e4f46e35f4ef90a30b5d6a924413a83c3fa110 #575

issue https://github.com/apache/pekko-http/issues/574


He-Pin (approved):
Lgtm

Issue #300, opened 2023-08-09
Internal server error triggered by “Accept-Charset” header with unexpected value


He-Pin:
Picked from: https://discuss.lightbend.com/t/internal-server-error-triggered-by-accept-charset-header-with-unexpected-value/10427

Which expects the header to be ignored.

pjfanning:
Could you ask for the stacktrace so that we don't have to run the samples for ourselves?

mdedetrich:
> As long as the unexpected header doesn't cause a server side crash - this would likely to be quite a minor issue.

If we are leaking a stack trace to the client this is a serious security issue

jrudolph:
Yes, this is a bug but only a minor issue.

Here's a stack trace (though, wrapped with another exception to understand how that exception can be triggered):

```
[info] java.lang.IllegalArgumentException: Unsupported charset: asd
[info] 	at org.apache.pekko.http.scaladsl.model.HttpCharset$$anonfun$nioCharset$1.applyOrElse(HttpCharset.scala:68)
[info] 	at org.apache.pekko.http.scaladsl.model.HttpCharset$$anonfun$nioCharset$1.applyOrElse(HttpCharset.scala:67)
[info] 	at scala.util.Failure.recover(Try.scala:233)
[info] 	at org.apache.pekko.http.scaladsl.model.HttpCharset.nioCharset(HttpCharset.scala:67)
[info] 	at org.apache.pekko.http.scaladsl.model.HttpEntity$.apply(HttpEntity.scala:310)
[info] 	at org.apache.pekko.http.scaladsl.marshalling.PredefinedToEntityMarshallers.$anonfun$stringMarshaller$1(PredefinedToEntityMarshallers.scala:58)
[info] 	at org.apache.pekko.http.scaladsl.marshalling.Marshaller$$anon$3.$anonfun$apply$2(Marshaller.scala:161
[info] 	at org.apache.pekko.http.scaladsl.marshalling.Marshalling$WithOpenCharset.$anonfun$map$6(Marshaller.scala:219)
[info] 	at org.apache.pekko.http.scaladsl.marshalling.Marshal$$anonfun$selectMarshallingForContentType$2.$anonfun$applyOrElse$1(Marshal.scala:39)
[info] 	at org.apache.pekko.http.scaladsl.marshalling.Marshal.$anonfun$toResponseFor$1(Marshal.scala:84)
[...]

PR #584, opened 2024-08-12
handle invalid accept-charset in requests - default to utf-8

pjfanning:
see #300 

I've added #585 to look into the XML case - this PR focuses on text/plain.

### DIFF
--- a/http-tests/src/test/scala/org/apache/pekko/http/scaladsl/server/directives/MarshallingDirectivesSpec.scala
+++ b/http-tests/src/test/scala/org/apache/pekko/http/scaladsl/server/directives/MarshallingDirectivesSpec.scala
@@ -250,6 +250,11 @@ class MarshallingDirectivesSpec extends RoutingSpec with Inside {
         rejection shouldEqual UnacceptedResponseContentTypeRejection(Set(ContentType(`application/json`)))
       }
     }
+    "reject JSON rendering if an `Accept-Charset` request header requests an unknown encoding" in {
+      Get() ~> `Accept-Charset`(HttpCharset("unknown")(Nil)) ~> complete(foo) ~> check {
+        rejection shouldEqual UnacceptedResponseContentTypeRejection(Set(ContentType(`application/json`)))
+      }
+    }
     val acceptHeaderUtf = Accept.parseFromValueString("application/json;charset=utf8").right.get
     val acceptHeaderNonUtf = Accept.parseFromValueString("application/json;charset=ISO-8859-1").right.get
     "render JSON response when `Accept` header is present with the `charset` parameter ignoring it" in {
@@ -291,6 +296,26 @@ class MarshallingDirectivesSpec extends RoutingSpec with Inside {
     "render text with UTF-8 encoding if an `Accept-Charset` request header requests an unknown encoding" in {
       Get() ~> `Accept-Charset`(HttpCharset("unknown")(Nil)) ~> complete(foo) ~> check {
         responseEntity shouldEqual HttpEntity(ContentType(`text/plain`, `UTF-8`), foo)
+      }
+    }
+  }
+
+  "The marshalling infrastructure for text/xml" should {
+    val foo = <foo>Hällö</foo>
+    "render XML with UTF-8 encoding if no `Accept-Charset` request header is present" in {
+      Get() ~> complete(foo) ~> check {
+        responseEntity shouldEqual HttpEntity(ContentType(`text/xml`, `UTF-8`), foo.toString)
+      }
+    }
+    "render XML with requested encoding if an `Accept-Charset` request header requests a non-UTF-8 encoding" in {
+      Get() ~> `Accept-Charset`(`ISO-8859-1`) ~> complete(foo) ~> check {
+        responseEntity shouldEqual HttpEntity(ContentType(`text/xml`, `ISO-8859-1`), foo.toString)
+      }
+    }
+    "render XML with UTF-8 encoding if an `Accept-Charset` request header requests an unknown encoding" ignore {
+      // still returns Content-Type: text/xml; charset=unknown
+      Get() ~> `Accept-Charset`(HttpCharset("unknown")(Nil)) ~> complete(foo) ~> check {
+        responseEntity shouldEqual HttpEntity(ContentType(`text/xml`, `UTF-8`), foo.toString)
       }
     }
   }
--- a/http-tests/src/test/scala/org/apache/pekko/http/scaladsl/server/directives/MarshallingDirectivesSpec.scala
+++ b/http-tests/src/test/scala/org/apache/pekko/http/scaladsl/server/directives/MarshallingDirectivesSpec.scala
@@ -283,11 +283,15 @@ class MarshallingDirectivesSpec extends RoutingSpec with Inside {
         responseEntity shouldEqual HttpEntity(ContentType(`text/plain`, `UTF-8`), foo)
       }
     }
-    "rrender text with UTF-8 encoding if an `Accept-Charset` request header requests a non-UTF-8 encoding" in {
+    "render text with requested encoding if an `Accept-Charset` request header requests a non-UTF-8 encoding" in {
       Get() ~> `Accept-Charset`(`ISO-8859-1`) ~> complete(foo) ~> check {
         responseEntity shouldEqual HttpEntity(ContentType(`text/plain`, `ISO-8859-1`), foo)
       }
     }
-
+    "render text with UTF-8 encoding if an `Accept-Charset` request header requests an unknown encoding" in {
+      Get() ~> `Accept-Charset`(HttpCharset("unknown")(Nil)) ~> complete(foo) ~> check {
+        responseEntity shouldEqual HttpEntity(ContentType(`text/plain`, `UTF-8`), foo)
+      }
+    }
   }
 }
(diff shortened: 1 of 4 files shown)


http-tests/src/test/scala/org/apache/pekko/http/scaladsl/server/directives/MarshallingDirectivesSpec.scala
```diff
@@ -275,4 +275,23 @@ class MarshallingDirectivesSpec extends RoutingSpec with Inside {
...
+    "render text with requested encoding if an `Accept-Charset` request header requests a non-UTF-8 encoding" in {
+      Get() ~> `Accept-Charset`(`ISO-8859-1`) ~> complete(foo) ~> check {
+        responseEntity shouldEqual HttpEntity(ContentType(`text/plain`, `ISO-8859-1`), foo)
+      }
+    }
+    "render text with UTF-8 encoding if an `Accept-Charset` request header requests an unknown encoding" in {
+      Get() ~> `Accept-Charset`(HttpCharset("unknown")(Nil)) ~> complete(foo) ~> check {
+        responseEntity shouldEqual HttpEntity(ContentType(`text/plain`, `UTF-8`), foo)
+      }
+    }
```
raboof, line 295: I realize this is a comment on the current behavior, and not on your change, but this seems strange to me: it looks like the default string marshaller will return the string passed to it as `text/plain` claiming it is in whatever encoding that was requested by the user, without doing anything to make sure the string is actually in that encoding.

This is only correct if you assume the implementation correctly took into account the charset, which seems extremely unlikely. I wonder if it wouldn't make more sense to leave out the optional `charset` parameter to the return content type in this case. If people have specific requirements (which seems somewhat unlikely) they can specify the charset explicitly.

This would be a behavior change, but it seems like a reasonable one if we note it in the release notes.

pjfanning: before the change to the main source code in this PR, this test actually fails - the response is an error response

michaelf-crwd: Hi, 

Thanks for having a look at this bug / feature request. I recently sent in a duplicate issue (https://github.com/apache/pekko-http/issues/583) unaware of this one. I worked on a similar PR the past two days, and have now found and looked through your changes - that would seem to fix it nicely! Great. 

For your consideration, I'd like to suggest adding in an extra unit-test in "MarshallingSpac.scala" with something along the lines of:

```
val invalidAcceptCharsetHeader = `Accept-Charset`(HttpCharsetRange(HttpCharset.custom("invalid")))
val request = HttpRequest().withHeaders(Seq(invalidAcceptCharsetHeader))
marshalToResponse("Hello", request).entity.contentType.charsetOption should be(Some(HttpCharsets.`UTF-8`))
```

Some refactoring thoughts; Perhaps the try-catch in "PredefinedToEntityMarshallers.scala" could be more subtly placed in "HttpCharset.scala" as:

```
/** Returns this HttpCharset instance if the provided nio charset is available or a desired default otherwise */
  def withNioCharsetOrElse(default: HttpCharset): HttpCharset = {
    if (_nioCharset.isSuccess) {
      // Return this instance, as the provided nioCharset did not result in a java.nio.charset.UnsupportedCharsetException
      this
    } else {
      default
    }
  }
```

and then have the "PredefinedToEntityMarshallers.scala" do:

```
Marshaller.withOpenCharset(mediaType) { (s, cs) => {
  HttpEntity(mediaType.withCharset(cs.withNioCharsetOrElse(HttpCharsets.`UTF-8`)), s)
} }
```

Looking forward to seeing your changes merged.

Thanks again and best regards,
Michael

pjfanning: Thanks @michaelf-crwd - I have added a new HttpCharsets function similar to what you suggested (c8514e78919d4bc41cc459aee9cb88a0a94043a6).
I will look later at adding extra tests like the one you suggested.

pjfanning: @raboof would you be able to take a look at this again? I don't think this change breaks anything and makes the pekko-http support for `Accept-Charset` a bit more tolerant.

pjfanning: @mdedetrich would you have time to look at this? It's basically the last item to decide on before a 1.1.0 release.

mdedetrich: Ill check this tomorrow, there is context I need to grock

nvollmar (approved):
lgtm

raboof:
Sorry for not finding time to look into this sooner, will do so now.

http-tests/src/test/scala/org/apache/pekko/http/scaladsl/server/directives/MarshallingDirectivesSpec.scala
jrudolph: > without doing anything to make sure the string is actually in that charset

@raboof, WDYM with that? Java strings are basically UTF-16 encoded, so they are known to be unicode. What else could go wrong?

jrudolph (approved):
lgtm, I agree with the idea to treat the `Accept-XYZ` headers rather softly, especially `Accept-Charset` which is losing relevancy these days. As long as the server clearly indicates what representation it uses, it seems fine.

raboof (approved):
I tested various scenario's:

# Explicit charset in response

i.e. a route with `complete(HttpEntity(ContentTypes.`text/html(UTF-8)`, "Say Hällö to Pekko HTTP"))`:

- request: `curl -v localhost:8080/hello`
  old behavior: response in Content-Type: text/html; charset=UTF-8 :green_heart: 
  new behavior: same :green_heart: 
- request: `curl -v -H "Accept-Charset: blop" localhost:8080/hello`
  old behavior: 500 error and stacktrace in the logs :stop_sign: 
  new behavior: 406 Not Acceptable, "Resource representation is only available with these types: text/html; charset=UTF-8" :green_heart: 
- request: `curl -v -H "Accept-Charset: US-ASCII" localhost:8080/hello`
  old behavior: 406 Not Acceptable, "Resource representation is only available with these types: text/html; charset=UTF-8" :green_heart: 
  new behavior: same :green_heart: 
- curl -v -H "Accept-Charset: ISO-8859-1" localhost:8080/hello
  old -> response 406 Not Acceptable :green_heart: 
  new -> same :green_heart: 

# No explicit charset in response

i.e. a route with `complete("Say Hällö to Pekko HTTP")`:

with `complete("Say Hällö to Pekko HTTP")`
- request: `curl -v localhost:8080/hello`
  old behavior: response in `text/plain; charset=UTF-8` :green_heart: 
  new behavior: same :green_heart: 
- request: `curl -v -H "Accept-Charset: blop" localhost:8080/hello`
  old behavior: 500 error and stacktrace in the logs :stop_sign: 
  new behavior: response in `text/plain; charset=UTF-8` :green_heart: 
- request: `curl -v -H "Accept-Charset: US-ASCII" localhost:8080/hello`
  old behavior: response with `Content-Type: text/plain; charset=US-ASCII` header, ~but UTF-8 body~ response converted to US-ASCII ~:stop_sign:~ :green_heart: 
  new behavior: same ~:stop_sign:~ :green_heart: 
- curl -v -H "Accept-Charset: ISO-8859-1" localhost:8080/hello
  old -> response with `Content-Type: text/plain; charset=ISO-8859` header, ~but UTF-8 body~ response converted to US-ASCII ~:stop_sign:~ :green_heart: 
  new -> same ~:stop_sign:~ :green_heart: 

In summary, I think this PR is a definite improvement. ~Some behavior seems wrong to me, but that behavior was *already* like that, so that doesn't need to delay merging this PR or releasing 1.1.0. I'll put my code where my mouth is and propose a fix for the behavior that looks wrong to me, but that's independent of 1.1.0. Sorry for taking so long before confirming ;)~ (actually the strings are converted taking into the encodings just fine, sorry about the noise :facepalm: )

http-tests/src/test/scala/org/apache/pekko/http/scaladsl/server/directives/MarshallingDirectivesSpec.scala
raboof: what I meant is, when the route does `complete(some_unicode_string)`, and the marshalling logic results in a `Content-Type: text/plain; charset=ISO-8859-1` header, that response seems wrong: the header promises the body will be in ISO-8859-1, but the body is actually in unicode. I think it would make more sense to make sure we produce a `Content-Type: text/plain; charset=UTF-8` header unless the developer explicitly marked the response as being in a different charset.

raboof: or will the marshalling logic actually convert the unicode to the charset from the content type? in that case of course it's all good.

raboof: Indeed it does, sorry about the noise :facepalm:

jrudolph: :+1:

jrudolph:
> the strings are converted to bytes taking into account the encodings just fine, all good, sorry about the noise 🤦

The whole negotiation story is really sophisticated and tries to take that into account, indeed. It's also one of the performance hotspots of the routing layer.

jrudolph:
Anyway thanks for that comprehensive testing session, @raboof :+1:

pjfanning:
Thanks everyone. I'll merge as is but if anyone has any requests for changes, I'll look at them.

PR #590, opened 2024-09-04
Increase stream-cancellation-delay default to 1000 millis

mdedetrich:
References: https://github.com/apache/pekko-http/issues/422

Still need to update tests as per @jrudolph original comment, i.e.

> That broke lots of tests that probably didn't automatically propagate closing so they now fail while waiting for the now ten times longer cancellation times... (Might still might sense to "just" fix those tests)

### DIFF
--- a/http-core/src/test/resources/reference.conf
+++ b/http-core/src/test/resources/reference.conf
@@ -6,4 +6,5 @@ pekko {
     default-dispatcher.throughput = 1
   }
   stream.materializer.debug.fuzzing-mode = off
+  stream.testkit.all-stages-stopped-timeout = 20 s
 }
--- a/http2-tests/src/test/scala/org/apache/pekko/http/impl/engine/http2/H2SpecIntegrationSpec.scala
+++ b/http2-tests/src/test/scala/org/apache/pekko/http/impl/engine/http2/H2SpecIntegrationSpec.scala
@@ -306,6 +306,7 @@ class H2SpecIntegrationSpec extends PekkoFreeSpec(
         executable,
         "-k", "-t",
         "-p", port.toString,
+        "-o", "9",
         "-j", junitOutput.getPath) ++
         specSectionNumber.toList.map(number => s"http2/$number")
 
--- a/docs/src/main/paradox/release-notes/releases-1.1.md
+++ b/docs/src/main/paradox/release-notes/releases-1.1.md
@@ -12,6 +12,7 @@ All the changes in the @ref:[1.0.x releases](releases-1.0.md) up to and includin
 ### Changes
 * Changed names of HTTP status codes 413 and 422 ([PR87](https://github.com/apache/pekko-http/pull/87))
 * Parse entire HTTP chunk size ([PR528](https://github.com/apache/pekko-http/pull/528))
+* Increased default value of `pekko.http.server.stream-cancellation-delay` and `pekko.http.client.stream-cancellation-delay` from 100 to 1000 millis ([PR590](https://github.com/apache/pekko-http/pull/590))
 
 ### Additions
 * Add UnsupportedContentTypeException Java DSL ([PR376](https://github.com/apache/pekko-http/pull/376))
--- a/http-core/src/main/resources/reference.conf
+++ b/http-core/src/main/resources/reference.conf
@@ -222,7 +222,7 @@ pekko.http {
     # In most cases, there should be no reason to change this setting.
     #
     # Set to 0 to disable the delay.
-    stream-cancellation-delay = 100 millis
+    stream-cancellation-delay = 1000 millis
 
     http2 {
       # The maximum number of request per connection concurrently dispatched to the request handler.
@@ -532,7 +532,7 @@ pekko.http {
     # In most cases, there should be no reason to change this setting.
     #
     # Set to 0 to disable the delay.
-    stream-cancellation-delay = 100 millis
+    stream-cancellation-delay = 1000 millis
   }
   #client-settings
 


pjfanning:
Reading the Akka issue and PR seems to indicate that this will cause test failures. I can help if needed to try to fix up the test issues. I'm hoping to get Pekko HTTP 1.1.0 released but it seems useful to fix this.

mdedetrich:
> Reading the Akka issue and PR seems to indicate that this will cause test failures. I can help if needed to try to fix up the test issues. I'm hoping to get Pekko HTTP 1.1.0 released but it seems useful to fix this.

The test failures are expecting because the tests exect the stream to be cancelled within 100ms and now that is has been increased to 1000ms those tests are now broken (as per @jrudolph 's comment, the change itself is safe).

mdedetrich:
Ill have time to look into this tonight

JD557:
Should this actually be marked as closing the issue?
It's better than the current status quo, but the underlying issue is still there.

For reference, we've used `1s` for a long time with no issue until today, where we saw this resurface again. So `1s` is not bullet proof.

mdedetrich:
> Should this actually be marked as closing the issue? It's better than the current status quo, but the underlying issue is still there.
> 
> For reference, we've used `1s` for a long time with no issue until today, where we saw this resurface again. So `1s` is not bullet proof.

Fair point, ill change this so it won't close original issue

mdedetrich:
PR is ready, I managed to fix the issues with the tests.

http-core/src/test/resources/reference.conf
```diff
@@ -6,4 +6,5 @@ pekko {
     default-dispatcher.throughput = 1
   }
   stream.materializer.debug.fuzzing-mode = off
+  stream.testkit.all-stages-stopped-timeout = 20 s
```
mdedetrich, line 9: This timeout is related to `Utils.assertAllStagesStopped` which is designed to check that all stream stages finish when the test does. Since `stream-cancellation-delay` creates a stage in of itself this needs to be increased.

To me it makes sense to do this globally (for tests) rather than on a test by test basis because there is no issue in having it too long as it only increases the failure test case, not the positive.

The default value of this 5 s

http2-tests/src/test/scala/org/apache/pekko/http/impl/engine/http2/H2SpecIntegrationSpec.scala
```diff
@@ -306,6 +306,7 @@ class H2SpecIntegrationSpec extends PekkoFreeSpec(
         executable,
         "-k", "-t",
         "-p", port.toString,
+        "-o", "9",
```
mdedetrich, line 309: See https://github.com/summerwind/h2spec?tab=readme-ov-file#usage for reference. The default value is 2 seconds, I got to 9 with manual testing by incrememting 1 each time.

pjfanning (approved):
lgtm

mdedetrich:
@pjfanning Feel free to merge when you see fit

http-core/src/test/resources/reference.conf
jrudolph: I cannot remember the reason to choose 100ms for the delay cancellation limit. It is somewhat of an indication of what kind of real world implications this change might have: cleanup of broken/closed connections might now take longer.

IIRC, the whole reason to introduce the delay was to deal with the fact that stream cancellation could not be well-configured in Akka 2.5. With 2.6, it's not 100% clear if the delay is even still needed or whether it could be solved in a better way (i.e. by resolving the cancellation race more intelligently, taking the cancellation reason into account at the right/more places) to avoid symptoms as shown in the ticket.

mdedetrich: > I cannot remember the reason to choose 100ms for the delay cancellation limit. Having to increase the all-stages-stopped-timeout to higher values is somewhat of an indication of what kind of real world implications this change might have: cleanup of broken/closed connections might now take longer.

iirc its only in a few tests which used `Utils.assertAllStagesStopped` started timing out so I can granularly update those tests. My ultimate reasoning for making this change global is was to reduce surprises for people writing future tests (i.e. giving more QoL to developers), more specifically avoiding the "I just updated/wrote a test and now its failing due to a finely tuned timeout and now I have to figure out that it was because of that said timeout"

I think that as long as you handle shutdown of pekko http gracefully I don't think it should cause too many issues however I admit that many people may not even be handlong shutdowns of pekko http correctly.

> IIRC, the whole reason to introduce the delay was to deal with the fact that stream cancellation could not be well-configured in Akka 2.5. With 2.6, it's not 100% clear if the delay is even still needed or whether it could be solved in a better way (i.e. by resolving the cancellation race more intelligently, taking the cancellation reason into account at the right/more places) to avoid symptoms as shown in the ticket.

If this is the case then it should definitely be looked into, I am not aware of the updates to stream cancellation in akka 2.6 but judging by the date of the report at https://github.com/apache/pekko-http/issues/422#issue-2084209600, if that stream cancellation update is now the default it didn't seem to fix the underlying issue.

@pjfanning Is it worth it to try and solve this properly before the 1.1.0 release?

jrudolph: > My ultimate reasoning for making this change global is was to reduce surprises for people writing future tests (i.e. giving more QoL to developers), more specifically avoiding the "I just updated/wrote a test and now its failing due to a finely tuned timeout and now I have to figure out that it was because of that said timeout"

I agree with the goal of helping test writers but it's not good if tests only succeed or fail after a long timeout. Event the old 2 second timeout might have been somewhat of a test fixing bankruptcy issue where we didn't have the time to make sure all tests run through in a timely fashion.

mdedetrich: > I agree with the goal of helping test writers but it's not good if tests only succeed or fail after a long timeout. Event the old 2 second timeout might have been somewhat of a test fixing bankruptcy issue where we didn't have the time to make sure all tests run through in a timely fashion.

Should I go through and undo this global timeout increase and only apply it to the specific tests?

raboof: Perhaps a nice middle ground would be to increase the timeout in the configuration for the test classes that are affected? Not as neat as specifying it for an individual test, but more granular than applying it globally.

mdedetrich: Thats definitely true, however its usually only one specific test per test classes that requires an override (as an example with `WebSocketIntegrationSpec` its only `send back 100 elements and then terminate without error even when not ordinarily closed` that even needs the oeverride.

Basically its only tests which explicitly cancel some part of request/response as part of what is being tested and given how big test suites normally are in pekko-http it might not be enough of a difference to be useful.

PR #593, opened 2024-09-10
update extensions doc

pjfanning:
* some more extensions
* all the Akka HTTP equivalents are now listed in this doc (they all support Pekko now)


jrudolph (approved):
lgtm, great thanks for collecting these!

PR #597, opened 2024-09-12
Revert "Increase stream-cancellation-delay default to 1000 millis"

pjfanning:
Reverts apache/pekko-http#590

Still open for discussion. See https://lists.apache.org/thread/h59dphf5j20tzjwrm1z1c8olptnqrrhy

### DIFF
--- a/docs/src/main/paradox/release-notes/releases-1.1.md
+++ b/docs/src/main/paradox/release-notes/releases-1.1.md
@@ -12,7 +12,6 @@ All the changes in the @ref:[1.0.x releases](releases-1.0.md) up to and includin
 ### Changes
 * Changed names of HTTP status codes 413 and 422 ([PR87](https://github.com/apache/pekko-http/pull/87))
 * Parse entire HTTP chunk size ([PR528](https://github.com/apache/pekko-http/pull/528))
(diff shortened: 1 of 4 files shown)


raboof (approved):
It seems clear this work is in an 'intermediate' state: we would like the get back into a state where the "stream.testkit.all-stages-stopped-timeout = 20 s" is no longer needed, ideally either by improving the implementation (so the cancellation delay is no longer needed) or by improving the tests (so they no longer rely on the cancellation delay?), or otherwise by increasing this value only for specific tests where it is needed.

The question is what the risk of releasing with the "stream-cancellation-delay = 1000 millis" is. If this risk is low, it might be fine to release in this 'intermediate' state. If the risk is high, it would be safer to revert this change and live with the occasional 'SubscriptionWithCancelException$NoMoreElementsNeeded' until we find a more complete solution. Other than being ugly and possibly indicative of a deeper problem, these exceptions are essentially harmless, right?

What worries me most is the "-o 9" that apparently had to be added to H2SpecIntegrationSpec. This is a fairly realistic 'end-to-end' test, and needing to move this from 2 to 9 seconds seems like quite a big jump. 

Based on that, though I generally favour 'rolling forward', in this case I think it would probably be wise to indeed do the revert.

pjfanning:
I'm going to merge this and try to press on with a release. Users can set `stream-cancellation-delay` in their own configs. We continue to try to come up with a fuller solution for a future release.

PR #603, opened 2024-09-16
don't link to 'current' pekko docs

pjfanning:
link to the version specific docs


nvollmar (approved):
lgtm

raboof (approved):
LGTM.

At some point the "oldest Pekko version we support" will be older than the "Pekko version we recommend". In that case I think we should build against the "oldest Pekko version we support" but use the "Pekko version we recommend" in the documentation. That's not a problem we need to solve now, though.

PR #609, opened 2024-09-23
uptake sbt-apache-sonatype release via sbt-pekko-build

pjfanning:
uptakes https://github.com/mdedetrich/sbt-apache-sonatype/pull/40
via https://github.com/pjfanning/sbt-pekko-build/pull/15

we don't really need the typesafe resolver any more either


raboof (approved):
nice

PR #622, opened 2024-10-23
Fix partial match

NavidJalali:
I hit this case in a test and it threw. I am not sure if when `trailer.headers` is empty, it should return None or a parsed frame with empty headers but I figured the guard was there for a reason

### DIFF
--- a/http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/HttpMessageRendering.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/HttpMessageRendering.scala
@@ -99,7 +99,7 @@ private[http2] sealed abstract class MessageRendering[R <: HttpMessage] extends 
       r.attribute(AttributeKeys.trailer) match {
         case Some(trailer) if trailer.headers.nonEmpty =>
           OptionVal.Some(ParsedHeadersFrame(streamId, endStream = true, trailer.headers, None))
-        case None => OptionVal.None
+        case _ => OptionVal.None
       }
 
     Http2SubStream(r.entity, headersFrame, trailingHeadersFrame,
--- a/http2-tests/src/test/scala/org/apache/pekko/http/impl/engine/http2/HttpMessageRenderingSpec.scala
+++ b/http2-tests/src/test/scala/org/apache/pekko/http/impl/engine/http2/HttpMessageRenderingSpec.scala
@@ -25,7 +25,7 @@ import scala.collection.immutable.VectorBuilder
 import scala.util.Try
 import org.scalatest.matchers.should.Matchers
 import org.scalatest.wordspec.AnyWordSpec
-
+import scala.collection.immutable.Seq
 import scala.collection.immutable
 
 object MyCustomHeader extends ModeledCustomHeaderCompanion[MyCustomHeader] {
--- a/http2-tests/src/test/scala/org/apache/pekko/http/impl/engine/http2/HttpMessageRenderingSpec.scala
+++ b/http2-tests/src/test/scala/org/apache/pekko/http/impl/engine/http2/HttpMessageRenderingSpec.scala
@@ -26,6 +26,8 @@ import scala.util.Try
 import org.scalatest.matchers.should.Matchers
 import org.scalatest.wordspec.AnyWordSpec
 
+import scala.collection.immutable
+
 object MyCustomHeader extends ModeledCustomHeaderCompanion[MyCustomHeader] {
   override def name: String = "custom-header"
   override def parse(value: String): Try[MyCustomHeader] = ???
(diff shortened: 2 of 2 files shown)


pjfanning:
* seems sensible to change this
* is there an easy way to get to add a unit test?

NavidJalali:
@pjfanning I am honestly not sure what this code is really doing so I am a bit hesitant to assert things about this. I will have a look at the related tests to see if I can confidently test it

NavidJalali:
@pjfanning can you take another look when you get the time?

http2-tests/src/test/scala/org/apache/pekko/http/impl/engine/http2/HttpMessageRenderingSpec.scala
```diff
@@ -13,14 +13,14 @@
 
 package org.apache.pekko.http.impl.engine.http2
 
+import com.typesafe.config.ConfigFactory
 import java.time.format.DateTimeFormatter
 import org.apache.pekko
+import org.apache.pekko.http.scaladsl.settings.{ ClientConnectionSettings, ServerSettings }
```
mdedetrich, line 19: Can you fix this import so it matches the same style of imports currently in the file

mdedetrich (changes requested):
There is just one minor thing regarding imports but otherwise lgtm.

I also triggered the CI run

http2-tests/src/test/scala/org/apache/pekko/http/impl/engine/http2/HttpMessageRenderingSpec.scala
```diff
@@ -147,21 +145,39 @@ class HttpMessageRenderingSpec extends AnyWordSpec with Matchers {
...
+          HttpResponse().withAttributes(
+            Map(
+              AttributeKeys.trailer -> Trailer(),
+              Http2.streamId -> 0)))
+      }.isSuccess shouldBe true
+    }
+
   }
 
-  private def renderClientHeaders(headers: immutable.Seq[HttpHeader], builder: VectorBuilder[(String, String)],
```
pjfanning, line 152: please revert this change - it is probably here for Scala 2.12 support and there is no reason for this PR to be changing these methods anyway

NavidJalali: I think IntelliJ is deleting the Seq import, thinking its useless

pjfanning: it's not useless - add it back

http2-tests/src/test/scala/org/apache/pekko/http/impl/engine/http2/HttpMessageRenderingSpec.scala
```diff
@@ -13,14 +13,14 @@
 
 package org.apache.pekko.http.impl.engine.http2
 
+import com.typesafe.config.ConfigFactory
 import java.time.format.DateTimeFormatter
 import org.apache.pekko
+import pekko.http.scaladsl.settings.{ ClientConnectionSettings, ServerSettings }
```
pjfanning, line 19: this should be put with the other pekko.http imports

http2-tests/src/test/scala/org/apache/pekko/http/impl/engine/http2/HttpMessageRenderingSpec.scala
NavidJalali: yep sorry

http2-tests/src/test/scala/org/apache/pekko/http/impl/engine/http2/HttpMessageRenderingSpec.scala
```diff
@@ -13,14 +13,14 @@
...
 import pekko.event.NoLogging
 import pekko.http.impl.engine.rendering.DateHeaderRendering
 import pekko.http.scaladsl.model.headers._
-import pekko.http.scaladsl.model.{ ContentTypes, DateTime, HttpHeader, TransferEncodings }
-
-import scala.collection.immutable.Seq
+import pekko.http.scaladsl.model._
```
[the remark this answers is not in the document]
pjfanning: I've committed a change to the PR branch to try to order the imports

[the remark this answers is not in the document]
NavidJalali: I was a bit too slow 😄

mdedetrich (approved):
This is lgtm now, @pjfanning also check as well

pjfanning:
Thanks @NavidJalali - merging

pjfanning:
@mdedetrich this looks useful to backport to 1.0.x - wdyt?

http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/HttpMessageRendering.scala
```diff
@@ -99,7 +99,7 @@ private[http2] sealed abstract class MessageRendering[R <: HttpMessage] extends
       r.attribute(AttributeKeys.trailer) match {
         case Some(trailer) if trailer.headers.nonEmpty =>
           OptionVal.Some(ParsedHeadersFrame(streamId, endStream = true, trailer.headers, None))
-        case None => OptionVal.None
+        case _ => OptionVal.None
```
raboof, line 102: So the question here is if someone added a `AttributeKeys.trailer` but it contained no headers, should we produce a `ParsedHeadersFrame` here? Looking at `Http2SubStream`/`OutStreamImpl`, it does seem safe (it will produce an empty `DataFrame` with `endStream = true` if necessary). It seems unlikely that any downstream implementation would rely on receiving an empty frame of trailing headers.

All in all I think this is OK, though adding a `AttributeKeys.trailer` without any headers seems strange, the code that caused that might deserve a closer look.

pjfanning:
@raboof this relates to https://github.com/apache/pekko-grpc/pull/391

PR #630, opened 2024-11-15
Get string representation of javadsl Content and Media Type

mdedetrich:
Currently in the javadsl its not possible to get the string representation of the `Content-Type`/`Media-Type` because its only the scaladsl that implements the `value` function (via the `ValueRenderable` trait). This PR adds the `value` method to the core Java interfaces (the `render` method is anyways implemented in the subclasses hence why this PR is compiling without any additional changes)

As a workaround I currently have to do this (its in kotlin but should be understandable)

```kotlin
internal fun contentTypeToString(contentType: ContentType): String {
    return when (contentType) {
        is org.apache.pekko.http.scaladsl.model.ContentType ->
            return contentType.value()
        else -> throw RuntimeException("Unsupported content type: ${contentType::class.qualifiedName}")
    }
}
```

This works because it just so happens that every javdsl `ContentType` is also a scaladsl `ContentType`

### DIFF
--- a/http-core/src/main/mima-filters/1.2.x.backwards.excludes/content-media-type-value-string.backwards.excludes
+++ b/http-core/src/main/mima-filters/1.2.x.backwards.excludes/content-media-type-value-string.backwards.excludes
@@ -1,0 +1,2 @@
+ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.pekko.http.javadsl.model.ContentType.value")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.pekko.http.javadsl.model.MediaType.value")
--- a/http-core/src/main/scala/org/apache/pekko/http/javadsl/model/ContentType.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/javadsl/model/ContentType.scala
@@ -13,6 +13,8 @@
 
 package org.apache.pekko.http.javadsl.model
 
+import org.apache.pekko.annotation.DoNotInherit
+
 import java.util.Optional
 
 // Has to be defined in Scala even though it's JavaDSL because of:
@@ -45,6 +47,7 @@ object ContentType {
  *
  * See [[ContentTypes]] for convenience access to often used values.
  */
+@DoNotInherit
 trait ContentType {
 
   /**
@@ -61,4 +64,10 @@ trait ContentType {
    * Returns the charset if this ContentType is non-binary.
    */
   def getCharsetOption: Optional[HttpCharset]
+
+  /**
+   * Returns the string representation of this ContentType
+   * @since 1.2.0
+   */
+  def value: String
 }
--- a/docs/src/test/java/docs/http/javadsl/CustomMediaTypesExampleTest.java
+++ b/docs/src/test/java/docs/http/javadsl/CustomMediaTypesExampleTest.java
@@ -70,7 +70,7 @@ public class CustomMediaTypesExampleTest extends JUnitRouteTest {
         extractRequest(
             req ->
                 complete(
-                    req.entity().getContentType().toString()
+                    req.entity().getContentType().value()
                         + " = "
                         + req.entity().getContentType().getClass()));
 
--- a/http-core/src/main/scala/org/apache/pekko/http/javadsl/model/MediaType.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/javadsl/model/MediaType.scala
@@ -13,6 +13,8 @@
 
 package org.apache.pekko.http.javadsl.model
 
+import org.apache.pekko.annotation.DoNotInherit
+
 /**
  * Represents an Http media-type. A media-type consists of a main-type and a sub-type.
  *
@@ -65,6 +67,7 @@ object MediaType {
   }
 }
 
+@DoNotInherit
 trait MediaType {
 
   /**
@@ -110,4 +113,10 @@ trait MediaType {
    * Creates a media-range from this media-type with a given qValue.
    */
   def toRange(qValue: Float): MediaRange
+
+  /**
+   * Returns the string representation of this MediaType
+   * @since 1.2.0
+   */
+  def value: String
 }


http-core/src/main/scala/org/apache/pekko/http/javadsl/model/ContentType.scala
```diff
@@ -61,4 +61,9 @@ trait ContentType {
    * Returns the charset if this ContentType is non-binary.
    */
   def getCharsetOption: Optional[HttpCharset]
+
+  /**
+   * Returns the string representation of this ContentType
+   */
```
pjfanning, line 67: can you add `@since 1.1.1` to the 2 new public functions?

pjfanning: if we do go the 1.2.0 route, we will need to delay this PR merge until we get to create the 1.1.x branch and the CI jobs related to it

mdedetrich: Thats fine, its not urgent

pjfanning: * ok - let's go the 1.2.0 route
* I'll try to set up the CI builds over the next few days

pjfanning (approved):
lgtm - the 1.1.x branch exists now so this is safe to merge but I would appreciate a delay on the merge until we get #631 merged

mdedetrich:
> lgtm - the 1.1.x branch exists now so this is safe to merge but I would appreciate a delay on the merge until we get #631 merged

I would wait for 1.2.x, there is a workaround and since we are adding methods to an interface we should be bumping the minor version

pjfanning:
> > lgtm - the 1.1.x branch exists now so this is safe to merge but I would appreciate a delay on the merge until we get #631 merged
> 
> I would wait for 1.2.x, there is a workaround and since we are adding methods to an interface we should be bumping the minor version

* we use 'main' branch for the development version and 1.2.0 is becoming the development version
* future 1.1.x patch releases will be built using the '1.1.x' branch

http-core/src/main/scala/org/apache/pekko/http/javadsl/model/ContentType.scala
```diff
@@ -61,4 +61,10 @@ trait ContentType {
    * Returns the charset if this ContentType is non-binary.
    */
   def getCharsetOption: Optional[HttpCharset]
+
+  /**
+   * Returns the string representation of this ContentType
+   * @since 1.2.0
+   */
+  def value: String
 }
```
pjfanning, line 70: I think this should be getValue. More Java API like. Other existing methods seem to be getters.

pjfanning (changes requested):

http-core/src/main/scala/org/apache/pekko/http/javadsl/model/ContentType.scala
mdedetrich: Let me check because this isn't consistent, i.e. there is `getCharsetOption` but then there is also `mediaType` (which if it was consistent would be `getMediaType`)

pjfanning: reasonable point - it does seem like the existing code can't decide on getter style versus field name style.

http-core/src/main/scala/org/apache/pekko/http/javadsl/model/ContentType.scala
He-Pin: Maybe just override the `toString` method?

http-core/src/main/scala/org/apache/pekko/http/javadsl/model/ContentType.scala
He-Pin: In Spring-web, there is a `MimeTypeUtils#toString`, for me this method should be `getValue` or just override the `toString` method.

jxnu-liguobin: You're right, the type is always String, toString is more in line with Java.

pjfanning: toString would have the useful side effect that we would be ok to merge this to v1.1.x patch line

mdedetrich: I don't think putting `toString` on the interface is a good idea, its something that should concretely be implemented by its sub types.

mdedetrich: So fyi, changing it to `.getValue()` is more contrived than initially anticipated. Naively you would do something like

```scala
def getValue: String = value
```

However the `org.apache.pekko.http.javadsl.model.ContentType` has no visibility for the `value` method within the `ValueRenderable` trait (after all this is the exact issue we are dealing with). In order to make this work `org.apache.pekko.http.javadsl.model.ContentType` would have to extend `ValueRenderable` which would also expose the `value` method since its public i.e.
```
/**
 * INTERNAL API
 *
 * An entity that has a rendered value (like an HttpHeader)
 */
@InternalApi
private[http] trait ValueRenderable extends ToStringRenderable {
  def value: String = toString
}
```
which means the trait would both have `value` and `getValue` methods, far from ideal (note that even though the trait is marked as `private[http]`, this is only visible to the Scala compiler, Java has no concept of it and hence it only sees it as a standard interface with a public method).

I think this discussion is revealing of what appears to be an overlooked design misdecision, which is that we should be having 2 `ValueRenderable` traits, one for Java api and one for Scala api and only the respective implementations extending those traits.

I can go ahead and do this change but its obviously going to be more work. Thankfully `ValueRenderable` trait is marked as internal/private which means I only need to take care that a public `def value: String = toString` exists for the Scala api.

Wdyt?

pjfanning: If leaving this as `def value = ???` is easier, I am ok with that. It seems like the Java API is already inconsistent stylistically.

mdedetrich: > If leaving this as `def value = ???` is easier, I am ok with that. It seems like the Java API is already inconsistent stylistically.

I personally would like to avoid that, I think that leaving unimplemented methods is taking it a bridge too far. @raboof what are your thoughts on this? I don't have an issue with doing this properly, this PR is slated for 1.2.0 anyways.

raboof: I think it would make sense to keep this as `def value: String`. I don't think we need the `= ???` as all implementations of this trait have an implementation for this value, right?

Indeed 'getValue' would be more Java-esque, but since we already have `mediaType` this seems fine.

http-core/src/main/mima-filters/1.2.x.backwards.excludes/content-media-type-value-string.backwards.excludes
```diff
@@ -0,0 +1,2 @@
+ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.pekko.http.javadsl.model.ContentType.value")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.pekko.http.javadsl.model.MediaType.value")
```
raboof, line 2: This is only justified if there are no user types extending from these classes (or their subclasses), i.e. if these traits are marked `@DoNotInherit`. The traits are not marked `@DoNotInherit`, so naively you'd say this change is not allowed. However, I think you can make the argument that the fact that `@DoNotInherit` is missing from this trait is an oversight, and it was never intended to be user-extended. It would be good to add the `@DoNotInherit` now, then?

mdedetrich: Do you mean putting `@DoNotInherit` on `ContentType`/`MediaType` traits themselves?

mdedetrich:
> CustomMediaTypesExampleTest

Done

http-core/src/main/mima-filters/1.2.x.backwards.excludes/content-media-type-value-string.backwards.excludes
mdedetrich: No thats fair, I thought you were talking about `ValueRenderable` but that wouldn't make sense its already marked as `@InternalApi`.

We can remove and also fix the Java namings (i.e. `value` should be `getValue`) as well as removing `@DoNotInherit` in 2.x.x

http-core/src/main/scala/org/apache/pekko/http/javadsl/model/ContentType.scala
mdedetrich: Okay fair, I will mark this as resolved (if someone still disagrees then unresolve conversation).

mdedetrich:
@pjfanning Can you re-review this PR given @raboof 's comments?

raboof (approved):

pjfanning:
> @pjfanning Can you re-review this PR given @raboof 's comments?

The changes look ok. We still need to block the merge until we have the 1.1.x branch and related CI builds.

raboof:
> We still need to block the merge until we have the 1.1.x branch and related CI builds.

I think I missed why we'd want to delay this merge - this PR as it now stands seems like a regular backwards-compatible change (assuming nobody inherited from `ContentType` or `MediaType`, which seems like a reasonable assumption)?

pjfanning:
> > We still need to block the merge until we have the 1.1.x branch and related CI builds.
> 
> I think I missed why we'd want to delay this merge - this PR as it now stands seems like a regular backwards-compatible change (assuming nobody inherited from `ContentType` or `MediaType`, which seems like a reasonable assumption)?

The PR currently has `@since 1.2.0`. If these change to `@since 1.1.1` then we don't need to start preparing for 1.2.0 snapshots.

raboof:
> > > We still need to block the merge until we have the 1.1.x branch and related CI builds.
> > 
> > I think I missed why we'd want to delay this merge - this PR as it now stands seems like a regular backwards-compatible change (assuming nobody inherited from `ContentType` or `MediaType`, which seems like a reasonable assumption)?
> 
> The PR currently has `@since 1.2.0`. If these change to `@since 1.1.1` then we don't need to start preparing for 1.2.0 snapshots.

Aah, gotcha. This begs the question whether those `@since` annotations are really worth having. I can easily imagine situations where it isn't clear yet whether the next release will be a minor, patch or major one. Should/have we discuss(ed) this on the mailinglist?

mdedetrich:
> I think I missed why we'd want to delay this merge - this PR as it now stands seems like a regular backwards-compatible change (assuming nobody inherited from `ContentType` or `MediaType`, which seems like a reasonable assumption)?

So actually, the reason for putting it in 1.2.0 is that while its not a breaking compatible change it is a forwards breaking change (hence the MiMa changes) which is allowed by semver in minor version bumps (its only backwards compatibility thats disallowed, this has to be done in a major patch update).

pjfanning (approved):
approved for main branch merge - now that main branch is repurposed for v1.2 dev

mdedetrich:
Thanks, I will go ahead and merge it.

PR #631, opened 2024-11-15
protect 1.1.x branch

pjfanning:


.asf.yaml
```diff
@@ -46,6 +46,18 @@ github:
         dismiss_stale_reviews: false
         require_code_owner_reviews: false
         required_approving_review_count: 1
+    1.1.x:
+      required_status_checks:
+        # strict means "Require branches to be up to date before merging".
+        strict: false
```
raboof, line 52: isn't that the default?

raboof (approved):
makes sense to me. link validator error is #632

PR #645, opened 2025-01-02
add source headers to CI files

pjfanning:
header depends on if the yml was added first for Pekko or if it is based on one that we inherited from Akka code


raboof (approved):

.github/workflows/dependency-graph.yml
```diff
@@ -12,6 +29,6 @@ jobs:
       - uses: actions/checkout@v4
       - name: Install sbt
         uses: sbt/setup-sbt@v1      
-      - uses: scalacenter/sbt-dependency-submission@v2
+      - uses: scalacenter/sbt-dependency-submission@v3
```
raboof, line 32: sneaky :)

PR #651, opened 2025-01-15
Fix some compiler warnings

NavidJalali:
Replaces some deprecated stuff

### DIFF
--- a/http-core/src/main/scala/org/apache/pekko/http/impl/engine/server/HttpServerBluePrint.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/impl/engine/server/HttpServerBluePrint.scala
@@ -135,7 +135,7 @@ private[http] object HttpServerBluePrint {
         }
 
         // optimization: this callback is used to handle entity substream cancellation to avoid allocating a dedicated handler
-        override def onDownstreamFinish(): Unit = {
+        override def onDownstreamFinish(cause: Throwable): Unit = {
           if (entitySource ne null) {
             // application layer has cancelled or only partially consumed response entity:
             // connection will be closed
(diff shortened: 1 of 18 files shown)


http-core/src/main/scala/org/apache/pekko/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -336,13 +336,14 @@ private[http] object HttpServerBluePrint {
...
-          override def onUpstreamFailure(ex: Throwable) = fail(responseOut, ex)
+          override def onUpstreamFinish(): Unit = complete(responseOut)
+          override def onUpstreamFailure(ex: Throwable): Unit = fail(responseOut, ex)
         })
       setHandler(responseOut,
         new OutHandler {
-          def onPull(): Unit = pull(responseIn)
-          override def onDownstreamFinish() = cancel(responseIn)
+          def onPull():
+          Unit = pull(responseIn)
```
pjfanning, line 345: Can you revert this change? It causes scalafmt check to fail.

NavidJalali: oops, I think I must have accidentally pressed enter here 😄

NavidJalali: Should be good now

He-Pin: run with `scala-cli fmt`

pjfanning (approved):
lgtm

pjfanning:
merged - thanks

PR #675, opened 2025-05-09
missing asf header

pjfanning:
need Apache source header


pjfanning:
@raboof are these files in the right directory - should they be in 1.1.x.backwards.excludes instead of 1.2.x.backwards.excludes?
I get the impression that we sometimes say the dir is called after the release previous version and sometimes call it after the expected next release version.

He-Pin (approved):

PR #680, opened 2025-05-16
issues with stackoverflow links in CI runs

pjfanning:
failing on basically every CI run now

https://github.com/apache/pekko-http/actions/workflows/link-validator.yml


raboof (approved):
LGTM (but also we should move the link validator out of CI and into a cronjob)

Issue #682, opened 2025-05-28
Support media type for .mjs


nvollmar:
JavaScript module files using .mjs should have content-type `application/javascript`

See:
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/MIME_types/Common_types
* https://stackoverflow.com/questions/57492546/what-is-the-difference-between-js-and-mjs-files

PR #683, opened 2025-05-28
Add mjs extension for javascript media type

nvollmar:
As described in #682


nvollmar:
added issue to the commit

PR #684, opened 2025-05-28
Add trailer header

NavidJalali:
According to https://datatracker.ietf.org/doc/html/rfc7230#section-4.4
Introduced to pekko-grpc but moving it here
https://github.com/apache/pekko-grpc/pull/456#discussion_r2109260332

### DIFF
--- a/http-core/src/main/scala/org/apache/pekko/http/scaladsl/model/headers/headers.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/scaladsl/model/headers/headers.scala
@@ -1226,6 +1226,7 @@ final case class `X-Real-Ip`(address: RemoteAddress) extends jm.headers.XRealIp
   protected def companion = `X-Real-Ip`
 }
 
+@ApiMayChange
 object Trailer extends ModeledCompanion[Trailer] {
   private implicit val trailersRenderer: Renderer[immutable.Iterable[String]] =
     Renderer.defaultSeqRenderer[String]


http-core/src/main/java/org/apache/pekko/http/javadsl/model/headers/Trailer.java
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.pekko.http.javadsl.model.headers;
+
+import org.apache.pekko.http.impl.util.Util;
+
+public abstract class Trailer extends org.apache.pekko.http.scaladsl.model.HttpHeader {
```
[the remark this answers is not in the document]
pjfanning: for a new class wih methods, it is sufficient to just annotate the class

NavidJalali:
The naming clash between this and the other class named Trailer is somewhat annoying, and enclosing the class name in backticks actually doesnt make any difference, they would still collide, I am open to suggestions for a different name tho

NavidJalali:
> > > The naming clash between this and the other class named Trailer is somewhat annoying, and enclosing the class name in backticks actually doesnt make any difference, they would still collide, I am open to suggestions for a different name tho
> > 
> > 
> > Is there anyway that
> > https://github.com/apache/pekko-http/blob/887c2175a92c703cb3349450228aa101912b987c/http-core/src/main/scala/org/apache/pekko/http/scaladsl/model/Trailer.scala#L22
> > 
> > could be extended to support what you are doing here?
> 
> Tbh while I'm sure we could make it work I think it would be extra confusing to use the same class for 'the header that announces the trailing headers' and 'the actual trailing headers'.
> 
> > If there needs to be 2 separate classes, maybe we could call the new one `TrailerHeader`. It's a bit untidy but its easier to having 2 classes in the same jar with the same class name even if the package differs.
> 
> There is precedent for having header classes that have a different name from the name of the header itself: for example, the `TE` class represents the `Transfer-Encoding` header, as `TransferEncoding` was already taken for the encodings 'themselves'. I wonder if we could make it even more descriptive, perhaps `TrailerAnnouncement`? But `TrailerHeader` works for me as well.

I have some suggestions, maybe you can pick one:

How about any of `TrailerHint`, `TrailerFields`, `TrailerDeclaration`? `TrailerHeader` is okay too, it just sounds a bit off to me

NavidJalali:
I've hit a little snag, modeled companion class always pre-initializes the name of the header as the name of the class! 
Even `TE` is rendered as `TE: ???`. If I overwrite it it will cause initialization issues and NPEs
image
I guess I could solve that by turning that val into a def but its a change in behavior and I am not sure of the performance impacts. Any suggestions?

pjfanning:
Calling the class `Trailer` is ok if this is the easiest way to go. We can always alias the class in the import statements if we need to differentiate it from the existing Trailer class.
```
import org.apache.pekko.http.javadsl.model.headers.{Trailer => TrailerHint}
```

NavidJalali:
> Calling the class `Trailer` is ok if this is the easiest way to go. We can always alias the class in the import statements if we need to differentiate it from the existing Trailer class.
> 
> ```
> import org.apache.pekko.http.javadsl.model.headers.{Trailer => TrailerHint}
> ```

Yeah that would work in scala, I haven't really written a ton of Java but I think it doesnt have import renames 😞

http-core/src/main/scala/org/apache/pekko/http/scaladsl/model/headers/headers.scala
```diff
@@ -1226,3 +1225,70 @@ final case class `X-Real-Ip`(address: RemoteAddress) extends jm.headers.XRealIp
...
+  def apply(values: immutable.Seq[String]): Trailer = {
+    val clean = values.map(_.trim).filter(_.nonEmpty)
+    val (forbidden, allowed) = clean.partition(name => isForbidden(name.toRootLowerCase))
+    if (clean.isEmpty)
+      throw new IllegalArgumentException(s"Trailer values must not be empty: No valid header names specified")
+    else if (forbidden.nonEmpty) {
+      val forbiddenInput = forbidden.mkString("[", ", ", "]")
+      throw new IllegalArgumentException(
+        s"Trailer values must not contain forbidden header names: Trailer contained $forbiddenInput")
+    } else new Trailer(allowed)
```
raboof, line 1242: I'm not entirely sure this validation belongs here, but OK I guess.

Could we use an exception type from the `ExceptionWithErrorInfo` hierarchy, that allows splitting the error message 'itself' from the possibly-unsafe 'details'?

NavidJalali: Can you think of a better place to put it?  I only put it here because I didn't know of a better place

NavidJalali:
Since changing the name to TrailerHeader is an invasive change, I'll keep it as is for now. If its not okay then I will have to modify the constructor of ModelCompanion

http-core/src/main/java/org/apache/pekko/http/javadsl/model/headers/Trailer.java
```diff
@@ -0,0 +1,44 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.pekko.http.javadsl.model.headers;
+
+import org.apache.pekko.http.impl.util.Util;
+
+/**
+ * The `Trailer` header is used before a message body to indicate which fields will be present in
+ * the trailers when using chunked transfer encoding. See <a
```
raboof, line 24: perhaps mention it's also used in HTTP/2?

http-core/src/test/scala/org/apache/pekko/http/impl/model/parser/HttpHeaderSpec.scala
```diff
@@ -19,7 +19,7 @@ import pekko.http.impl.model.parser.HeaderParser.Settings
 import org.scalatest.matchers.{ MatchResult, Matcher }
 import pekko.http.impl.util._
 import pekko.http.scaladsl.model.{ HttpHeader, _ }
-import headers._
+import headers.{ Trailer => `Trailer`, _ }
```
raboof, line 22: Hmm, this is creative use of `. Not sure if I like it, but I guess it can't hurt.

raboof (approved):
Some comments for consideration, but LGTM

NavidJalali:
> lgtm - will merge and tag the commit to start the 1.3 dev line

Thanks, I will clean up Pekko Http once 1.3 is released

pjfanning:
@NavidJalali we probably don't want to make pekko-grpc dependent on pekko-http 1.3 (which is what I presume is what you suggesting).
A little bit of code duplication is no harm if it avoids adding tight dependencies between modules.
I raised https://github.com/apache/pekko-grpc/issues/470

PR #690, opened 2025-06-14
avoid unsafe in StringTools

pjfanning:
* targeted at pekko-http 1.3
* in my microbenchmark testing, the String.getBytes method has improved in java 21 and 24
* avoiding Unsafe in recent Java versions reduces our warnings at runtime because sun.misc.Unsafe is now noisy (lots of logging)


He-Pin (approved):
lgtm

PR #714, opened 2025-08-05
compiler warning issues

pjfanning:
* reduces the issues that we silence with `-Wcat`
* fixes a couple of issues
* adds explicit nowarns and TODOs in places where we need to think about making changes
* there are some other issues left behind that I hope to revisit


He-Pin (approved):
lgtm

PR #715, opened 2025-08-06
don't use `@Deprecated` when we also have `@deprecated`

pjfanning:
relates to #714 

We get a lot of compiler warnings due to this.

Example
```
[warn] /home/code/pekko-http/http-core/src/main/scala/org/apache/pekko/http/javadsl/ConnectionContext.scala:115:4: Prefer the Scala annotation over Java's `@Deprecated` to provide a message and version: @deprecated("message", since = "MyLib 1.0")
```


raboof:
I think we added `@Deprecated` to Java APIs because Java users/tools wouldn't understand the Scala-specific `@deprecated` annotation. Did scala at some point start auto-generating the Java-style annotations also for the Scala annotation? A quick test suggests it doesn't, but I might be looking at it wrong.

If this didn't change, perhaps we should suppress this warning for all `javadsl` packages at least, that might go a long way towards removing this friction?

PR #718, opened 2025-08-17
stop using deprecated method Flow.lazyInitAsync

pjfanning:
pekko 2.0.0 removes a lot of deprecated methods and it would be nice to tidy up pekko-http to avoid such methods


He-Pin (approved):
lgtm

PR #740, opened 2025-09-03
remove withFuzzing call in tests

pjfanning:
The fuzzing setting is set in the config - so this is not needed.
The API is deprecated and is being removed.


He-Pin (approved):
lgtm

PR #744, opened 2025-09-05
SSE Client Oversized Messages Handling

rrwright:
# Problem

The current SSE (Server-Sent Events) client implementation fails its stream every time a message arrives over the size limits. When a Pekko SSE client connects to a Pekko SSE server who serves a message larger than the size limit, the client fails the stream, tries to reconnect and resume at the same (oversized) message and fails again. The client falls into an infinite connection retry loop. The oversized message error is swallowed by the sse-connector (another PR for the `pekko-connectors` library will follow), so the user never knows what is happening. The observed behavior appears as if the stream is still active and just gets stuck at the message that preceeded the oversized message.

# Solution
 This PR introduces two new configuration values: `pekko.http.sse.oversized-line-handling` and `pekko.http.sse.oversized-event-handling` which give the user additional options for how the SSE client should handle receiving a message sized over the configured limit. Four options are available for each:
1. `fail-stream`: Terminates the stream when an oversized line/event is encountered (this is the previous behavior and the new default behavior) 
2. `log-and-skip`: Log oversized line/event as a warning, skips it and continues processing the stream
3. `truncate`: Truncates the oversized line/event to fit within limits and continue (event truncation is by-line, for safety reasons)
4. `dead-letter`: Sends the oversized line/event to dead letter queue (truncated for safety, if need be) and continues

# Tests

```
OversizedSseStrategySpec:
OversizedSseStrategy
- should parse valid string values correctly (12 milliseconds)
- should throw IllegalArgumentException for invalid string values (2 milliseconds)
- should convert strategy objects back to strings correctly (0 milliseconds)
- should handle case-sensitive strings (1 millisecond)
- should handle empty and null strings (0 milliseconds)
EventStreamParserOversizedSpec:
An EventStreamParser with oversized message handling
- should parse normal SSE messages correctly with all strategies (52 milliseconds)
- should fail the stream when using FailStream strategy with oversized SSE line (2 milliseconds)
- should skip oversized SSE lines and continue processing with LogAndSkip strategy (1 millisecond)
- should truncate oversized SSE lines and continue processing with Truncate strategy (2 milliseconds)
- should send oversized SSE lines to dead letters and continue processing with DeadLetter strategy (1 millisecond)
- should handle multiple oversized lines in complex SSE events with LogAndSkip strategy (1 millisecond)
- should handle multiline data with some oversized lines using Truncate strategy (1 millisecond)
- should handle streaming SSE data with oversized content across chunks (4 milliseconds)
- should handle event field oversizing with different strategies (8 milliseconds)
- should work with unlimited line sizes when maxLineSize is 0 (2 milliseconds)
- should fail the stream when using FailStream strategy with oversized SSE event (max-event-size) (2 milliseconds)
- should skip oversized SSE events and continue processing with LogAndSkip strategy (max-event-size) (1 millisecond)
- should truncate oversized SSE events and continue processing with Truncate strategy (max-event-size) (1 millisecond)
- should send oversized SSE events to dead letters and continue processing with DeadLetter strategy (max-event-size) (1 millisecond)
- should handle events that exceed max-event-size during construction with DeadLetter strategy (0 milliseconds)
- should work with unlimited event sizes when maxEventSize is 0 (1 millisecond)
- should handle mixed line and event size violations with consistent strategies (0 milliseconds)
- should skip remaining lines in event after hitting size limit during construction (1 millisecond)
OversizedSseStrategySimpleTest:
- testEnumValues
- testFromScala
EventStreamUnmarshallingSimpleSpec:
EventStreamUnmarshalling with oversized message handling
- should fail the stream with FailStream strategy (47 milliseconds)
- should skip oversized content with LogAndSkip strategy (2 milliseconds)
- should truncate oversized content with Truncate strategy (1 millisecond)
- should send oversized content to dead letters with DeadLetter strategy (1 millisecond)
EventStreamUnmarshallingOversizedTest:
- testOversizedStrategyEnum
- testOversizedLineStrategyStringCompatibility
```

# Debatable Topics

Calling out specifically some changes which could very reasonably be otherwise:

## ~~Config values~~

**_The following config value changes were pulled out to simplify the PR. I'll submit an issue with a recommendation and summary of my comparison research for what better default values could be. Then a separate PR can consider that alone._**

~~Size limits of oversized messages are determined by two config values (one deprecated), `max-event-size` and `max-live-size` which have their own complex relationships and very small default values. This PR increases the default values to the smallest reasonable size for modern uses (estimated by looking at standards applied in browsers. I found none for SSE, but Websocket limits seem to be within the same spirit and some data on browser limits was available.~~

~~The previous limits (4k and 8k) are remarkably small for modern use cases. As an example: ethereum block chain events (a common enough SSE use case) can be hundreds of kilobytes. While it is good practice to keep Pekko messages small, these limits affect what the client is allowed to consume from the server over HTTP. They aren't primarily about Pekko messages. Larger defaults seem like a good idea, but it's not core to the primary change proposed in this PR.~~

## Unlimited message size

This PR adds the option to disable line and event size limits entirely by setting those config settings to zero.

### DIFF
--- a/http/src/main/scala/org/apache/pekko/http/scaladsl/unmarshalling/sse/LineParser.scala
+++ b/http/src/main/scala/org/apache/pekko/http/scaladsl/unmarshalling/sse/LineParser.scala
@@ -60,18 +60,18 @@ private final class LineParser(maxLineSize: Int,
       setHandlers(in, out, this)
 
       override def onPush() = {
-        def handleLineOversized(lineLength: Int, line: String): Option[String] = {
+        def handleLineOversized(lineByteSize: Int, line: String): Option[String] = {
           oversizedStrategy match {
             case OversizedSseStrategy.FailStream =>
               failStage(new IllegalStateException(
-                s"SSE line size: $lineLength exceeds max-line-size: $maxLineSize. " +
+                s"SSE line size: $lineByteSize exceeds max-line-size: $maxLineSize. " +
                 s"Configure pekko.http.sse.max-line-size or use oversized-message-handling setting."))
               None
             case OversizedSseStrategy.LogAndSkip =>
-              log.warning("Skipping oversized SSE message: {} bytes > {} max-line-size", lineLength, maxLineSize)
+              log.warning("Skipping oversized SSE message: {} bytes > {} max-line-size", lineByteSize, maxLineSize)
               None
             case OversizedSseStrategy.Truncate =>
-              log.info("Truncating oversized SSE message: {} bytes > {} max-line-size", lineLength, maxLineSize)
+              log.info("Truncating oversized SSE message: {} bytes > {} max-line-size", lineByteSize, maxLineSize)
               Some(line.take(maxLineSize))
             case OversizedSseStrategy.DeadLetter =>
               materializer.system.deadLetters ! OversizedSseLine(line)
@@ -92,10 +92,10 @@ private final class LineParser(maxLineSize: Int,
             bs(at) match {
               case CR if at < bs.length - 1 && bs(at + 1) == LF =>
                 // Lookahead for LF after CR
-                val lineLength = at - from
+                val lineByteSize = at - from
                 val line = bs.slice(from, at).utf8String
-                val processedLine = if (maxLineSize > 0 && lineLength > maxLineSize) {
-                  handleLineOversized(lineLength, line)
+                val processedLine = if (maxLineSize > 0 && lineByteSize > maxLineSize) {
+                  handleLineOversized(lineByteSize, line)
                 } else {
                   Some(line)
                 }
@@ -103,10 +103,10 @@ private final class LineParser(maxLineSize: Int,
                 parseLines(bs, at + 2, at + 2, newParsedLines, lastCharWasCr = false)
               case CR =>
                 // if is a CR but we don't know the next character, slice it but flag that the last character was a CR so if the next happens to be a LF we just ignore
-                val lineLength = at - from
+                val lineByteSize = at - from
                 val line = bs.slice(from, at).utf8String
-                val processedLine = if (maxLineSize > 0 && lineLength > maxLineSize) {
-                  handleLineOversized(lineLength, line)
+                val processedLine = if (maxLineSize > 0 && lineByteSize > maxLineSize) {
+                  handleLineOversized(lineByteSize, line)
                 } else {
                   Some(line)
                 }
@@ -117,10 +117,10 @@ private final class LineParser(maxLineSize: Int,
                 parseLines(bs, at + 1, at + 1, parsedLines, lastCharWasCr = false)
               case LF =>
                 // a LF that wasn't preceded by a CR means we found a new slice
-                val lineLength = at - from
+                val lineByteSize = at - from
                 val line = bs.slice(from, at).utf8String
-                val processedLine = if (maxLineSize > 0 && lineLength > maxLineSize) {
-                  handleLineOversized(lineLength, line)
+                val processedLine = if (maxLineSize > 0 && lineByteSize > maxLineSize) {
+                  handleLineOversized(lineByteSize, line)
                 } else {
                   Some(line)
                 }


He-Pin:
That's ture, we at work configured it to 5MB

docs/src/main/paradox/common/sse-support.md
```diff
@@ -56,6 +56,49 @@ Scala
...
+
+```hocon
+pekko.http.sse {
+  # How to handle messages that exceed max-line-size limit
+  # Options:
+  #   "fail-stream" - Fail the stream with a clear error message (default)
+  #   "log-and-skip" - Log a warning and skip the oversized message
+  #   "truncate" - Log a warning and truncate the message to max-line-size
+  #   "dead-letter" - Log a warning, send oversized message to dead letters
+  oversized-message-handling = "fail-stream"
```
He-Pin: Seems dead later is not needed, because we already has a log and skip

http-tests/src/test/scala/org/apache/pekko/http/scaladsl/unmarshalling/sse/EventStreamParserOversizedSpec.scala
```diff
@@ -0,0 +1,271 @@
...
+        .runWith(Sink.seq)
+        .map { result =>
+          result should have size 3
+          result(0) shouldBe ServerSentEvent("before")
+          result(1) shouldBe ServerSentEvent(veryLongData)
+          result(2) shouldBe ServerSentEvent("after")
+        }
+    }
+  }
+}
```
He-Pin: New line

http/src/main/scala/org/apache/pekko/http/scaladsl/settings/OversizedSseStrategy.scala
```diff
@@ -0,0 +1,39 @@
...
+import org.apache.pekko.annotation.InternalApi
+
+sealed trait OversizedSseStrategy
+
+object OversizedSseStrategy {
+  case object FailStream extends OversizedSseStrategy
+  case object LogAndSkip extends OversizedSseStrategy
+  case object Truncate extends OversizedSseStrategy
+  case object DeadLetter extends OversizedSseStrategy
+
```
He-Pin, line 32: Java api is needed

He-Pin (commented):
Thanks , but seems the Java api is missing, it is needed , if we want to build a client like Spring Ai

docs/src/main/paradox/common/sse-support.md
rrwright: My thought was that, while dead letters often just get logged, a pekko user can subscribe/intercept the events that go to dead letters and then handle them if some special logic is needed. So the dead letter queue is potentially more than just logging but also a way to add last-ditch error handling. But maybe it should just send to dead letters without explicitly logging; let logging happen naturally for unhandled dead letters. …changes pushed; I removed logging from the dead letter option.

http/src/main/scala/org/apache/pekko/http/scaladsl/settings/OversizedSseStrategy.scala
rrwright: 👍 I just pushed changes that should now include the java api.

docs/src/main/paradox/common/sse-support.md
He-Pin: if the dead leatter can not be consumed fast , may be an oom？

rrwright: It's just the standard pekko dead letter queue. So it's a message sent inside the same JVM. I'm pretty sure the dead letter queue is event-driven (I think it's a system actor, but I don't remember if it's actually backed by an actor or not, but it does have an ActorRef). After it's handled, it'll be garbage collected when there's no held reference like any other message. The timing to consume the message wouldn't require any special handling. It would be like any other pekko message.

docs/src/main/paradox/common/sse-support.md
```diff
@@ -56,6 +56,49 @@ Scala
...
+}
+```
+
+For applications that need to handle very large messages (like blockchain data or detailed JSON payloads), consider 
+setting `max-line-size = 0` to disable limits entirely, or use one of the non-failing handling modes.
+
 Notice that if you are looking for a resilient way to permanently subscribe to an event stream,
-Apache Pekko Connectors provides the [EventSource](https://pekko.apache.org/docs/pekko-connectors/current/sse.html)
-connector which reconnects automatically with the id of the last seen event.
+Apache Pekko Connectors provides the [EventSource](https://pekko.apache.org/docs/pekko-connectors/current/sse.html)connector which reconnects automatically with the id of the
```
pjfanning: can you add back a space or new line before the word 'connector' after the link?

http-tests/src/test/java/org/apache/pekko/http/javadsl/settings/OversizedSseStrategySimpleTest.java
```diff
@@ -0,0 +1,40 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * license agreements; and to You under the Apache License, version 2.0:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * This file is part of the Apache Pekko project, which was derived from Akka.
```
pjfanning: This is the wrong header for new files - that are not derived from Akka files.
The right header is https://github.com/apache/pekko-http/blob/main/project/AddMetaInfLicenseFiles.scala#L1-L16

pjfanning: can you check all the new files and their source headers?

http/src/main/resources/reference.conf
```diff
@@ -52,11 +52,20 @@ pekko.http {
 
   # server-sent events
   sse {
-    # The maximum size for parsing server-sent events.
-    max-event-size = 8192
+    # The maximum size for parsing received server-sent events.
+    # This value must be larger than `max-line-size`. Set to 0 to disable limit entirely (unlimited).
+    max-event-size = 115713
```
pjfanning: why change the values? - that will affect other users

http-tests/src/test/scala/org/apache/pekko/http/scaladsl/unmarshalling/sse/LineParserSpec.scala
```diff
@@ -1,66 +0,0 @@
...
-package unmarshalling
-package sse
-
-import org.apache.pekko
-import pekko.stream.scaladsl.{ Sink, Source }
-import pekko.util.ByteString
-import org.scalatest.matchers.should.Matchers
-import org.scalatest.wordspec.AsyncWordSpec
-
-final class LineParserSpec extends AsyncWordSpec with Matchers with BaseUnmarshallingSpec {
```
pjfanning, line 25: why delete this?

http/src/main/scala/org/apache/pekko/http/javadsl/settings/ServerSentEventSettings.scala
```diff
@@ -42,9 +45,47 @@ abstract class ServerSentEventSettings private[pekko] () { self: ServerSentEvent
    */
   def emitEmptyEvents: Boolean
 
-  // ---
+  /**
+   * How to handle messages that exceed max-line-size limit.
+   * Valid options: "fail-stream", "log-and-skip", "truncate", "dead-letter"
+   */
```
pjfanning, line 52: new methods in src/main should have `@since 1.3.0` (next non patch release)

docs/src/main/paradox/common/sse-support.md
rrwright: ✅

http-tests/src/test/java/org/apache/pekko/http/javadsl/settings/OversizedSseStrategySimpleTest.java
rrwright: 👍 I believe this is updated now for all the relevant files.

http/src/main/resources/reference.conf
rrwright: As noted in the PR description and in [other reviewer comments](https://github.com/apache/pekko-http/pull/744#issuecomment-3260298481), 8k (and 4k) is extremely small for any modern use case. I believe the argument for keeping these values small is because Pekko messages should be small (and instead convey a link to fetch larger payloads through an outside channel if needed), but these limits are not actually limiting Pekko messages, they are limiting what is read out of the HTTP connection. Small limits combined with the existing behavior of the SSE connector (swallows stream failure messages caused by this limit being triggered) cause silent failures and lost data. 

Also noted in the PR description, this is another new arbitrary value and very reasonably argued against, and not core the changes proposed in this PR.

http-tests/src/test/scala/org/apache/pekko/http/scaladsl/unmarshalling/sse/LineParserSpec.scala
rrwright: git oversight. thank you for catching/questioning. fixed.

http/src/main/scala/org/apache/pekko/http/javadsl/settings/ServerSentEventSettings.scala
rrwright: Added. Looking through code for precedents, it looks like it may also apply to enums, sealed traits, or other things. I'm not entirely sure where it's required, but please advise if some were missed or others are unnecessary. (in latest commit)

He-Pin:
need sbt javafmtAll and scalafmt

pjfanning:
Scala 2.12 build has
```
[error] /home/runner/work/pekko-http/pekko-http/http/src/main/scala/org/apache/pekko/http/javadsl/settings/ServerSentEventSettings.scala:65:56: value fromString is not a member of object org.apache.pekko.http.javadsl.settings.OversizedSseStrategy
[error]     self.copy(oversizedStrategy = OversizedSseStrategy.fromString(newValue))
[error]                                                        ^
Error: value fromString is not a member of object org.apache.pekko.http.javadsl.settings.OversizedSseStrategy
[warn] /home/runner/work/pekko-http/pekko-http/http/src/main/scala/org/apache/pekko/http/javadsl/settings/ServerSentEventSettings.scala:20:37: imported `OversizedSseStrategy` is permanently hidden by definition of Java enum OversizedSseStrategy in package settings
[warn] import pekko.http.scaladsl.settings.OversizedSseStrategy
[warn]                                     ^
Warning: imported `OversizedSseStrategy` is permanently hidden by definition of Java enum OversizedSseStrategy in package settings
Warning: Unused import
```
Can you fix the import and try to work out what causes the compile issue - which doesn't seem to happen in Scala 2.13 build?

http/src/main/resources/reference.conf
mdedetrich: Is it possible to look at current state of art for other http clients when it comes to default max message size, maybe we can just copy it from there?

rrwright: for sure. I had used browsers as the prototypical comparison. The proposed size, `115713`, comes from Chrome's limit on web-socket message size. Size limits for SSE lines and messages apparently doesn't exist for browsers (at least that I could find). I'll try to collect some other related examples from other clients and include them here to make the case.

rrwright: I changed the default values back to their original values. I'll put the argument and evidence for changes in an issue and we can consider whether to update default values in that issue and a separate PR.

rrwright:
Pending CI runs/success, I believe everything else for the PR is addressed and it should be mergeable.

http/src/main/scala/org/apache/pekko/http/scaladsl/unmarshalling/sse/LineParser.scala
```diff
@@ -43,44 +55,86 @@ private final class LineParser(maxLineSize: Int) extends GraphStage[FlowShape[By
...
-            lastCharWasCr: Boolean = false): (ByteString, Vector[String], Boolean) =
+            lastCharWasCr: Boolean): (ByteString, Vector[String], Boolean) =
           if (at >= bs.length)
             (bs.drop(from), parsedLines, lastCharWasCr)
           else
             bs(at) match {
               case CR if at < bs.length - 1 && bs(at + 1) == LF =>
                 // Lookahead for LF after CR
-                parseLines(bs, at + 2, at + 2, parsedLines :+ bs.slice(from, at).utf8String)
+                val lineLength = at - from
```
He-Pin: I think the `lineSize` should be changed to `lineByteSize` in the setting, because some Char will use two bytes but just count as one char.

Here, the code seems like count the bytes

http/src/main/java/org/apache/pekko/http/javadsl/settings/OversizedSseStrategy.java
```diff
@@ -0,0 +1,79 @@
...
+ * Strategy for handling oversized SSE messages that exceed the configured max-line-size.
+ *
+ * @since 1.3.0
+ */
+public enum OversizedSseStrategy {
+  /**
+   * Fail the stream with an IllegalStateException when an oversized message is encountered. This is
+   * the default behavior to maintain backward compatibility.
+   */
+  FailStream,
```
raboof, line 30: I think the convention for Java enums is still `FAIL_STREAM`, right? or is that outdated? For example compared to https://github.com/apache/pekko/blob/main/stream/src/main/java/org/apache/pekko/stream/javadsl/AsPublisher.java ?

http/src/main/scala/org/apache/pekko/http/scaladsl/unmarshalling/sse/ServerSentEventParser.scala
```diff
@@ -110,18 +120,57 @@ private final class ServerSentEventParser(
...
+          if (line.isEmpty) shouldSkipUntilEventEnd = false // Stop skipping when end of event (empty line) is reached
+          pull(in) // Already reported oversized event (below). Drop and continue to next line.
+        } else if (maxEventSize > 0 && builder.size + line.length > maxEventSize) { // Next line exceeds the size limit
+          shouldSkipUntilEventEnd = true
+          oversizedStrategy match {
+            case OversizedSseStrategy.FailStream =>
+              builder.appendData(line)
+              val event = builder.build()
+              failStage(new IllegalStateException(
+                s"Oversized SSE Event ${event.id.fold("") { id => s"at ID: $id " }}" +
```
raboof, line 140: Can the `id` ever be user-controlled here? Perhaps to be safe it might make sense to use a `ExceptionWithErrorInfo` here where the id is in the 'detail' rather than the 'summary'.

raboof (commented):
This looks like an awesome contribution: useful goal, reasonable implementation, great tests. I have a few minor comments to doublecheck, but this looks very close to being merge-able to me.

http/src/main/java/org/apache/pekko/http/javadsl/settings/OversizedSseStrategy.java
rrwright: My thought on style choices like that is to prioritize consistency with the other code in the repo. Looks like there is some variation throughout all Pekko libraries, though consistent usage of UPPERCASE in pekko streams (like the `AsPublisher` example you linked to). But in this pekko-http repo, a quick check looks like CamelCase is used everywhere in the production code (e.g.: `Coder`, `SameSite`), and the UPPERCASE notation is used in a test (e.g. `MyRole`). That makes me think it fits best in the local scope of this repository to keep it as CamelCase. But no strong opinion here.

http/src/main/scala/org/apache/pekko/http/scaladsl/unmarshalling/sse/ServerSentEventParser.scala
rrwright: Hmm, I'm not sure what you mean by "user-controlled". The exception would occur in the SSE client. The `id` value comes from the SSE `id:` field which is data sent from the server. The meaning of the ID field from the [SSE spec](https://html.spec.whatwg.org/multipage/server-sent-events.html) is that it "sets the last event ID string," which means it identifies and distinguishes each semantically-whole SSE event (which will be spread across multiple lines), and is used for tracking events and resuming connections (e.g. browsers automatically send `Last-Event-ID` header on reconnect). The spec only limits it to a UTF-8 string of non-null characters. No length constraint (hence the need for a line limit which is the topic of this PR). In general, it's usually a counter that behaves like the numeric offset in Kafka.

I'm not really familiar with the intended use of `ExceptionWithErrorInfo` and its docstring didn't help me much. Looking through the code, I see 13 total uses. Those uses look like they're all for processing HTTP entities and contexts on the server-side. I don't really have an opinion on whether that's the right exception choice to use here because I don't understand the intention of that exception class.

The choice of failing the stage with an `IllegalStateException` is actually just preserving what the code used to do previously and keeps the default behavior (and exception type) unless the user opts into one of the new choices. I'm happy to change it though if there's clear guidance.

http/src/main/scala/org/apache/pekko/http/scaladsl/unmarshalling/sse/LineParser.scala
rrwright: great point! I updated the name.

rrwright:
Related tiny PR to avoid swallowing the error message in the SSE connector is here: https://github.com/apache/pekko-connectors/pull/1205

He-Pin:
SSE is needed to working with Ai:)

http/src/main/java/org/apache/pekko/http/javadsl/settings/OversizedSseStrategy.java
```diff
@@ -0,0 +1,79 @@
...
+ */
+
+package org.apache.pekko.http.javadsl.settings;
+
+/**
+ * Strategy for handling oversized SSE messages that exceed the configured max-line-size.
+ *
+ * @since 1.3.0
+ */
+public enum OversizedSseStrategy {
```
He-Pin, line 25: How about adding the getInstance method in the Scala OversizedSseStrategy, this will keep the same as pekko, otherwise, there will be two OversizedSseStrategys shows in IDE

http/src/main/scala/org/apache/pekko/http/scaladsl/unmarshalling/sse/ServerSentEventParser.scala
raboof: The scenario would be an application that connects to 'untrusted' SSE servers and exposes the result somewhere, including showing an error based on the exception. As a 'defense in depth' mechanism, it would be good to avoid including attacker-controllable input in that error, but it would be good for diagnostics if it's possible to have it in the logs. For that reason we have exceptions in the `ExceptionWithErrorInfo` hierarchy, which allow splitting the message into `summary` and 'details', where only the 'summary' is exposed by default, and 'details' can contain additional (possibly attacker-controlled) details. This is briefly documented at https://github.com/apache/pekko-http/blob/main/http-core/src/main/scala/org/apache/pekko/http/scaladsl/model/ErrorInfo.scala#L25 and https://pekko.apache.org/docs/pekko-http/current/routing-dsl/exception-handling.html#including-sensitive-data-in-exceptions

> The choice of failing the stage with an `IllegalStateException` is actually just preserving what the code used to do previously and keeps the default behavior (and exception type) unless the user opts into one of the new choices. I'm happy to change it though if there's clear guidance.

That makes sense (though the original exception didn't include the id...)

I guess perhaps it's somewhat far-fetched in this case, I'm ok with a 'traditional' exception.

http/src/main/java/org/apache/pekko/http/javadsl/settings/OversizedSseStrategy.java
raboof: > My thought on style choices like that is to prioritize consistency with the other code in the repo.

I agree!

I don't see clear guidance in https://github.com/apache/pekko/blob/main/CONTRIBUTING.md or https://pekko.apache.org/docs/pekko/current/typed/style-guide.html . Javafmt doesn't seem to have a preference either. I agree it's already inconsistent, and while I'd prefer a consistent style across all Pekko projects, following the 'local convention' here now seems sensible.

http/src/main/java/org/apache/pekko/http/javadsl/settings/OversizedSseStrategy.java
raboof: do we have any precedence for that elsewhere? I think it's OK to have two, people will recognize the javadsl/scaladsl package names.

He-Pin: See Pekko-core `org.apache.pekko.stream.OverflowStrategy`

raboof: right, so that's not in a `javadsl`/`scaladsl` package but moves it to the common parent package. https://github.com/apache/pekko/blob/main/CONTRIBUTING.md#java-apis-in-pekko does not provide a clear preference, and I don't think I have a clear preference either.

PR #749, opened 2025-09-09
remove more deprecated code

pjfanning:
* I expect more PRs - this is just some of the deprecated code that can be removed in 2.0.0
* Some of the remaining deprecated code is used
* The ParserSettings deprecations might need to be removed because the shared trait SettingsCompanion is used by many other classes and the other implementations haven't deprecated the methods that are deprecated in ParserSettings


He-Pin:
I agree,  we can do removal in several steps

He-Pin (approved):
Lgtm

He-Pin:
If we drop 2.12.x in Pekko 2.0.0, then we will drop 2.12.x in every sub project right？

pjfanning:
> If we drop 2.12.x in Pekko 2.0.0, then we will drop 2.12.x in every sub project right？

Yes. It is implied that dropping Scala 2.12 support for the core libs means that every lib that we maintain will need to drop Scala 2.12 support for their next major releases. We will still support patching older releases that support Scala 2.12 if there are problems found in those versions.

PR #751, opened 2025-09-09
remove LowerPriorityRouteResultImplicits

pjfanning:
* LowerPriorityRouteResultImplicits only has 1 method and it is deprecated
* so seems best to remove the trait altogether - I think it is pretty safe to assume that it is not a trait that users would extend thinking it was for public use
* updated a test that relied on the deprecated implicit and made it use the recommended change (of making the ActorSystem implicit)


He-Pin (approved):
lgtm

PR #753, opened 2025-09-09
remove deprecated bind methods

pjfanning:
* for 2.0.0 release


He-Pin (approved):
lgtm，but need fmt

pjfanning:
> @pjfanning Can we use an autofix app to handle the fmt thing ? eg [com-lihaoyi/mill#5817](https://github.com/com-lihaoyi/mill/pull/5817)

Would be nice to automate this but I'm not sure where that bot comes from and we would probably need to work with ASF Infra team to allow it to work in an ASF GitHub project where we have Security restrictions.

PR #754, opened 2025-09-09
avoid deprecated methods in docs example

pjfanning:
* a backport of doc example fix from #753 
* it is not ideal for our doc examples to rely on deprecated methods


He-Pin (approved):
lgtm

PR #775, opened 2025-09-14
don't use pekko.dispatch.ExecutionContexts

pjfanning:
with Scala 2.12 support removed, we can use Scala methods directly


He-Pin (approved):
Lgtm

PR #776, opened 2025-09-14
don't use pekko.util.JavaDurationConverters

pjfanning:
now that we don't support Scala 2.12, we can use Scala calls directly

### DIFF
--- a/http-core/src/main/scala/org/apache/pekko/http/javadsl/ServerBinding.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/javadsl/ServerBinding.scala
@@ -20,10 +20,10 @@ import org.apache.pekko
 import pekko.Done
 import pekko.actor.ClassicActorSystemProvider
 import pekko.annotation.DoNotInherit
-import pekko.util.JavaDurationConverters._
 
 import scala.concurrent.ExecutionContext
 import scala.concurrent.duration.FiniteDuration
+import scala.jdk.DurationConverters._
 import scala.jdk.FutureConverters._
 
 /**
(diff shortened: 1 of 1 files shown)


http-core/src/main/scala/org/apache/pekko/http/javadsl/settings/ConnectionPoolSettings.scala
```diff
@@ -37,7 +36,7 @@ abstract class ConnectionPoolSettings private[pekko] () { self: ConnectionPoolSe
   def getMaxRetries: Int = maxRetries
   def getMaxOpenRequests: Int = maxOpenRequests
   def getPipeliningLimit: Int = pipeliningLimit
-  def getMaxConnectionLifetime: JDuration = maxConnectionLifetime.asJava
+  def getMaxConnectionLifetime: JDuration = JDuration.ofNanos(maxConnectionLifetime.toNanos)
```
He-Pin: can be toJava

pjfanning: scala.jdk.DurationConverters only works with FiniteDuration - maxConnectionLifetime is not FiniteDuration

He-Pin: if maxConnectionLifetime.isFinate maxConnectionLifetime.toJava else JDuration.Zero

pjfanning: Basically we have a bug in pekko.util.JavaDurationConverters.
It allows you to compile:
```
import pekko.util.JavaDurationConverters._
scala.concurrent.duration.Duration.Inf.asJava
```

But this will fail at runtime because pekko.util.JavaDurationConverters calls toNanos on the scala duration and this fails with
```
java.lang.IllegalArgumentException: toNanos not allowed on infinite Durations
	at scala.concurrent.duration.Duration$Infinite.fail(Duration.scala:219)
```

I'm going to park this PR for a while - and concentrate on fixing the bug.
I think pekko.util.JavaDurationConverters should return the max possible Java Duration because Java Duration does not support infinite.

PR #778, opened 2025-09-15
use Scala FunctionConverter and remove some unused implicits

pjfanning:
* some of this is facilitated by removing Scala 2.12 support
* some of the imports may not even have been required before


He-Pin (approved):
lgtm

http/src/main/scala/org/apache/pekko/http/scaladsl/common/StrictForm.scala
```diff
@@ -19,7 +19,6 @@ import scala.concurrent.{ ExecutionContext, Future }
 import scala.concurrent.duration._
 import org.apache.pekko
 import pekko.stream.Materializer
-import pekko.http.ccompat._
```
He-Pin, line 22: why no error reporting, seems like we need some `unused` lint in this project.

PR #780, opened 2025-09-16
Java duration in javadsl

pjfanning:
* targeting v1.3.0 and will forward fit to main - main is in flux due to all the 2.0.0 changes
* I looked at the javadsl packages in this repo and found a number of classes where we use scala.duration.Duration instead of java.time.Duration in the Java DSL
* I added overloaded methods that accept java.time.Duration where feasible.
* There are some getXXX methods that return scala.duration.Duration - changing this is a breaking change that we can't do in 1.x - I think we should live with and fix it in 2.0.0 where we can make breaking changes
    * Adding extra methods but with names like getXXXAsJava is too messy and so far, we don't have too many complaints from users


http/src/main/scala/org/apache/pekko/http/javadsl/server/directives/BasicDirectives.scala
```diff
@@ -391,9 +440,43 @@ abstract class BasicDirectives {
...
+   * than allowed by the `pekko.http.parsing.max-to-strict-bytes` configuration setting.
+   *
+   * Extracts the [[pekko.http.javadsl.server.RequestContext]] itself with the strict HTTP entity,
+   * or fails the route if unable to drain the entire request body within the timeout.
+   *
+   * @param timeout The directive is failed if the stream isn't completed after the given timeout.
+   * @deprecated As of 1.3.0, use the overload taking a `java.time.Duration` instead.
+   */
+  @Deprecated
+  @deprecated("use the overload taking a `java.time.Duration` instead.", "1.3.0")
```
He-Pin: the overload method?

nvollmar (approved):
lgtm

He-Pin (approved):
lgtm

Issue #783, opened 2025-09-17
Change Java DSL methods that return Scala Durations to return Java Durations


pjfanning:
follow up to #780

Issue #782, opened 2025-09-17
switch to Scala DurationConverters


pjfanning:
* https://github.com/apache/pekko-http/pull/776 is in progress but there are more changes that need to be done

PR #784, opened 2025-09-17
Java duration in javadsl (#780)

pjfanning:
* Forward fit of #780 
* there is some follow up work
    * #782
    * #783

### DIFF
--- a/http-core/src/main/java/org/apache/pekko/http/javadsl/TimeoutAccess.java
+++ b/http-core/src/main/java/org/apache/pekko/http/javadsl/TimeoutAccess.java
@@ -17,7 +17,6 @@ import org.apache.pekko.annotation.DoNotInherit;
 import org.apache.pekko.http.javadsl.model.HttpRequest;
 import org.apache.pekko.http.javadsl.model.HttpResponse;
 import org.apache.pekko.japi.function.Function;
-import scala.concurrent.duration.Duration;
 
 /**
  * Enables programmatic access to the server-side request timeout logic.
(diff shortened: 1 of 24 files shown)


http-core/src/main/scala/org/apache/pekko/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -404,9 +407,18 @@ private[http] object HttpServerBluePrint {
     import pekko.http.impl.util.JavaMapping.Implicits._
 
     /** JAVA API * */
-    def update(timeout: Duration, handler: Function[model.HttpRequest, model.HttpResponse]): Unit =
+    override def updateTimeout(timeout: java.time.Duration): Unit = {
+      val stimeout = if (timeout eq null) null else timeout.asScala
```
He-Pin: Not sure why timeout can be `null`.

pjfanning: the existing code passes around nulls internally unfortunately

He-Pin: Okay, I actually think that should be an NPE.

pjfanning: I adjusted the code. The Scala DSL does support nulls but the Java DSL doesn't need to

http-core/src/main/java/org/apache/pekko/http/javadsl/TimeoutAccess.java
```diff
@@ -34,17 +33,31 @@ public interface TimeoutAccess {
    * <p>Due to the inherent raciness it is not guaranteed that the returned timeout was applied
    * before the previously set timeout has expired!
    */
-  Duration getTimeout();
+  scala.concurrent.duration.Duration getTimeout();
```
He-Pin, line 36: this should be deprecated or at least add some doc about that it will be Java duration

pjfanning: I intend to break this API in a later commit. This is just a forward fit of #780. For traceability, I want to add those changes in one PR and do another PR with the 2.0.0 breaking changes.
This API method isn't changing in this PR, I just fully referenced scala.concurrent.duration.Duration to make it more obvious that that was class that was being returned.
In my follow up PR, I will change the return type to java.time.Duration.

http/src/main/scala/org/apache/pekko/http/javadsl/server/directives/BasicDirectives.scala
```diff
@@ -43,13 +41,16 @@ import pekko.http.scaladsl.server.{ Directives => D }
...
+import java.time.{ Duration => JDuration }
 import java.util.{ List => JList }
 import java.util.concurrent.CompletionStage
+import java.util.function.{ Function => JFunction }
 import java.util.function.Predicate
+import java.util.function.Supplier
 
 import scala.concurrent.{ ExecutionContext, ExecutionContextExecutor }
 import scala.concurrent.duration.FiniteDuration
+import scala.jdk.DurationConverters._
```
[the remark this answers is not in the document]
pjfanning: in 2.0.0, we are switching to scala.jdk converters

PR #790, opened 2025-09-18
copy over sortImports from main pekko repo

pjfanning:
Support `sbt sortImports`

### DIFF
--- a/.scalafix.conf
+++ b/.scalafix.conf
@@ -1,0 +1,55 @@
+// .scalafix.conf
+rules = [
+  RemoveUnused
+  SortImports
+  ExplicitResultTypes
+  "github:ohze/scalafix-rules/FinalObject"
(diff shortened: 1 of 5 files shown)


mdedetrich:
As stated in https://github.com/apache/pekko/discussions/2228#discussioncomment-14446340 this is going to break all paradox documentation imports. There is a reason why we haven't used `sortImports`/`scalafix.

pjfanning:
> As stated in [apache/pekko#2228 (comment)](https://github.com/apache/pekko/discussions/2228#discussioncomment-14446340) this is going to break all paradox documentation imports and it also breaks groupings with `import org.apache.pekko`. There is a reason why we haven't used `sortImports`/`scalafix`.

I've disabled it on the docs module.

He-Pin:
@pjfanning I think this one can be closed.

PR #791, opened 2025-09-18
Resort imports

pjfanning:
* Demo of #790 
* I changed some of the imports so that the sortImports can handle them (usually by more fully qualifying the imports)
* CustomMediaTypesSpec is used in our docs and none of the features that I tried that were meant to disable scalafix stopped it being processed - I think I managed to get it to a state where it still works for paradox but where scalafix has processed it

### DIFF
--- a/http-tests/src/test/scala/org/apache/pekko/http/scaladsl/CustomMediaTypesSpec.scala
+++ b/http-tests/src/test/scala/org/apache/pekko/http/scaladsl/CustomMediaTypesSpec.scala
@@ -20,6 +20,7 @@ import pekko.http.scaladsl.client.RequestBuilding
 import pekko.http.scaladsl.model._
 import pekko.http.scaladsl.model.MediaType.WithFixedCharset
 import pekko.http.scaladsl.server.Directives
+import pekko.http.scaladsl.settings.{ ParserSettings, ServerSettings }
 import pekko.testkit._
 import pekko.util.ByteString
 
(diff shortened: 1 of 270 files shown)


pjfanning:
This sort is repeatable (run `sbt sortImports`) and doesn't break anything. It runs quickly.

If someone comes up with another sorting solution, we can add that later.

http-tests/src/test/scala/org/apache/pekko/http/scaladsl/CustomMediaTypesSpec.scala
```diff
@@ -32,22 +34,21 @@ class CustomMediaTypesSpec extends PekkoSpec with ScalaFutures
...
+      set.add(MediaType.customBinary("application", "x-Pekko-TEST", MediaType.NotCompressible))
 
       set.contains(MediaType.parse("application/msword").right.get) should ===(true)
       set.contains(MediaType.parse("application/MsWord").right.get) should ===(true)
       set.contains(MediaType.parse("application/vnd.ms-POWERPOINT").right.get) should ===(true)
       set.contains(MediaType.parse("application/VnD.MS-eXceL").right.get) should ===(true)
-      set.contains(MediaType.parse("application/x-akka-test").right.get) should ===(true)
-      set.contains(MediaType.parse("application/x-Akka-TEST").right.get) should ===(true)
+      set.contains(MediaType.parse("application/x-pekko-test").right.get) should ===(true)
+      set.contains(MediaType.parse("application/x-Pekko-TEST").right.get) should ===(true)
```
He-Pin, line 44: nice

He-Pin (approved):
lgtm

He-Pin:
@pjfanning Would you think it valuable to have this in pekko-grpc too.

pjfanning:
> @pjfanning Would you think it valuable to have this in pekko-grpc too.

* not as much code in pekko-grpc and generally, the build is more complicated (mix of sbt, mvn and gradle)
* may not be worth the effort

PR #799, opened 2025-09-25
more performant boundary check

pjfanning:
* tries to avoid passing the bytestring twice
* need to debug some broken tests

### DIFF
--- a/http-core/src/main/scala/org/apache/pekko/http/impl/engine/parsing/BodyPartParser.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/impl/engine/parsing/BodyPartParser.scala
@@ -16,18 +16,17 @@ package org.apache.pekko.http.impl.engine.parsing
 import org.apache.pekko
 import pekko.NotUsed
 import pekko.annotation.InternalApi
-
-import scala.annotation.tailrec
 import pekko.event.LoggingAdapter
-import org.parboiled2.CharPredicate
+import pekko.http.impl.util._
+import pekko.http.scaladsl.model._
+import pekko.http.scaladsl.model.headers._
+import pekko.stream.{ Attributes, FlowShape, Inlet, Outlet }
 import pekko.stream.scaladsl.Source
 import pekko.stream.stage._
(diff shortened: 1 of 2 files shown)


http-core/src/main/scala/org/apache/pekko/http/impl/engine/parsing/BodyPartParser.scala
```diff
@@ -359,11 +359,30 @@ private[http] object BodyPartParser {
...
-      else this
+      checkForBoundary(byteString) match {
+        case CR => DefinedEndOfLineConfiguration("\r\n", boundary)
+        case LF => DefinedEndOfLineConfiguration("\n", boundary)
+        case _  => this
+      }
+    }
+
+    private val CR = '\r'.toByte
+    private val LF = '\n'.toByte
```
He-Pin: Nice， There are many `LR`, `LF` thing, can we put them in a single file Netty does.
io.netty.handler.codec.http.HttpConstants

And use `final val LF = 10` make it a compiler constant.

http-core/src/main/scala/org/apache/pekko/http/impl/engine/parsing/BodyPartParser.scala
```diff
@@ -355,15 +355,23 @@ private[http] object BodyPartParser {
 
   case class UndefinedEndOfLineConfiguration(boundary: String) extends EndOfLineConfiguration {
     override def eol: String = "\r\n"
-    private val CR: Byte = 13
-    private val LF: Byte = 10
+    private final val CR: Byte = 13
+    private final val LF: Byte = 10
```
He-Pin: Can we have a dedicated new class `HttpConstants` to put these bytes

http-core/src/main/scala/org/apache/pekko/http/impl/engine/parsing/BodyPartParser.scala
pjfanning: I created HttpConstants and will follow up with another PR that makes more use if it

Issue #798, opened 2025-09-25
Http2 header parser can be made more performant


pjfanning:
https://github.com/apache/pekko-http/blob/af7ed86d6f00c67cf2d70a8a70e1ad3e46d0dcbd/http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/RequestParsing.scala#L194

* currently converts the strings back to bytes to parse
* adds a hacky 'x' to the content to get the current parser to accept the byte stream
* may need a dedicated new parser method that works at char or String level

PR #800, opened 2025-09-26
avoid ByteString when parsing HTTP/2 headers

pjfanning:
* experiment for #798 
* WIP

### DIFF
--- a/http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/RequestParsing.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/RequestParsing.scala
@@ -24,7 +24,6 @@ import pekko.http.scaladsl.model._
 import pekko.http.scaladsl.model.headers.`Tls-Session-Info`
 import pekko.http.scaladsl.settings.ServerSettings
 import pekko.stream.Attributes
-import pekko.util.ByteString
 import pekko.util.OptionVal
 
 import scala.annotation.tailrec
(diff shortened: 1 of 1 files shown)


http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/RequestParsing.scala
```diff
@@ -191,13 +190,18 @@ private[http2] object RequestParsing {
...
-    val concHeaderLine = name + ": " + value + "\r\nx"
-    httpHeaderParser.parseHeaderLine(ByteString(concHeaderLine))()
-    httpHeaderParser.resultHeader
+    import HttpHeader.ParsingResult
+    if (name.startsWith(":")) {
+      val concHeaderLine = s"$name: $value\r\nx"
+      httpHeaderParser.parseHeaderLine(pekko.util.ByteString(concHeaderLine))()
+      httpHeaderParser.resultHeader
+    } else {
+      HttpHeader.parse(name, value, httpHeaderParser.settings) match {
```
jrudolph, line 202: In general that's really cleaner. I don't expect a performance hit, there's a bit of a difference of how the parsers are looked up between `HttpHeader.parse` and `httpHeaderParser.parseHeaderLine`. `HttpHeader.parse` uses binary search to find the right parser and `httpHeaderParser.parseHeaderLine` uses kind of a lut / radix lookup once the parser is primed.

pjfanning: Thanks @jrudolph.
I had to use the `startsWith(":")` check because `HttpHeader.parse` doesn't support pseudo headers and 1 test includes a `:grpc-status` pseudo header.

jrudolph: Yeah, that's annoying

jrudolph: But wait, does it even parse it into anything useful?

pjfanning: I'll debug it later. There is an existing test that relies on the header parser not failing. Whether it needs the parser to actually process the pseudo header at all is something that I haven't yet tested.

jrudolph: Maybe it's a typo in the test? Is `:grpc-status` a thing, i.e. is it allowed for gRPC (outside the HTTP/2 spec) to extend the set of `:` pseudo headers?

jrudolph: pekko-grpc also doesn't use the colon form so maybe it's just a typo and we could just fix the test.

http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/RequestParsing.scala
```diff
@@ -191,13 +190,21 @@ private[http2] object RequestParsing {
...
   private[http2] def parseHeaderPair(httpHeaderParser: HttpHeaderParser, name: String, value: String): HttpHeader = {
-    // FIXME: later modify by adding HttpHeaderParser.parseHttp2Header that would use (name, value) pair directly
-    //        or use a separate, simpler, parser for Http2
-    // The odd-looking 'x' below is a by-product of how current parser and HTTP/1.1 work.
-    // Without '\r\n\x' (x being any additional byte) parsing will fail. See HttpHeaderParserSpec for examples.
-    val concHeaderLine = name + ": " + value + "\r\nx"
-    httpHeaderParser.parseHeaderLine(ByteString(concHeaderLine))()
-    httpHeaderParser.resultHeader
+    import HttpHeader.ParsingResult
+    if (name.startsWith(":")) {
```
He-Pin, line 194: Maybe just a look ahead char is better

PR #809, opened 2025-10-01
shared character toLowerCase function

pjfanning:
current code uses 3 different ways to lowercase a char
* parboiled2 CharUtils method
* an internal method on HttpHeaderParser
* uses Java built-in Character.toLowerCase

I did a jmh benchmark in my pekko-bench project
https://github.com/pjfanning/pekko-bench/blob/main/src/main/scala/example/LowerCaseBench.scala

And this showed that function in HttpHeaderParser was better than or matched the perf of the others on all Java versions.
Parboiled2 was always slower. The Java built-in was as good for Java 21 and 25 but slowest approach in java 17 and below.

This change switches all code to use the pekko-http internal method.

### DIFF
--- a/http-core/src/main/scala/org/apache/pekko/http/impl/model/parser/CommonActions.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/impl/model/parser/CommonActions.scala
@@ -80,7 +80,7 @@ private[parser] trait CommonActions {
         val char2 = str2.charAt(at)
 
         (char1 | char2) < 0x80 &&
-        Character.toLowerCase(char1) == Character.toLowerCase(char2) &&
+        CharUtils.toLowerCase(char1) == CharUtils.toLowerCase(char2) &&
         stringEquals(at + 1, length)
       } else true
 
--- a/http-core/src/main/scala/org/apache/pekko/http/impl/model/parser/UriParser.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/impl/model/parser/UriParser.scala
@@ -16,13 +16,13 @@ package org.apache.pekko.http.impl.model.parser
 import java.nio.charset.Charset
 
 import org.apache.pekko
-import org.parboiled2._
-import pekko.http.impl.util.{ enhanceString_, StringRendering }
+import pekko.annotation.InternalApi
+import pekko.http.impl.util.{ enhanceString_, CharUtils => CU, StringRendering }
 import pekko.http.scaladsl.model.{ Uri, UriRendering }
 import pekko.http.scaladsl.model.headers.HttpOrigin
-import Parser.DeliveryScheme.Either
 import Uri._
-import pekko.annotation.InternalApi
+import org.parboiled2._
+import org.parboiled2.Parser.DeliveryScheme.Either
 
 /**
  * INTERNAL API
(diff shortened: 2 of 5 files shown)


http-core/src/main/scala/org/apache/pekko/http/impl/model/parser/CommonActions.scala
```diff
@@ -80,7 +80,7 @@ private[parser] trait CommonActions {
         val char2 = str2.charAt(at)
 
         (char1 | char2) < 0x80 &&
-        Character.toLowerCase(char1) == Character.toLowerCase(char2) &&
+        CharUtils.toLowerCase(char1) == CharUtils.toLowerCase(char2) &&
```
raboof, line 83: 👍 clearly safe

http-core/src/main/scala/org/apache/pekko/http/impl/model/parser/UriParser.scala
```diff
@@ -365,7 +365,7 @@ private[http] final class UriParser(
 
   ///////////// helpers /////////////
 
-  private def appendLowered(): Rule0 = rule { run(sb.append(CharUtils.toLowerCase(lastChar))) }
+  private def appendLowered(): Rule0 = rule { run(sb.append(CU.toLowerCase(lastChar))) }
```
raboof, line 368: can we assume `lastChar` is 7-bit ascii here? or are we OK with letting other characters through?

raboof (commented):
I like this change, but it'd be good to double-check that this doesn't allow sneaking in harmful characters that were sanitized out before. How did parboiled behave for non-7-bit-ascii characters?

http-core/src/main/scala/org/apache/pekko/http/impl/model/parser/CommonActions.scala
pjfanning: Actually, in hindsight, this might be the riskiest change. CharUtils.toLowerCase only lower cases the 26 standard letters of the latin alphabet (the ones in 7 bit ASCII table) while java.lang.Character.toLowerCase seems to support other letters that appear in Unicode table. Let me re-review this change.

http-core/src/main/scala/org/apache/pekko/http/impl/model/parser/UriParser.scala
pjfanning: Parboiled CharUtils only deals with the 26 standard letters of the latin alphabet (the ones in 7 bit ASCII table) - same as the pekko-http internal toLowerCase.

http-core/src/main/scala/org/apache/pekko/http/impl/model/parser/CommonActions.scala
pjfanning: actually with  (char1 | char2) < 0x80 in the check, I think we can assume that it doesn't matter that Character.toLowerCase can handle non-ASCII chars.

pjfanning:
@raboof I've looked again and I think these changes are safe. wdyt?

raboof:
I'm not sure why d55eed032fbcc4b7e0eca15dd5e872d625c827b7 makes sense? It seems 'correct' but looks like it'd be slower if anything?

pjfanning:
> I'm not sure why [d55eed0](https://github.com/apache/pekko-http/commit/d55eed032fbcc4b7e0eca15dd5e872d625c827b7) makes sense? It seems 'correct' but looks like it'd be slower if anything?

The main part d55eed0 was to change to calling byteChar function. This should get inlined by Hotspot compiler. It is used in lots of places already and this usage seems strange not to use it. I can remove this change though and come back to it. Maybe the correct thing to do is actually make the inlining of byteChar explicit (by declaring `inline def byteChar` and `@inline def byteChar` which means slitting the code into Scala 2 and 3 versions of the package.scala).

raboof (approved):
ok!

PR #812, opened 2025-10-02
compiler warning about varargs

pjfanning:
issues like
```
[warn] /home/runner/work/pekko-http/pekko-http/http-core/src/main/scala/org/apache/pekko/http/impl/model/UriJavaAccessor.scala:40:99: Passing an explicit array value to a Scala varargs method is deprecated (since 2.13.0) and will result in a defensive copy; Use the more efficient non-copying ArraySeq.unsafeWrapArray or an explicit toIndexedSeq call
```

### DIFF
--- a/http-core/src/main/scala/org/apache/pekko/http/impl/model/UriJavaAccessor.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/impl/model/UriJavaAccessor.scala
@@ -40,7 +40,8 @@ private[http] object UriJavaAccessor {
   def hostApply(string: String, charset: Charset): Host = Uri.Host(string, charset = charset)
   def emptyHost: Uri.Host = Uri.Host.Empty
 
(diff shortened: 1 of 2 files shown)


http-core/src/main/scala/org/apache/pekko/http/impl/model/UriJavaAccessor.scala
```diff
@@ -37,7 +40,7 @@ private[http] object UriJavaAccessor {
   def hostApply(string: String, charset: Charset): Host = Uri.Host(string, charset = charset)
   def emptyHost: Uri.Host = Uri.Host.Empty
 
-  def queryApply(params: Array[pekko.japi.Pair[String, String]]): Uri.Query = Uri.Query(params.map(_.toScala): _*)
+  def queryApply(params: Array[pekko.japi.Pair[String, String]]): Uri.Query = Uri.Query(ArraySeq.unsafeWrapArray(params.map(_.toScala)): _*)
```
He-Pin: should we remove the _*

pjfanning: the code doesn't compile with it

PR #814, opened 2025-10-02
close input streams

pjfanning:
#813 failed but I thought it would be tidier to close InputStreams even though the data is in memory


http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/hpack/HeaderDecompression.scala
```diff
@@ -102,7 +103,7 @@ private[http2] final class HeaderDecompression(masterHeaderParser: HttpHeaderPar
             // this is signalled by the decoder when it failed, we want to react to this by rendering a GOAWAY frame
             fail(eventsOut,
               new Http2Compliance.Http2ProtocolException(ErrorCode.COMPRESSION_ERROR, "Decompression failed."))
-        }
+        } finally bis.close()
```
He-Pin: with a block？

pjfanning: changed - the existing code usually formats 1 line `finally` statements like I did in my first draft but I'm neutral about which is best

PR #815, opened 2025-10-02
fix some compiler warnings related to use of Either

pjfanning:
most issues are like
```
[warn] /home/code/pekko-http/http2-tests/src/test/scala/org/apache/pekko/http/impl/engine/http2/Http2FrameHpackSupport.scala:82:114: method get in class RightProjection is deprecated (since 2.13.0): Use `Either.toOption.get` instead
[warn]         case ("content-type", value) => old.withEntity(old.entity.withContentType(ContentType.parse(value).right.get))
```


He-Pin:
I think this performance is bad

PR #829, opened 2025-10-06
inline some functions used in parsing

pjfanning:
Scala 2 and 3 have different ways to inline so I split the classes into Scala 2 and 3 variants.


http-core/src/main/scala-2.13/org/apache/pekko/http/impl/engine/package.scala
```diff
@@ -36,17 +37,29 @@ package object parsing {
     case x                              => x.toString
   }
 
+  /**
+   * Like `byteChar` but doesn't throw `NotEnoughDataException` if the index is out of bounds.
+   * Used in places where we know that the index is valid because we checked the length beforehand.
+   */
+  @inline
+  private[http] def safeByteChar(input: ByteString, ix: Int): Char =
```
pjfanning, line 45: @jrudolph thanks for the review. I can look at doing a jmh benchmark but it will take me a week or 2 to get back to it.

One question though: do you think the safeByteChar function is a useful change instead of having a few places in our code where we do the masking in place while usually, we call the byteChar function below? There are places where we know it safe so that we don't need to waste a call on checking the index is in bounds.

PR #833, opened 2025-10-10
http2: remove single use ByteStringInputStream

### LINKED ISSUE #553: Twitter Decoder cannot handle all InputStream implementations
See https://github.com/twitter/hpack/issues/43

Works fine with ByteArrayInputStream but breaks if you use SequenceInputStream.

It is unlikely that the twitter/hpack issue will be fixed.

We could modify our fork of the HPack Decoder to avoid using `available()` method by
* adding private methods `boolean isEndOfStream()` and `byte nextByte()`
    * one implementation is to read one byte ahead 
    * read into a byte array and keep track of the pos and remaining bytes and refresh the byte array when remaining is zero
* find an InputStream implementation that we can use to wrap the provided stream but that gives us a way to work out if we are at the end of the stream without consuming bytes

jrudolph:
Also remove useless nesting / finally blocks, arrays do not need closing.

Fixes #553

Simpler variant of #820. WDYT, @pjfanning ?

### DIFF
--- a/http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/hpack/HeaderDecompression.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/hpack/HeaderDecompression.scala
@@ -92,8 +92,9 @@ private[http2] final class HeaderDecompression(masterHeaderParser: HttpHeaderPar
             }
           }
         }
+        val stream = payload.compact.asInputStream
         try {
-          decoder.decode(payload.compact.asInputStream, Receiver) // only compact ByteString supports InputStream with mark/reset
+          decoder.decode(stream, Receiver) // only compact ByteString supports InputStream with mark/reset
           decoder.endHeaderBlock() // TODO: do we have to check the result here?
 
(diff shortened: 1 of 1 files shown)


http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/hpack/HeaderDecompression.scala
```diff
@@ -92,9 +92,8 @@ private[http2] final class HeaderDecompression(masterHeaderParser: HttpHeaderPar
             }
           }
         }
-        val bis = ByteStringInputStream(payload)
         try {
-          decoder.decode(bis, Receiver)
+          decoder.decode(payload.compact.asInputStream, Receiver) // only compact ByteString supports InputStream with mark/reset
```
pjfanning: I'm not against this generally. One edge case is a ByteString that has a backing array but a non-zero offset.
We can get a safe InputStream directly from those ByteStreams without compacting them (an InputStream that supports mark/reset). I think in practice such ByteStrings are rare. So if we agree that such ByteStrings are not worrying about then I'm happy enough to go with this change.

PR #849, opened 2025-10-20
update http/2 docs to say the feature is not in preview

pjfanning:
* partial backport of #848 - some doc changes that were missed in #847 
* also deprecates a related method that was removed in #848


pjfanning:
the link validator issues are not related - javadoc.io has become unreliable for our checks, they could be rate limiting us

raboof (approved):

PR #906, opened 2025-12-19
Update PEKKO_VERSION to '1.x' in workflow

pjfanning:
* test with latest 1.x snapshot
* means we won't need to change this plan every time we do a minor release for pekko 1.x


nvollmar (approved):
lgtm

PR #923, opened 2026-01-14
chore: updates to release candidate staging

raboof:
looks good now AFAICS, I'll use this to release 2.0.0-M1 after merge.

### DIFF
--- a/.github/workflows/stage-release-candidate.yml
+++ b/.github/workflows/stage-release-candidate.yml
@@ -148,8 +148,10 @@ jobs:
           persist-credentials: false
 
       - name: Generate source archive
+        id: generate-source-archive
         run: |-
           VERSION=$(echo $REF | sed -e "s/.\(.*\)-.*/\\1/")
+          echo "version=$VERSION" >> $GITHUB_OUTPUT
           PREFIX=apache-pekko-http-$VERSION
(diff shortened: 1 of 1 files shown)


.github/workflows/stage-release-candidate.yml
```diff
@@ -327,7 +330,7 @@ jobs:
           
           https://pekko.apache.org/download.html#verifying-downloads
           
-          You can vote on ATR at https://release-test.apache.org/vote/pekko/$VERSION
+          You can vote on ATR at https://release-test.apache.org/vote/pekko-http/$VERSION
```
pjfanning, line 333: I had expected that the `pekko` in the URL was the project name and that we wouldn't be able to suffix it. If release-test.apache.org supports us using `pekko-http` then this is fine.

raboof: That is a good question. It is certainly possible to make `pekko-http` a 'sibling' project of `pekko` in ATR, but I've asked in their slack to double-check that this is how that is meant to be used.

raboof: confirmed, some docs in https://release-test.apache.org/docs/components

pjfanning (approved):
Lgtm

PR #837, opened 2025-10-17
support Jackson v3

pjfanning:
Idea is to leave pekko-http-jackson as supporting Jackson v2.x and to add a pekko-http-jackson3.

Jackson 3 is a significant rewrite and has lots of config changes compared to Jackson 2.

Aim is to backport this to 1.3.x branch to include in the 1.3.0 release.

PR #924, opened 2026-01-14
Release notes for 2.0.0-M1

raboof:
Kept it fairly short, I think it doesn't make sense to have an exhaustive list but we should collect some of the highlights when we get closer to a non-milestone 2.0.0 release.

### DIFF
--- a/docs/src/main/paradox/release-notes/releases-2.0.md
+++ b/docs/src/main/paradox/release-notes/releases-2.0.md
@@ -1,0 +1,14 @@
+# 0. Release Notes (2.0.x)
+
+## 2.0.0-M1
+
+This is milestone release and is aimed at testing this new major version
+by early adopters. This is experimental. This release should not be used in production.
+
+### Main changes
+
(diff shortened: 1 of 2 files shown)


docs/src/main/paradox/release-notes/releases-2.0.md
```diff
@@ -0,0 +1,14 @@
...
+This is milestone release and is aimed at testing this new major version
+by early adopters. This is experimental. This release should not be used in production.
+
+### Main changes
+
+* Pekko 2.0.0-M1 is the new minimum Pekko version
+* Java 17 is the new minimum JRE version
+* Scala 2.12 support dropped
+* A lot of deprecated code removed
+* A lot of pekko.util classes for Scala version compatibility have been removed
```
pjfanning, line 14: This last one is more of a pekko core change #783 and #837  might be worth highlighting. Maybe a generic mention of using latest dependency versions too.

pjfanning:
lgtm - merged

PR #940, opened 2026-01-28
Support appending header ranges to CorsSettings.

timw:
Provide a CorsSettings builder method to allow additional header ranges to be added, and the underlying ability to concatenate `HttpHeaderRange` instances to support it. 

This supports scenarios where complex `CorsSettings` defaults are provided by downstream libraries (e.g. pekko-grpc) but still need to be extended/adapted by end user applications.

A concrete motivating example is the pekko-grpc [defaultCorsSettings](https://github.com/apache/pekko-grpc/blob/df65ad955dc0dfc8d563eac6b959046a04590bb4/runtime/src/main/scala/org/apache/pekko/grpc/scaladsl/WebHandler.scala#L35C7-L35C26), which include:
```
  val defaultCorsSettings: CorsSettings = CorsSettings(ConfigFactory.load())
    .withAllowCredentials(true)
    .withAllowedMethods(immutable.Seq(HttpMethods.POST, HttpMethods.OPTIONS))
    .withExposedHeaders(immutable.Seq(headers.`Status`.name, headers.`Status-Message`.name, `Content-Encoding`.name))
    .withAllowedHeaders(
      HttpHeaderRange(
        "x-user-agent",
        "x-grpc-web",
        `Content-Type`.name,
        Accept.name,
        "grpc-timeout",
        `Accept-Encoding`.name))
```

In one of our (development) environments, we need the `Authorization` header to be added to the allowed headers range, but the `withAllowedHeaders` method only works by override, requiring us to either maintain our own distinct set of headers (which could diverge from the upstream) or do hacky matches on the unsealed `HttpHeaderRange` type.

This is a proposal for discussion on a more ergonomic/robust way to extend default settings like this.
The same pattern could equally be applied to the HTTP origin matchers and possibly the exposed headers.

I've tried to make the implementation in-line with conventions (e.g. I would prefer to seal `scaladsl.model.HttpHeaderRange` and have all the matches over it be required exhaustive, but the rest of pekko-http isn't really that way inclined), and the `concat` implementation is also awkward as exposing it to the Java DSL means you have to deal with those type params in the Scala methods and discourage extension.

Happy to discuss and rework if maintainers have opinions/better ideas about approach here.

### DIFF
--- a/http-cors/src/main/java/org/apache/pekko/http/cors/javadsl/model/HttpHeaderRange.java
+++ b/http-cors/src/main/java/org/apache/pekko/http/cors/javadsl/model/HttpHeaderRange.java
@@ -24,6 +24,11 @@ import org.apache.pekko.http.cors.scaladsl.model.HttpHeaderRange$;
 public abstract class HttpHeaderRange {
   public abstract boolean matches(String header);
 
+  /**
+   * Produces a new range that matches the headers of this range and the given range.
+   *
+   * @since 2.0.0
+   */
   public abstract HttpHeaderRange concat(HttpHeaderRange range);
 
   public static HttpHeaderRange create(String... headers) {
--- a/http-cors/src/main/scala/org/apache/pekko/http/cors/scaladsl/model/HttpHeaderRange.scala
+++ b/http-cors/src/main/scala/org/apache/pekko/http/cors/scaladsl/model/HttpHeaderRange.scala
@@ -28,6 +28,11 @@ import pekko.util.Helpers
 sealed abstract class HttpHeaderRange extends javadsl.model.HttpHeaderRange {
   override def concat(range: javadsl.model.HttpHeaderRange): HttpHeaderRange
 
+  /**
+   * Operator alias for [[concat]].
+   *
+   * @since 2.0.0
+   */
   def ++(range: javadsl.model.HttpHeaderRange): HttpHeaderRange = concat(range)
 }
 


pjfanning (commented):
This should target main branch, not 1.4.

When it gets merged, we can consider backporting to 1.4.

The code doesn't compile in Scala 2.12 and causes binary compatibility issues for other Scala versions. This will be less of an issue in main branch (no Scala 2.12 support, for example).

http-cors/src/main/scala/org/apache/pekko/http/cors/scaladsl/model/HttpHeaderRange.scala
```diff
@@ -17,22 +17,33 @@
 
 package org.apache.pekko.http.cors.scaladsl.model
 
-import org.apache.pekko
-import pekko.http.cors.javadsl
-import pekko.util.Helpers
+import org.apache.pekko.http.cors.javadsl
+import org.apache.pekko.util.Helpers
```
pjfanning: Please stick to code style. Don't mess with the imports like this.

timw: Corrected in latest push.

http-cors/src/main/scala/org/apache/pekko/http/cors/javadsl/settings/CorsSettings.scala
```diff
@@ -47,6 +47,7 @@ abstract class CorsSettings { self: CorsSettingsImpl =>
   def withAllowCredentials(newValue: Boolean): CorsSettings
   def withAllowedOrigins(newValue: HttpOriginMatcher): CorsSettings
   def withAllowedHeaders(newValue: HttpHeaderRange): CorsSettings
+  def withAdditionalAllowedHeaders(newValue: HttpHeaderRange): CorsSettings
```
pjfanning: Why do we need this? Can't you concat 2 HttpHeaderRanges based on another change in this PR? So then you can use the existing withAllowedHeaders to add the concatenated result.

If we absolutely must also have this extra method, it would be better named 'appendAllowedHeaders'.

timw: It’s not strictly required - just a little nicer syntax than:

`WebHandler.defaultCorsSettings.withAllowedHeaders(WebHandler.defaultCorsSettings.allowedHeaders ++ otherHeaders)`

So not material to the use case - happy to leave it out.

pjfanning: I'd prefer to leave it out

http-cors/src/main/java/org/apache/pekko/http/cors/javadsl/model/HttpHeaderRange.java
```diff
@@ -17,13 +17,17 @@
 
 package org.apache.pekko.http.cors.javadsl.model;
 
+import org.apache.pekko.annotation.DoNotInherit;
 import org.apache.pekko.http.impl.util.Util;
 import org.apache.pekko.http.cors.scaladsl.model.HttpHeaderRange$;
 
 /** @see HttpHeaderRanges for convenience access to often used values. */
+@DoNotInherit
```
pjfanning: can you remove this annotation?

http-cors/src/main/scala/org/apache/pekko/http/cors/javadsl/settings/CorsSettings.scala
pjfanning: the code is still here - can it be removed?

http-cors/src/main/scala/org/apache/pekko/http/cors/scaladsl/model/HttpHeaderRange.scala
```diff
@@ -23,16 +23,30 @@ import org.apache.pekko
...
 
   final case class Default(headers: Seq[String]) extends HttpHeaderRange {
     private val lowercaseHeaders: Seq[String] = headers.map(Helpers.toRootLowerCase)
     def matches(header: String): Boolean = lowercaseHeaders.contains(Helpers.toRootLowerCase(header))
+
+    override def concat(range: javadsl.model.HttpHeaderRange): HttpHeaderRange = {
+      range match {
+        case `*`              => `*`
+        case Default(headers) => Default(this.headers ++ headers)
+        case _                => throw new IllegalArgumentException(s"Unexpected header range implementation type ${range.getClass}")
```
pjfanning: Can the error message be more like `s"Concat is not supported for unexpected header range implementation type ${range.getClass}"`?

http-cors/src/main/java/org/apache/pekko/http/cors/javadsl/model/HttpHeaderRange.java
timw: Sure

http-cors/src/main/scala/org/apache/pekko/http/cors/scaladsl/model/HttpHeaderRange.scala
timw: I've actually switched this over to use `JavaMapping.toScala`, which is consistent with other param casting in the Scala DSLs. 
Let me know what you think.

timw:
> Ideally, the public methods should have javadoc with @since 2.0.0.

Pushed update with this.

pjfanning (approved):

pjfanning:
Thanks - merged

PR #1003, opened 2026-04-06
perf: use VarHandle int view in FrameEventParser.mask with byte-per-byte fallback

pjfanning:
## Summary

Optimizes the WebSocket frame masking/unmasking loop in `FrameEventParser` by using `MethodHandles.byteArrayViewVarHandle` (available since Java 9, part of the JDK standard library) to process 4 payload bytes as a single big-endian `int` per loop iteration, rather than 4 individual byte reads and writes.

## Changes

**`FrameEventParser.scala`**

- Added `import java.lang.invoke.{ MethodHandles, VarHandle }` and `import java.nio.ByteOrder`.
- Added a `private val intArrayViewHandle: VarHandle` initialized once at object-load time via `MethodHandles.byteArrayViewVarHandle(classOf[Array[Int]], ByteOrder.BIG_ENDIAN)`, wrapped in a `try/catch` that sets the field to `null` if the VarHandle cannot be created (e.g. under a restrictive security policy).
- Refactored `mask(bytes: ByteString, mask: Int)`:
  - **Fast path** (`intArrayViewHandle != null`): reads 4 bytes as one `int`, XORs with the full 32-bit mask in a single operation, then writes the result back — replacing 8 array accesses (4 reads + 4 writes) with 2 VarHandle operations per 4-byte group.
  - **Fallback path** (`intArrayViewHandle == null`): keeps the existing 4-byte-unrolled per-byte loop unchanged.
  - Both paths share the same trailing 1–3 byte remainder logic (XOR with `m0`/`m1`/`m2`).
- `m3` byte is now only computed inside the fallback path where it is actually needed.

## Why `classOf[Array[Int]]` is correct

`byteArrayViewVarHandle(viewArrayClass, byteOrder)` takes the **view** type — i.e. `int[].class` (≡ `classOf[Array[Int]]` in Scala) — because the handle views a `byte[]` *as if* it were an `int[]`. Passing `byte[].class` here would be incorrect and would throw `IllegalArgumentException` at runtime.

## Why plain `get`/`set` rather than `getAndBitwiseXor`

`getAndBitwiseXor` is an *atomic* access mode that inserts full memory barriers. The local mutable `byte[]` buffer is not shared between threads, so atomicity is not needed and its overhead would be counterproductive.

### DIFF
--- a/http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/FrameEventParser.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/FrameEventParser.scala
@@ -13,6 +13,9 @@
 
 package org.apache.pekko.http.impl.engine.ws
 
+import java.lang.invoke.{ MethodHandles, VarHandle }
+import java.nio.ByteOrder
+
 import org.apache.pekko
 import pekko.annotation.InternalApi
 import pekko.stream.impl.io.ByteStringParser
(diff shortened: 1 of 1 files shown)


pjfanning:
The benchmarks don't look good for this.

```
With Changes
[info] Benchmark                             Mode  Cnt       Score       Error  Units
[info] MaskingBench.benchRequestProcessing  thrpt    3   18617.921 ± 16149.899  ops/s

Without Changes
[info] Benchmark                             Mode  Cnt       Score       Error  Units
[info] MaskingBench.benchRequestProcessing  thrpt    3  156433.715 ±  5550.151  ops/s
```

PR #1017, opened 2026-04-18
fix: Recover handler errors, turn into HTTP 500 response (port from akka-http PR #4242)

pjfanning:
Port of https://github.com/akka/akka-http/pull/4242 to Pekko HTTP.
Now Apache licensed. https://github.com/akka/akka-http/releases/tag/v10.5.1

References: https://github.com/akka/akka-http/issues/4241

## Summary

When an HTTP/2 handler throws an exception (either synchronously or by failing a Future), the server previously propagated that error in a way that could tear down the connection. This change wraps the user-provided handler to catch both synchronous and asynchronous errors, converting them into HTTP 500 Internal Server Error responses, keeping the connection alive for further requests.

## Changes

- **`Http2.scala`**: Added `withErrorHandling` method that wraps the user handler to recover from `NonFatal` exceptions (sync throws and Future failures), returning an HTTP 500 response and logging the error. The wrapped handler is used for both HTTP/1 upgrade handling and the HTTP/2 stream handler.
- **`Http2HighlevelServerSpec.scala`** (new): Integration test verifying that when a handler fails, the server returns HTTP 500 and the connection remains usable for subsequent requests.

All imports and config keys updated from `akka.*` to `pekko.*` namespace.

### DIFF
--- a/http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/Http2.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/Http2.scala
@@ -93,10 +93,12 @@ private[http] final class Http2Ext(implicit val system: ActorSystem)
     val handlerWithErrorHandling = withErrorHandling(log, handler)
 
     val http1: HttpImplementation =
-      Flow[HttpRequest].mapAsync(settings.pipeliningLimit)(handleUpgradeRequests(handlerWithErrorHandling, settings, log))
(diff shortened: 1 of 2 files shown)


Philippus:
Failing on this part of the spec:
```
            8.1.2.1. Pseudo-Header Fields
              1: Sends a HEADERS frame that contains a unknown pseudo-header field
```

pjfanning:
closing in favour of #1041

PR #1021, opened 2026-04-27
Avoid unnecessary ByteString concat for strict HTTP/2 entities

He-Pin:
Motivation:
HTTP/2 DATA aggregation for small strict request entities may concatenate an empty `ByteString` with the incoming payload. That adds unnecessary work and allocation on unary/small-request gRPC-like paths without changing stream semantics.

Modification:
- keep the existing collecting state machine
- avoid `collectedData ++ payload` when either side is empty

Result:
The first DATA frame on the strict collection path now reuses the existing payload reference instead of allocating a concatenated `ByteString`. Non-empty aggregation behavior is unchanged.

Verification:
- `sbt "http-core / scalafmtCheck" "http2-tests / Test / testOnly org.apache.pekko.http.impl.engine.http2.Http2ServerSpec -- -z collect"`

Risk / Boundary:
- HTTP/2 stream state transitions are unchanged
- the change is limited to the empty/non-empty `ByteString` boundary
- existing HTTP/2 collection tests pass


pjfanning:
ByteString ++ already checks if either the left or right bytestring is empty and then does not return a new ByteStrings instance.

PR #1037, opened 2026-05-12
Fix ServerSentEventParser to handle CRLF line endings by stripping trailing CR

### LINKED ISSUE #797: ServerSentEventParser may not support event ends with `\r\n`
private val Field = """([^:]+): ?(.*)""".r
https://github.com/apache/pekko-http/blob/main/http/src/main/scala/org/apache/pekko/http/scaladsl/unmarshalling/sse/ServerSentEventParser.scala#L98

pjfanning:
fixes #797

Agent-Logs-Url: https://github.com/pjfanning/incubator-pekko-http/sessions/5bc4919b-f100-40bb-b6b9-c0c94294b4cb


pjfanning:
@He-Pin LineParser already handles the different line endings and ServerSentEventParser is never used without LineParser.

PR #1074, opened 2026-06-18
fix: Scala 3.9 compatibility fixes

He-Pin:
### Motivation
Scala 3.9.0-RC1 introduces several breaking changes:
1. `-language:_` wildcard option is no longer accepted
2. `-Yfuture-lazy-vals` option removed (behavior is now default)
3. `-Ywarn-dead-code` is Scala 2 only
4. Context bound desugaring changes, breaking explicit `ClassTag`/`Tuple`/`LogSource` passing patterns

### Modification
- **build.sbt**: Make `-language:_` conditional on Scala 3 minor < 9
- **project/Common.scala**: Move `-Ywarn-dead-code` to Scala 2 only, conditionally include `-Yfuture-lazy-vals` for Scala 3 < 3.9
- **HttpMessage.scala, RejectionHandler.scala, PredefinedToResponseMarshallers.scala**: Use `implicit val ClassTag` instead of context bound `[T: ClassTag]`
- **Directive.scala, BasicDirectives.scala, PathMatcher.scala**: Use `implicit val Tuple` instead of context bound `[L: Tuple]`
- **PoolInterface.scala, StageLoggingWithOverride.scala**: Use `implicit val LogSource` instead of explicit passing

### Result
pekko-http `parsing`, `http-core`, `http`, and `http-testkit` modules compile cleanly with Scala 3.9.0-RC1.

### Tests
- `sbt "++3.9.0-RC1!; parsing/compile; http-core/compile; http/compile; http-testkit/compile"` passes

### References
None - Scala 3.9 forward compatibility

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -108,7 +108,11 @@ lazy val parsing = project("parsing")
   .settings(AutomaticModuleName.settings("pekko.http.parsing"))
   .addPekkoModuleDependency("pekko-actor", "provided", PekkoCoreDependency.default)
   .settings(Dependencies.parsing)
-  .settings(scalacOptions += "-language:_")
+  .settings(scalacOptions ++= {
+    val v = scalaVersion.value
+    if (v.startsWith("3") && CrossVersion.partialVersion(v).exists(_._2 >= 9)) Seq.empty
+    else Seq("-language:_")
+  })
   .settings(scalaMacroSupport)
   .enablePlugins(ScaladocNoVerificationOfDiagrams)
   .enablePlugins(ReproducibleBuildsPlugin)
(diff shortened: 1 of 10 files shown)


build.sbt
```diff
@@ -108,7 +108,11 @@ lazy val parsing = project("parsing")
   .settings(AutomaticModuleName.settings("pekko.http.parsing"))
   .addPekkoModuleDependency("pekko-actor", "provided", PekkoCoreDependency.default)
   .settings(Dependencies.parsing)
-  .settings(scalacOptions += "-language:_")
+  .settings(scalacOptions ++= {
```
pjfanning: this is done in 3 places in this file - would it be possible to add a shared function or val to help here?

He-Pin (commented):
Done. Extracted `notOnScala39Plus` helper in `Common.scala` and replaced all 3 inline blocks.

pjfanning (approved):
lgtm

PR #1081, opened 2026-06-21
Optimize HTTP/2 request/response processing: eliminate double dispatch, reduce allocations, and streamline stream pipeline

He-Pin:
## Motivation

Profiling gRPC-over-HTTP/2 workloads with async-profiler revealed several performance bottlenecks in the HTTP/2 request/response pipeline:

1. **Double ExecutionContext dispatch**: `handleWithStreamIdHeader` wrapped the user handler call in `Future { }`, adding an unnecessary EC dispatch hop on top of `mapAsyncUnordered`'s own scheduling
2. **Redundant pattern matching**: `Http2Demux.onPush` performed two separate pattern matches per incoming frame
3. **Per-frame lambda allocations**: `handleStreamEvent`, `updateState`, `handleOutgoingCreated`, and `handleOutgoingEnded` each created lambda closures per invocation
4. **Per-response OutHandler allocation**: `HeaderCompression` created a new `OutHandler` for each `CompositeFrame` continuation
5. **Missing synchronous fast path**: `withErrorHandling` always called `.recover` even for already-completed successful futures

### Request path (3 commits)
- **Eliminate double EC dispatch** (`cc2531b8b`): Call user handler directly in `mapAsyncUnordered` lambda instead of wrapping in `Future { }`. The `mapAsyncUnordered` stage already schedules on the EC, so the extra `Future { }` wrapper doubled the scheduling overhead.
- **HPACK header parsing cache** (`331e94ac8`): Cache common gRPC headers (`:method`, `:path`, `:scheme`, `content-type`) in `HeaderDecompression` to avoid repeated parsing.
- **Merge double pattern match** (`84a77de46`): Combine two separate pattern matches in `Http2Demux.onPush` into a single match, eliminating redundant frame type checks.

### Frame handling (3 commits)
- **Coalesce data+trailer frames** (`b11508060`): Render DATA and HEADERS frames into a single buffer allocation in `FrameRenderer`, reducing per-response buffer allocations.
- **HeaderCompression continuation fusing** (`7f666f45c`): Replace per-`CompositeFrame` `OutHandler` allocation with a state field in `HeaderCompression`'s `GraphStageLogic`, draining continuation frames without new object creation.
- **`withErrorHandling` fast path** (`a785e7023`): Check `future.value` before calling `.recover`; for already-completed successful futures (the common case in gRPC unary), skip the `.recover` allocation entirely.

### Stream processing (4 commits)
- **RequestErrorFlow handler merge** (`f18c4b664`): Merge the response path handler into `RequestErrorFlow`'s `GraphStageLogic` (using `with InHandler with OutHandler`), eliminating 2 handler object allocations per materialization.
- **Inline `updateState`** (`56890d1ab`): Extract `commitStreamState` bookkeeping method and inline state transitions in `handleStreamEvent`, eliminating the per-call `x => (handle(x), ())` lambda wrapper.
- **`handleStreamEvent` lambda elimination** (`ae71cfbd0`): Inline the `_.handle(e)` lambda in `Http2Demux.handleStreamEvent`, using `commitStreamState` directly.
- **`handleOutgoingCreated/Ended` lambda elimination** (`9cea60bad`): Inline state transitions in `handleOutgoingCreated` and `handleOutgoingEnded`, eliminating 2 lambda allocations per response.

Note: This branch also contains HeaderPairs-related commits that were reverted (`5c24d3131`, `6449372d8`, `5bbd883fa`, `014156000`). The net effect of these 4 commits is zero — they cancel each other out. The effective changes are the 10 commits listed above.

## Result

Benchmarked with `ghz` (complex_proto, 1000 concurrency, 50 connections, 120s, SerialGC, pekko-grpc optimized):

| Metric | Baseline (1.2.0) | Optimized | Improvement |
|--------|-----------------|-----------|-------------|
| Throughput (ForkJoinPool) | 62,899 req/s | 73,584 req/s | **+17.0%** |
| P99 latency (ForkJoinPool) | 32.46 ms | 29.20 ms | **-10.0%** |
| Avg latency (ForkJoinPool) | 8.34 ms | 6.77 ms | **-18.8%** |
| Throughput (AffinityPool) | — | 78,348 req/s | **+24.6%** |
| P99 latency (AffinityPool) | — | 24.17 ms | **-25.5%** |

Allocation profiling (async-profiler) confirms reduced per-request allocations in the HTTP/2 pipeline.

## Tests

- `sbt http-core / Test / test`
- All HTTP/2 related tests pass

## References

- Profiling with async-profiler (CPU + allocation)
- Benchmark suite: [grpc_bench](https://github.com/LesnyRumcajs/grpc_bench) scala_pekko_bench
- Companion PR: apache/pekko-grpc#739

### DIFF
--- a/http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/Http2Blueprint.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/Http2Blueprint.scala
@@ -303,17 +303,26 @@ private[http] object Http2Blueprint {
       implicit ec: ExecutionContext): Flow[HttpRequest, HttpResponse, NotUsed] =
     Flow[HttpRequest]
       .mapAsyncUnordered(parallelism) { req =>
-        // The handler itself may do significant work so make sure to schedule it separately. This is especially important for HTTP/2 where it is expected that
-        // multiple requests are handled concurrently on the same connection. The complete stream including `mapAsyncUnordered` shares one GraphInterpreter, so
-        // that this extra indirection will guard the GraphInterpreter from being starved by user code.
-        Future {
-          val response = handler(req)
+        // mapAsyncUnordered already schedules on the execution context,
+        // so call the handler directly without wrapping in Future { } to avoid
+        // an extra EC dispatch hop. This is significant for fast handlers
+        // (e.g. gRPC unary handlers returning Future.successful) where the
+        // extra hop would double the scheduling overhead.
+        val response = try handler(req) catch { case scala.util.control.NonFatal(ex) => Future.failed(ex) }
 
-          req.attribute(Http2.streamId) match {
-            case Some(streamIdHeader) => response.map(_.addAttribute(Http2.streamId, streamIdHeader)) // add stream id attribute when request had it
-            case None                 => response
-          }
-        }.flatten
+        req.attribute(Http2.streamId) match {
+          case Some(streamIdHeader) =>
+            // Fast path: if response is already completed, add attribute synchronously
+            response.value match {
+              case Some(scala.util.Success(resp)) =>
+                Future.successful(resp.addAttribute(Http2.streamId, streamIdHeader))
+              case Some(scala.util.Failure(ex)) =>
+                Future.failed(ex)
+              case None =>
+                response.map(_.addAttribute(Http2.streamId, streamIdHeader))
+            }
+          case None => response
+        }
       }
 
   private[http2] def logParsingError(info: ErrorInfo, log: LoggingAdapter,
--- a/http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/Http2Blueprint.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/Http2Blueprint.scala
@@ -308,7 +308,8 @@ private[http] object Http2Blueprint {
         // an extra EC dispatch hop. This is significant for fast handlers
         // (e.g. gRPC unary handlers returning Future.successful) where the
         // extra hop would double the scheduling overhead.
-        val response = try handler(req) catch { case scala.util.control.NonFatal(ex) => Future.failed(ex) }
+        val response = try handler(req)
+        catch { case scala.util.control.NonFatal(ex) => Future.failed(ex) }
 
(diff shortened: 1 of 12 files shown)


He-Pin:
The AffinityPool gives me better number  than the ForkJoinPool on my local pc

http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/Http2Blueprint.scala
```diff
@@ -303,17 +303,26 @@ private[http] object Http2Blueprint {
...
+            response.value match {
+              case Some(scala.util.Success(resp)) =>
+                Future.successful(resp.addAttribute(Http2.streamId, streamIdHeader))
+              case Some(scala.util.Failure(ex)) =>
+                Future.failed(ex)
+              case None =>
+                response.map(_.addAttribute(Http2.streamId, streamIdHeader))
+            }
+          case None => response
+        }
```
He-Pin, line 326: @pjfanning Should I extract these commits one by one by one as seperated PR?

pjfanning: I'm ok with leaving this as is. I'm hoping that other people will review this but if not, I'll ok this in a few days.

pjfanning:
@He-Pin could you fix the scalafmt and look into the test failure in CI?

jrudolph:
Nice investigation!

I haven't looked at all points but "Double ExecutionContext dispatch: handleWithStreamIdHeader wrapped the user handler call in Future { }, adding an unnecessary EC dispatch hop on top of mapAsyncUnordered's own scheduling" was an explicit feature. The question is how `mapAsyncUnordered` schedules the Future-returning thunks these days. It used to be the case, that the thunks themselves were executed synchronously and could block the stream engine. In particular, the optimizations in FastFuture could mean that a whole routing structure and request processing could run synchronously. Without the existing wrapping Future, we might de-facto serialize request processing on a single http2 connection.

He-Pin:
I’ve considered that, too; it would help avoid heavier renewal overhead for users. However, it’s not very friendly to the `grpcbench` benchmarks, so perhaps we could look into making it configurable.

Issue #1087, opened 2026-06-24
Add directional tests for control characters in the request line


He-Pin:
### Motivation

A request-smuggling class of vulnerability exists when HTTP parsers silently strip boundary control bytes (NUL, CR, LF, VT, FF) from the version token before matching — see [netty/netty#16970](https://github.com/netty/netty/issues/16970).

Pekko HTTP uses a fundamentally different parsing strategy — byte-by-byte exact matching at fixed offsets in `HttpMessageParser.parseProtocol` (`HttpMessageParser.scala:130-140`) with no `trim()` or whitespace/control-character stripping. Analysis confirms that Pekko HTTP is **not affected** by this vulnerability: the NUL byte is consumed into the URI by `parseRequestTarget` and rejected by the URI parser before the version token is ever reached.

However, no existing test in `RequestParserSpec` exercises this code path with control characters. A directional test should be added to lock in this behavior and prevent regressions from future parsing refactors.

### Modification

Add directional test cases to `RequestParserSpec` that verify control characters (NUL) in the request line are rejected with an appropriate error status.

### Result

The test suite will explicitly cover the control-character-in-request-line scenario, ensuring that Pekko HTTP continues to reject such malformed requests and that any future parsing changes do not introduce a trim-based vulnerability.

### References

- Netty issue: https://github.com/netty/netty/issues/16970

PR #1088, opened 2026-06-24
Add directional test for NUL byte in request target

He-Pin:
### Motivation

A request-smuggling class of vulnerability exists when HTTP parsers silently strip boundary control bytes (NUL, CR, LF, VT, FF) from the version token before matching — see [netty/netty#16970](https://github.com/netty/netty/issues/16970).

Pekko HTTP uses byte-level exact matching at fixed offsets in `HttpMessageParser.parseProtocol` (`HttpMessageParser.scala:130-140`) — no `trim()` or whitespace/control-character stripping. Analysis confirms Pekko HTTP is **not affected** by this vulnerability: a NUL byte in the request line is consumed into the URI by `parseRequestTarget` and rejected by the URI parser before the version token is ever reached.

However, no existing test in `RequestParserSpec` exercises this code path with control characters. This PR adds a directional test to lock in the behavior and prevent regressions from future parsing refactors.

### Modification

Add a directional test case `"a NUL byte in the request target"` in `RequestParserSpec` that feeds a NUL byte inside the request target and asserts the parser rejects the request with `BadRequest` and an `"Illegal request-target"` error summary.

### Result

The test suite now explicitly covers control-character rejection in the request line, guarding against regressions from future parsing refactors that might introduce `trim()`-like behavior.

### Tests

- `sbt "http-core / Test / testOnly *RequestParserCRLFSpec *RequestParserLFSpec"` — 106 tests passed, 0 failed

### References

- Refs #1087
- Refs https://github.com/netty/netty/issues/16970

### DIFF
--- a/http-core/src/test/scala/org/apache/pekko/http/impl/engine/parsing/RequestParserSpec.scala
+++ b/http-core/src/test/scala/org/apache/pekko/http/impl/engine/parsing/RequestParserSpec.scala
@@ -625,6 +625,16 @@ abstract class RequestParserSpec(mode: String, newLine: String) extends AnyFreeS
           ErrorInfo("The server does not support the HTTP protocol version used in the request."))
       }
 
+      "a NUL byte in the request target" in new Test {
+        val result = multiParse(newParser)(Seq("GET /\u0000HTTP/1.1 HTTP/1.1\r\n"))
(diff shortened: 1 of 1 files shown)


pjfanning (approved):
lgtm

PR #1089, opened 2026-06-25
refactor: migrate Collections/Arrays.asList to Java 9+ collection factories

He-Pin:
### Motivation
Java 9+ provides `List.of()`, `Map.of()`, `Set.of()` factory methods that are more concise and return truly immutable collections.

### Modification
- Replace `Arrays.asList(...)` with `List.of(...)` across 14 files
- Replace `Collections.singletonList()` with `List.of()`, `Collections.emptyList/Map/Set()` with `List/Map/Set.of()`
- Replace `new HashMap<>()` + `put()` chains with `Map.of()`
- Replace `new HashSet<>(Arrays.asList(...))` with `Set.of(...)`
- Correctly skips shaded `com.twitter.hpack` code and array-parameter conversions in `Util.java` and `MediaTypes.java`

### Result
Cleaner code using Java 17 immutable collection factories. 20 files changed.

### Tests
Not run - compile-only test file changes

### References
None - Java 17 API modernization

### DIFF
--- a/http-core/src/main/java/org/apache/pekko/http/javadsl/model/MediaTypes.java
+++ b/http-core/src/main/java/org/apache/pekko/http/javadsl/model/MediaTypes.java
@@ -587,7 +587,7 @@ public final class MediaTypes {
   public static MediaType.Binary customBinary(
       String mainType, String subType, boolean compressible) {
     return customBinary(
-        mainType, subType, compressible, java.util.Collections.<String, String>emptyMap(), false);
+        mainType, subType, compressible, java.util.Map.<String, String>of(), false);
   }
 
   public static MediaType.Binary customBinary(
@@ -596,7 +596,7 @@ public final class MediaTypes {
         mainType,
         subType,
         compressibility,
-        java.util.Collections.<String, String>emptyMap(),
+        java.util.Map.<String, String>of(),
         false);
   }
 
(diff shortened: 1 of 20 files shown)


http-core/src/main/java/org/apache/pekko/http/javadsl/model/Multiparts.java
```diff
@@ -109,7 +108,7 @@ public static Multipart.FormData.BodyPart createFormDataPartFromPath(
   public static Multipart.FormData.BodyPart createFormDataBodyPart(
       String name, BodyPartEntity entity) {
     List nil = Nil$.MODULE$;
-    Map<String, String> additionalDispositionParams = Collections.emptyMap();
```
pjfanning, line 112: I prefer Collections.emptyMap() - does not create a new map

http-core/src/main/java/org/apache/pekko/http/javadsl/model/MediaTypes.java
```diff
@@ -596,7 +596,7 @@ public static MediaType.Binary customBinary(
         mainType,
         subType,
         compressibility,
-        java.util.Collections.<String, String>emptyMap(),
+        java.util.Map.<String, String>of(),
```
pjfanning: I don't like the new map here, I prefer the singleton that we already use

pjfanning (approved):
lgtm

PR #1098, opened 2026-06-26
add since markers on Deprecated annotations

pjfanning:
allowed now that we don't support Java 8

I am looking separately at whether we can remove some of these methods but they seem to be used heavily in our tests and the alternative code is more complicated.


He-Pin (approved):
thanks

PR #1104, opened 2026-06-27
make withLevel package private

pjfanning:
some code that is marked deprecated is actually used by the JavaDSL code under the hood

### DIFF
--- a/http/src/main/scala/org/apache/pekko/http/scaladsl/coding/Deflate.scala
+++ b/http/src/main/scala/org/apache/pekko/http/scaladsl/coding/Deflate.scala
@@ -18,7 +18,7 @@ import pekko.annotation.InternalApi
 import pekko.http.scaladsl.model._
 import pekko.http.scaladsl.model.headers.HttpEncodings
 
-@InternalApi // used by javadsl.Coder
+@InternalApi // used by javadsl.coding.Coder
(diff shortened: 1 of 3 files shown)


http/src/main/scala/org/apache/pekko/http/scaladsl/coding/Deflate.scala
```diff
@@ -18,7 +18,7 @@ import pekko.annotation.InternalApi
 import pekko.http.scaladsl.model._
 import pekko.http.scaladsl.model.headers.HttpEncodings
 
-@InternalApi
+@InternalApi // used by javadsl.Coder
```
He-Pin: btw, that's the mean of `Coder`? do you mean code?

pjfanning: There is a class called Coder, maybe could be better called Encoder but it's called Coder for years.

pjfanning: https://github.com/apache/pekko-http/blob/main/http/src/main/java/org/apache/pekko/http/javadsl/coding/Coder.java

pjfanning:
@He-Pin is it ok to merge this as is. Renaming the classes is not really feasible with deprecating the old classes as well. I think the names are better left as is. I'm a native English speaker and while Encoder would be more consistent, Coder would still be an acceptable name for these classes too.

He-Pin (approved):

PR #1105, opened 2026-06-27
remove deprecated javadsl route directive

pjfanning:
RouteDirectives.concat is the documented replacment


He-Pin (approved):

He-Pin:
Maybe we can deprecate the Coder and add a `codec` package, and  `Encoder` `Decoder` just as Netty.

PR #1109, opened 2026-06-28
build: enable -Xfatal-warnings for Scala 2 and -Werror for Scala 3

He-Pin:
### Motivation
Treating warnings as errors ensures that deprecated API usage and other code quality issues are caught at compile time rather than accumulating silently.

### Modification
- Add `-Xfatal-warnings` to Scala 2 `scalacOptions` in `Common.scala`
- Add `-Werror` to Scala 3 `scalacOptions`
- Remove `-Wconf:cat=deprecation:s` for Scala 3 so deprecation warnings are no longer silenced
- Exclude `-Xfatal-warnings` from `Compile / doc` scope to avoid doc generation failures

### Result
All compiler warnings are now treated as errors, enforcing higher code quality and catching deprecated API usage at compile time.

### Tests
- CI will verify compilation

### References
None — build configuration improvement

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -341,7 +341,9 @@ lazy val httpScalafixRules =
   Project(id = "http-scalafix-rules", base = file("http-scalafix/scalafix-rules"))
     .settings(
       name := "pekko-http-scalafix-rules",
-      libraryDependencies += Dependencies.Compile.scalafix)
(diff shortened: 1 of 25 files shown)


pjfanning (approved):
lgtm

PR #1114, opened 2026-06-30
Add server-side WebSocket compression

### LINKED ISSUE #61: Please add support for WebSocket compression
See discussion in akka-http, there a links to RFC's etc.: https://github.com/akka/akka-http/issues/659

mkurz:
## What changed

Adds server-side support for RFC 7692 `permessage-deflate` WebSocket compression.

Compression is negotiated during the WebSocket handshake when the client offers `permessage-deflate` in `Sec-WebSocket-Extensions`. It is enabled by default and can be disabled globally through configuration.

Also adds an API option to decline negotiated compression for a single accepted WebSocket upgrade.

- Fixes #61.

## Details

This PR adds:

- `permessage-deflate` negotiation for server-side WebSockets
- inbound compressed message inflation
- outbound message deflation
- fragmented compressed message handling
- context takeover and no-context-takeover negotiation
- configurable compression level
- configurable max decompressed allocation
- configurable preferred `client_max_window_bits`
- per-upgrade compression disabling via new `WebSocketUpgrade` overloads
- documentation, reference config, release notes, and MiMa filters

## Netty comparison

The implementation follows Netty’s `permessage-deflate` behavior where it makes sense for Pekko HTTP:

- compressed inbound messages are inflated before they reach the application
- outbound text/binary messages are compressed when compression was negotiated
- fragmented compressed messages keep RSV1 only on the first frame
- control frames can be interleaved with fragmented compressed messages
- context takeover is retained by default
- `server_no_context_takeover` and `client_no_context_takeover` can be negotiated
- `client_max_window_bits` can be negotiated
- decompressed output is bounded through `max-allocation`

One deliberate difference is zlib implementation scope. Netty also uses the JDK `Deflater` and `Inflater` APIs for the default zlib settings, and only switches to JZlib when non-default zlib `windowBits` or `memLevel` support is needed. Pekko HTTP currently uses only the JDK `Deflater` and `Inflater` APIs, like Tomcat and Jetty. The JDK API does not expose zlib `windowBits` or `memLevel`, so Pekko HTTP does not accept `server_max_window_bits` values below `15` and does not provide server window-size or memory-level settings.

Clients may still request `client_max_window_bits`; when they do, Pekko HTTP can include the configured `preferred-client-window-size` in the handshake response to ask the client to use that window size for client-to-server messages.

## Security notes

WebSocket compression can increase CPU and memory usage.

The default context takeover settings match common server defaults and favor compression efficiency. Applications that need a more conservative setup can configure no-context-takeover behavior or disable compression globally/per accepted WebSocket.

`max-allocation` limits decompressed message size. If the limit is exceeded while inflating a compressed message, Pekko HTTP closes the connection with a WebSocket protocol error.

## Testing

Ran:

```bash
sbt "http-core / Test / testOnly org.apache.pekko.http.impl.engine.ws.WebSocketServerSpec"
sbt +mimaReportBinaryIssues
```

Also verified Play Framework integration against the locally published Pekko HTTP artifact:
(This is a WIP pull request for Play which I will submit tomorrow - I will you ping you then)

```bash
sbt -Dpekko.http.version=<local-version> \
  "Play-Integration-Test / Test / testOnly play.it.http.websocket.PekkoHttpWebSocketCompressionSpec"

sbt -Dpekko.http.version=<local-version> \
  "Play-Integration-Test / Test / testOnly play.it.http.websocket.PekkoHttpWebSocketSpec"
```

## References

- https://github.com/playframework/playframework/issues/4186
- https://github.com/akka/akka-http/issues/659
- https://github.com/playframework/playframework/issues/10863
- https://github.com/playframework/playframework/pull/11169
- https://github.com/playframework/playframework/pull/12477
- https://github.com/playframework/netty-reactive-streams/pull/126
- https://github.com/apache/pekko-http/issues/61

### DIFF
--- a/http-core/src/main/resources/reference.conf
+++ b/http-core/src/main/resources/reference.conf
@@ -348,6 +348,44 @@ pekko.http {
 
       # Enable verbose debug logging for all ingoing and outgoing frames
       log-frames = false
+
+      compression {
+        # Whether the server should support WebSocket compression using the RFC 7692
+        # permessage-deflate extension. Compression is negotiated during the
+        # WebSocket handshake and is only used when the client requests it.
+        enabled = true
+
+        # Maximum size of a decompressed WebSocket message. If this value is
+        # exceeded while inflating a compressed message, the connection is closed
+        # with a WebSocket protocol error.
+        # Set to 0 to disable this limit.
+        max-allocation = 64k
+
+        permessage-deflate {
+          # Pekko HTTP uses the JDK Deflater/Inflater implementation for
+          # permessage-deflate. The JDK API does not expose zlib windowBits or
+          # memLevel, so clients cannot negotiate a server_max_window_bits value
+          # below 15. client_max_window_bits can still be negotiated through
+          # preferred-client-window-size.
+
+          # DEFLATE compression level used for server-to-client messages. Valid
+          # values are 0-9, where 0 uses no compression and 9 favors compression
+          # ratio over CPU usage.
+          compression-level = 6
+
+          # The client_max_window_bits value Pekko HTTP should request for
+          # client-to-server messages when the client sends client_max_window_bits
+          # without an explicit value. Valid values are 8-15.
+          preferred-client-window-size = 15
+
+          # Whether a client may request server_no_context_takeover.
+          allow-server-no-context = false
+
+          # Whether Pekko HTTP should request client_no_context_takeover when the
+          # client indicates that it supports this parameter.
+          preferred-client-no-context = false
+        }
+      }
     }
   }
 
--- a/http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/Handshake.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/Handshake.scala
@@ -136,20 +136,34 @@ private[http] object Handshake {
 
             def handle(
                 handler: Either[Graph[FlowShape[FrameEvent, FrameEvent], Any], Graph[FlowShape[Message, Message], Any]],
-                subprotocol: Option[String]): HttpResponse = {
+                subprotocol: Option[String],
+                compressionEnabled: Boolean): HttpResponse = {
               require(
                 subprotocol.forall(chosen => clientSupportedSubprotocols.contains(chosen)),
                 s"Tried to choose invalid subprotocol '$subprotocol' which wasn't offered by the client: [${requestedProtocols.mkString(", ")}]")
-              buildResponse(key.get, handler, subprotocol, perMessageDeflate, settings, log)
+              val acceptedPerMessageDeflate = if (compressionEnabled) perMessageDeflate else None
+              buildResponse(key.get, handler, subprotocol, acceptedPerMessageDeflate, settings, log)
             }
 
             def handleFrames(
                 handlerFlow: Graph[FlowShape[FrameEvent, FrameEvent], Any], subprotocol: Option[String]): HttpResponse =
-              handle(Left(handlerFlow), subprotocol)
+              handle(Left(handlerFlow), subprotocol, compressionEnabled = true)
+
+            override private[http] def handleFrames(
+                handlerFlow: Graph[FlowShape[FrameEvent, FrameEvent], Any],
+                subprotocol: Option[String],
+                compressionEnabled: Boolean): HttpResponse =
+              handle(Left(handlerFlow), subprotocol, compressionEnabled)
 
             override def handleMessages(handlerFlow: Graph[FlowShape[Message, Message], Any],
                 subprotocol: Option[String] = None): HttpResponse =
-              handle(Right(handlerFlow), subprotocol)
+              handle(Right(handlerFlow), subprotocol, compressionEnabled = true)
+
+            override def handleMessages(
+                handlerFlow: Graph[FlowShape[Message, Message], Any],
+                subprotocol: Option[String],
+                compressionEnabled: Boolean): HttpResponse =
+              handle(Right(handlerFlow), subprotocol, compressionEnabled)
           }
           OptionVal.Some(header)
         } else OptionVal.None
--- a/http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/Handshake.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/Handshake.scala
@@ -23,6 +23,7 @@ import scala.collection.immutable.Seq
 import pekko.event.LoggingAdapter
 import pekko.http.impl.util._
 import pekko.http.impl.engine.server.UpgradeToOtherProtocolResponseHeader
+import pekko.http.impl.settings.WebSocketSettingsImpl
 import pekko.http.scaladsl.model.headers._
 import pekko.http.scaladsl.model.ws.Message
 import pekko.http.scaladsl.model._
@@ -122,6 +123,13 @@ private[http] object Handshake {
             case OptionVal.Some(p) => p.protocols
             case _                 => Nil
           }
+          val clientRequestedExtensions = headers.collect {
+            case extensions: `Sec-WebSocket-Extensions` => extensions.extensions
+          }.flatten
+          val perMessageDeflate =
+            PerMessageDeflate.negotiate(
+              clientRequestedExtensions,
+              settings.asInstanceOf[WebSocketSettingsImpl].compression)
 
           val header = new UpgradeToWebSocketLowLevel {
             def requestedProtocols: Seq[String] = clientSupportedSubprotocols
@@ -132,7 +140,7 @@ private[http] object Handshake {
               require(
                 subprotocol.forall(chosen => clientSupportedSubprotocols.contains(chosen)),
                 s"Tried to choose invalid subprotocol '$subprotocol' which wasn't offered by the client: [${requestedProtocols.mkString(", ")}]")
-              buildResponse(key.get, handler, subprotocol, settings, log)
+              buildResponse(key.get, handler, subprotocol, perMessageDeflate, settings, log)
             }
 
             def handleFrames(
@@ -169,11 +177,16 @@ private[http] object Handshake {
      */
     def buildResponse(key: `Sec-WebSocket-Key`,
         handler: Either[Graph[FlowShape[FrameEvent, FrameEvent], Any], Graph[FlowShape[Message, Message], Any]],
-        subprotocol: Option[String], settings: WebSocketSettings, log: LoggingAdapter): HttpResponse = {
+        subprotocol: Option[String],
+        perMessageDeflate: Option[PerMessageDeflate.Negotiated],
+        settings: WebSocketSettings,
+        log: LoggingAdapter): HttpResponse = {
       val frameHandler = handler match {
-        case Left(frameHandler)    => frameHandler
+        case Left(frameHandler) =>
+          perMessageDeflate.map(_.frameEventBidiFlow(settings.randomFactory).join(frameHandler)).getOrElse(frameHandler)
         case Right(messageHandler) =>
-          WebSocket.stack(serverSide = true, settings, log = log).join(messageHandler)
+          WebSocket.stack(serverSide = true, settings, perMessageDeflate = perMessageDeflate, log = log)
+            .join(messageHandler)
       }
 
       HttpResponse(
@@ -182,7 +195,9 @@ private[http] object Handshake {
         List(
           UpgradeHeader,
           ConnectionUpgradeHeader,
-          `Sec-WebSocket-Accept`.forKey(key),
+          `Sec-WebSocket-Accept`.forKey(key)) :::
+        perMessageDeflate.map(p => `Sec-WebSocket-Extensions`(Seq(p.responseExtension))).toList :::
+        List(
           UpgradeToOtherProtocolResponseHeader(WebSocket.framing.join(frameHandler))))
     }
   }
--- a/http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
@@ -1,0 +1,345 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.pekko.http.impl.engine.ws
+
+import java.io.ByteArrayOutputStream
+import java.util.Random
+import java.util.zip.Deflater
+import java.util.zip.Inflater
+import java.util.zip.DataFormatException
+
+import org.apache.pekko
+import pekko.NotUsed
+import pekko.annotation.InternalApi
+import pekko.http.impl.settings.WebSocketCompressionSettingsImpl
+import pekko.http.scaladsl.model.headers.WebSocketExtension
+import pekko.stream.scaladsl.BidiFlow
+import pekko.stream.scaladsl.Flow
+import pekko.stream.stage.GraphStage
+import pekko.stream.stage.GraphStageLogic
+import pekko.stream.stage.InHandler
+import pekko.stream.stage.OutHandler
+import pekko.stream.{ Attributes, FlowShape, Inlet, Outlet }
+import pekko.util.ByteString
+
+import scala.collection.immutable
+import scala.collection.immutable.ListMap
+
+/**
+ * INTERNAL API
+ */
+@InternalApi
+private[http] object PerMessageDeflate {
+  private val ExtensionName = "permessage-deflate"
+  private val ClientMaxWindowBits = "client_max_window_bits"
+  private val ServerMaxWindowBits = "server_max_window_bits"
+  private val ClientNoContextTakeover = "client_no_context_takeover"
+  private val ServerNoContextTakeover = "server_no_context_takeover"
+  private val EmptyStoredBlock = ByteString(0x00, 0x00, 0xFF.toByte, 0xFF.toByte)
+
+  final case class Negotiated(
+      responseExtension: WebSocketExtension,
+      serverNoContextTakeover: Boolean,
+      clientNoContextTakeover: Boolean,
+      settings: WebSocketCompressionSettingsImpl) {
+    def bidiFlow: BidiFlow[FrameEventOrError, FrameEventOrError, FrameEvent, FrameEvent, NotUsed] =
+      BidiFlow.fromFlows(inflaterFlow, deflaterFlow)
+
+    def frameEventBidiFlow(
+        maskRandom: () => Random): BidiFlow[FrameEvent, FrameEvent, FrameEvent, FrameEvent, NotUsed] =
+      BidiFlow.fromFlows(
+        Flow[FrameEvent]
+          .via(Masking.unmaskIf(condition = true))
+          .via(inflaterFlow)
+          .map {
+            case frame: FrameEvent => frame
+            case FrameError(ex)    => throw ex
+          }
+          .via(Masking.maskIf(condition = true, maskRandom)),
+        deflaterFlow)
+
+    private def inflaterFlow: Flow[FrameEventOrError, FrameEventOrError, NotUsed] =
+      Flow.fromGraph(new LifecycleMapConcatStage(
+        "PerMessageDeflate.inflater",
+        () => new InflaterFlow(clientNoContextTakeover, settings)))
+
+    private def deflaterFlow: Flow[FrameEvent, FrameEvent, NotUsed] =
+      Flow.fromGraph(new LifecycleMapConcatStage(
+        "PerMessageDeflate.deflater",
+        () => new DeflaterFlow(serverNoContextTakeover, settings)))
+  }
+
+  def negotiate(
+      requested: immutable.Seq[WebSocketExtension],
+      settings: WebSocketCompressionSettingsImpl): Option[Negotiated] = {
+    if (!settings.enabled) None
+    else {
+      requested.collectFirst(Function.unlift { extension =>
+        if (extension.name.equalsIgnoreCase(ExtensionName)) negotiate(extension, settings) else None
+      })
+    }
+  }
+
+  private def negotiate(
+      extension: WebSocketExtension,
+      settings: WebSocketCompressionSettingsImpl): Option[Negotiated] = {
+    var responseParams = ListMap.empty[String, String]
+    var clientNoContext = false
+    var serverNoContext = false
+    var accepted = true
+
+    extension.params.foreach {
+      case (ClientMaxWindowBits, value) =>
+        if (value.isEmpty) responseParams += ClientMaxWindowBits -> settings.preferredClientWindowSize.toString
+        else if (validWindowBits(value)) responseParams += ClientMaxWindowBits -> value
+        else accepted = false
+      case (ServerMaxWindowBits, value) =>
+        if (value == "15") responseParams += ServerMaxWindowBits -> value
+        else accepted = false
+      case (ClientNoContextTakeover, "") =>
+        clientNoContext = settings.preferredClientNoContext
+        if (clientNoContext) responseParams += ClientNoContextTakeover -> ""
+      case (ServerNoContextTakeover, "") =>
+        if (settings.allowServerNoContext) {
+          serverNoContext = true
+          responseParams += ServerNoContextTakeover -> ""
+        } else accepted = false
+      case _ =>
+        accepted = false
+    }
+
+    if (accepted) {
+      Some(Negotiated(WebSocketExtension(ExtensionName, responseParams), serverNoContext, clientNoContext, settings))
+    } else None
+  }
+
+  private def validWindowBits(value: String): Boolean =
+    value.length <= 2 && value.forall(_.isDigit) && {
+      val parsed = value.toInt
+      parsed >= 8 && parsed <= 15
+    }
+
+  private final class InflaterFlow(
+      noContextTakeover: Boolean,
+      settings: WebSocketCompressionSettingsImpl)
+      extends LifecycleMapConcat[FrameEventOrError, FrameEventOrError] {
+    private var inflater = new Inflater(true)
+    private var compressedFrame: Option[CompressedFrame] = None
+    private var compressedMessageInProgress = false
+    private var decompressedMessageBytes = 0L
+    private var bypassFrameInProgress = false
+
+    override def apply(event: FrameEventOrError): immutable.Iterable[FrameEventOrError] = event match {
+      case start @ FrameStart(header, data)
+          if header.rsv1 &&
+          (header.opcode == Protocol.Opcode.Text ||
+          header.opcode == Protocol.Opcode.Binary) =>
+        if (compressedMessageInProgress || compressedFrame.isDefined)
+          throw new ProtocolException("Unexpected data frame while fragmented message is open")
+        if (header.rsv2 || header.rsv3) throw new ProtocolException("Unexpected reserved bit for compressed message")
+        compressedMessageInProgress = !header.fin
+        compressedFrame = Some(CompressedFrame(header.copy(rsv1 = false, length = 0), data, appendTail = header.fin))
+        if (start.lastPart) finishFrame() else Nil
+      case start @ FrameStart(header, _) if bypassFrameInProgress =>
+        throw new ProtocolException(s"Unexpected frame ${header.opcode} while frame data is open")
+      case start @ FrameStart(header, _)
+          if (compressedFrame.isDefined || compressedMessageInProgress) && header.opcode.isControl =>
+        bypassFrameInProgress = !start.lastPart
+        start :: Nil
+      case start @ FrameStart(header, data)
+          if compressedMessageInProgress && header.opcode == Protocol.Opcode.Continuation =>
+        if (header.rsv1 || header.rsv2 || header.rsv3)
+          throw new ProtocolException("Unexpected reserved bit for continuation frame")
+        compressedMessageInProgress = !header.fin
+        compressedFrame = Some(CompressedFrame(header.copy(length = 0), data, appendTail = header.fin))
+        if (start.lastPart) finishFrame() else Nil
+      case start @ FrameStart(header, _) if compressedFrame.isDefined || compressedMessageInProgress =>
+        throw new ProtocolException(s"Unexpected frame ${header.opcode} while fragmented message is open")
+      case data: FrameData if bypassFrameInProgress =>
+        bypassFrameInProgress = !data.lastPart
+        data :: Nil
+      case data: FrameData if compressedFrame.isDefined =>
+        compressedFrame = compressedFrame.map(_.append(data.data))
+        if (data.lastPart) finishFrame() else Nil
+      case other => other :: Nil
+    }
+
+    private def finishFrame(): immutable.Iterable[FrameEventOrError] = {
+      val frame = compressedFrame.get
+      compressedFrame = None
+      val inflated = inflate(frame.data, frame.appendTail)
+      if (frame.appendTail) decompressedMessageBytes = 0L
+      if (frame.appendTail && noContextTakeover) {
+        inflater.end()
+        inflater = new Inflater(true)
+      }
+      FrameStart(frame.header.copy(length = inflated.length), inflated) :: Nil
+    }
+
+    private def inflate(data: ByteString, appendTail: Boolean): ByteString = {
+      try {
+        val input = if (appendTail) data ++ EmptyStoredBlock else data
+        inflater.setInput(input.toArray)
+        val output = new ByteArrayOutputStream()
+        val buffer = new Array[Byte](1024)
+        var count = inflater.inflate(buffer)
+        while (count > 0) {
+          decompressedMessageBytes += count
+          if (settings.maxAllocation > 0 && decompressedMessageBytes > settings.maxAllocation)
+            throw new ProtocolException("WebSocket decompressed message exceeds configured maximum allocation")
+          output.write(buffer, 0, count)
+          count = inflater.inflate(buffer)
+        }
+        ByteString.fromArray(output.toByteArray)
+      } catch {
+        case ex: DataFormatException =>
+          throw new ProtocolException(s"Invalid WebSocket compressed message: ${ex.getMessage}")
+      }
+    }
+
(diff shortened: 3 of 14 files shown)


http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
```diff
@@ -0,0 +1,341 @@
+/*
```
pjfanning, line 1: the source header should be the standard Apache header - see https://github.com/apache/pekko/blob/main/CONTRIBUTING.md#pull-request-requirements - unless this code is largely a copy/paste of 3rd party code (in which case, the existing source header in that code would be used)

http-core/src/main/scala/org/apache/pekko/http/javadsl/model/ws/WebSocketUpgrade.scala
```diff
@@ -41,20 +41,50 @@ trait WebSocketUpgrade {
...
    * of the ones offered by the client.
    */
   def handleMessagesWith(handlerFlow: Graph[FlowShape[Message, Message], ? <: Any], subprotocol: String): HttpResponse
 
+  /**
+   * Returns a response that can be used to answer a WebSocket handshake request. The connection will afterwards
+   * use the given handlerFlow to handle WebSocket messages from the client. The given subprotocol must be one
+   * of the ones offered by the client.
+   *
+   * The compressionEnabled flag allows declining negotiated WebSocket compression for this accepted WebSocket.
```
pjfanning, line 67: can you add `@since 2.0.0` on new public API methods?

docs/src/main/paradox/release-notes/releases-2.0.md
```diff
@@ -15,3 +15,4 @@ by early adopters. This is experimental. This release should not be used in prod
 * Dependency versions have been updated
 * Jackson3 is supported in a new pekko-http-jackson3 lib
 * Changed Java DSL methods that return Scala Durations to return Java Durations ([PR788](https://github.com/apache/pekko-http/pull/788))
+* Server-side WebSocket connections can now negotiate RFC 7692 `permessage-deflate` compression when requested by the client.
```
pjfanning: 2.0.0-M1 is already released. Just revert this and we will form 2.0.0-M2 release notes when that release is ready, hopefully in a month or 2.

He-Pin:
I see netty have many compression support, should we do that too?

http-core/src/main/resources/reference.conf
```diff
@@ -348,6 +348,44 @@ pekko.http {
 
       # Enable verbose debug logging for all ingoing and outgoing frames
       log-frames = false
+
+      compression {
+        # Whether the server should support WebSocket compression using the RFC 7692
+        # permessage-deflate extension. Compression is negotiated during the
+        # WebSocket handshake and is only used when the client requests it.
+        enabled = true
```
He-Pin, line 356: Should we do adaptive compression?

mkurz:
> I see netty have many compression support, should we do that too?

For this PR, no. This PR should stay focused on RFC 7692 `permessage-deflate`, which is the relevant standardized WebSocket compression extension. Netty also has legacy `deflate-frame` / `x-webkit-deflate-frame` support. I intentionally did not add those because they are historical, non-standard extension names from before RFC 7692 standardized `permessage-deflate`. For our new implementation in pekko-http, I think it is better to support the standardized extension first and only add legacy extensions later if users report a concrete compatibility need (which I highly doubt). This avoids increased code, tests, and configuration surface for very little practical benefit.

http-core/src/main/resources/reference.conf
mkurz: Could you clarify what kind of adaptive compression you mean?

With this PR, Pekko HTTP already supports adaptation in two ways:

1. Compression is only negotiated when the client offers `permessage-deflate`.
2. Compression can be disabled globally, or per accepted WebSocket via the new `compressionEnabled` overloads.

If you mean **adapting per outbound message**, for example compressing a large message but sending a small message uncompressed, I think that could make sense as a follow-up PR. `permessage-deflate` allows this: once the extension is negotiated, a sender can still choose to leave RSV1 unset for individual messages. This would need careful handling around fragmented messages and context takeover, but a minimum-size threshold or application-controlled filter could be useful.
In Play we do basically the same for the gzip filter: https://github.com/playframework/playframework/blob/d87116f880789e0dbc40b7f971e08638e5ab7557/web/play-filters-helpers/src/main/resources/reference.conf#L297-L300 - which avoids gziping content that is smaller then a certain threshold.
But, as said, I would do that in a follow up PR if desired.

Similarly, **skipping compression for certain message types or already-compressed payloads** could make sense later, but Pekko HTTP currently does not know application-level message semantics at this layer, so that would also need separate design.

If you mean **changing the compression level dynamically**, that is more complicated and probably less useful as a general automatic feature. A simpler and more realistic variant would be allowing applications to choose a **compression level per WebSocket connection**, similar to how this PR allows enabling/disabling compression per accepted WebSocket. That should be much easier to reason about than changing the level message-by-message.
I would also do that in a follow up PR if desired.

If you mean disabling compression automatically under CPU or memory pressure, I think that is much larger in scope and probably not a good fit for this PR. It would need metrics, policy, and predictable behavior, and I am not sure the added complexity would be worth it.

docs/src/main/paradox/release-notes/releases-2.0.md
mkurz: Removed.

http-core/src/main/scala/org/apache/pekko/http/javadsl/model/ws/WebSocketUpgrade.scala
mkurz: Added.

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
mkurz: Removed. This was just a mistake from my side as many other pekko files also contain the lightbend header (I think I started by copying some other files from this repo, can't remember to be honest).
Anyway, this code is not copied from anywhere. Akka HTTP still does not have websocket compression...

pjfanning: the standard Apache header looks like https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/SparkContext.scala - this one that mentions Akka is only for code that is derived from Akka.

mkurz:
All fixed.

For full disclosure: I was working on this for like 1,5 days intensively with heavy use of codex (gpt-5.5), doing lots of rubberducking, analysing how netty or projects like tomcat implements things and going over the RFC. Did many dozens iterations, back and forth (trying different approaches). Zero code copied, some manual adjustments. Since the netty implementation is different - by having different architecture in general -  and Akka HTTP does not even have support for compressed websocket, this code here is unique ("my own work").

Also, made sure that this will play nicely together with Play - where I locally build support for websocket compression as well based on this implementation. Will submit the Play PR soon so you see what I mean.

http-core/src/main/scala/org/apache/pekko/http/scaladsl/model/ws/WebSocketUpgrade.scala
```diff
@@ -97,13 +158,36 @@ trait WebSocketUpgrade extends jm.ws.WebSocketUpgrade {
...
    */
   def handleMessagesWith(
       inSink: Graph[SinkShape[jm.ws.Message], ? <: Any],
       outSource: Graph[SourceShape[jm.ws.Message], ? <: Any],
       subprotocol: String): HttpResponse =
     handleMessages(createScalaFlow(inSink, outSource), subprotocol = Some(subprotocol))
 
+  /**
+   * Java API
+   */
```
pjfanning, line 185: can you add `@since 2.0.0` here?

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
mkurz: Sorry this was overlooked by me. I hope it works now.

http-core/src/main/scala/org/apache/pekko/http/scaladsl/model/ws/WebSocketUpgrade.scala
mkurz: Done.

pjfanning (approved):
lgtm - thanks

Hopefully, we can make some of the suggested improvements over the next few weeks but I think this is already a very useful addition as is.

http-core/src/main/resources/reference.conf
mkurz: Did some research if netty has similiar feature(s):

What Netty does have is lower-level extension filters:
```java
  WebSocketExtensionFilter.mustSkip(WebSocketFrame frame)
```

Those filters can decide per frame whether an extension encoder/decoder should be skipped. Netty’s own tests show a threshold-style example:

```java
  return (frame instanceof TextWebSocketFrame || frame instanceof BinaryWebSocketFrame)
      && frame.content().readableBytes() < 100;
```

So Netty supports the mechanism for "don’t compress small messages", but it does not ship a named adaptive compression feature/config. Its default is `WebSocketExtensionFilterProvider.DEFAULT`, which means `NEVER_SKIP`, so negotiated messages are compressed.

Important detail: for `permessage-deflate`, Netty forbids changing the decision in the middle of a fragmented message. If compression/decompression is already in progress and the filter suddenly says “skip” for a continuation frame, Netty throws. 

So yeah, maybe in a follow up PR it might make sense to think about some mechanism to dynamicaly decide if an outbound message should be compressed or not

mkurz:
(only rebased on latest main with last force push)

He-Pin (commented):
Nice work on this PR — server-side WebSocket compression is a long-requested feature and the implementation looks solid overall. The negotiation logic is thorough, the test coverage is impressive (fragmentation, context takeover, max-allocation, low-level frames), and the per-upgrade compression control is a nice touch.

A few things I noticed while reading through the code:

### `PerMessageDeflate.scala`

1. **Buffer allocation in `inflate`/`deflate`**: The 1024-byte buffer is allocated on every call to `inflate()`/`deflate()`. Since these are called per-frame within a single connection, it might be worth making the buffer a class-level field in `InflaterFlow`/`DeflaterFlow` to reduce allocation pressure. Also, 1024 bytes is quite small — 8192 is a more typical buffer size for streaming decompression/compression and would reduce loop iterations for larger messages.

2. **`CompressedFrame.append` and `UncompressedFrame.append`**: Using `data ++ next` on ByteString for each incoming fragment creates a new ByteString each time, which can become O(n²) for heavily fragmented messages. Consider collecting fragments in a `Vector[ByteString]` and concatenating once in `finishFrame()`, or using a `ByteStringBuilder`.

3. **`asInstanceOf[WebSocketSettingsImpl]` in `Handshake.scala`**: The unchecked cast `settings.asInstanceOf[WebSocketSettingsImpl].compression` will throw a `ClassCastException` if a custom `WebSocketSettings` implementation is provided. Consider using a pattern match with a fallback (e.g. skip compression negotiation for unknown settings types) or documenting the requirement.

4. **`validWindowBits` edge case**: The function doesn't guard against an empty string explicitly. While the empty-string case for `client_max_window_bits` is handled before `validWindowBits` is called, the function itself would throw `NumberFormatException` on `"".toInt` if ever called with an empty value in the future. A simple `value.nonEmpty &&` guard would make it more defensive.

5. **`LifecycleMapConcatStage` — incomplete message on upstream finish**: If upstream finishes while a fragmented compressed message is still in progress (`compressedMessageInProgress = true` or `compressedFrame.isDefined`), the pending data is silently dropped. Consider whether this should emit a protocol error or at least log a warning.

### Configuration

6. **Default `max-allocation = 64k`**: 64 KB is quite conservative for a default decompressed message size limit. Many real-world WebSocket applications (e.g. real-time collaboration, gaming, large JSON payloads) routinely exceed this. While it's a security default and can be overridden, a value like 1 MB or 256 KB might be more practical and still provide protection against decompression bombs. Tomcat defaults to no limit, and Jetty defaults to 128 KB for the text message size limit.

### Minor / Style

7. **Header list construction in `buildResponse`**: The `List(...) ::: perMessageDeflate.map(...).toList ::: List(...)` pattern is a bit awkward. A `Seq.newBuilder` or a conditional `++` might read more cleanly.

8. **`WebSocketCompressionSettingsImpl.Disabled` sentinel**: The `Disabled` value has `maxAllocation = 0`, which the reference.conf documents as "no limit". While this value is never used when compression is disabled, the semantic mismatch is slightly confusing.

The test coverage is excellent — especially the fragmented compressed messages with interleaved control frames, the context takeover reset tests, and the max-allocation boundary tests. Overall this is a well-structured PR.

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
```diff
@@ -0,0 +1,345 @@
...
+      }
+      FrameStart(frame.header.copy(length = inflated.length), inflated) :: Nil
+    }
+
+    private def inflate(data: ByteString, appendTail: Boolean): ByteString = {
+      try {
+        val input = if (appendTail) data ++ EmptyStoredBlock else data
+        inflater.setInput(input.toArray)
+        val output = new ByteArrayOutputStream()
+        val buffer = new Array[Byte](1024)
```
He-Pin, line 199: The 1024-byte buffer is allocated fresh on every `inflate()` call. Since this method runs per frame within a single connection, it might be worth making this a class-level field to reduce GC pressure. Also, 1024 is pretty small — something like 8192 would cut down loop iterations for larger messages without meaningfully increasing memory usage.

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
```diff
@@ -0,0 +1,345 @@
...
+        deflater.end()
+        deflater = new Deflater(settings.compressionLevel, true)
+      }
+      FrameStart(current.header.copy(length = compressed.length), compressed) :: Nil
+    }
+
+    private def deflate(data: ByteString, removeTail: Boolean): ByteString = {
+      deflater.setInput(data.toArray)
+      val output = new ByteArrayOutputStream()
+      val buffer = new Array[Byte](1024)
```
He-Pin, line 280: Same as above — this buffer could be a field of `DeflaterFlow` and sized to 8192 or so. The deflater loop will spin many more times than necessary on larger payloads with a 1KB buffer.

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
```diff
@@ -0,0 +1,345 @@
...
+          else if (!hasBeenPulled(in)) pull(in)
+
+        setHandler(in, this)
+        setHandler(out, this)
+      }
+  }
+
+  private final case class CompressedFrame(header: FrameHeader, data: ByteString, appendTail: Boolean) {
+    def append(next: ByteString): CompressedFrame = copy(data = data ++ next)
+  }
```
He-Pin, line 340: '`data ++ next` creates a new ByteString copy on every fragment. For heavily fragmented messages this becomes O(n²) in the total payload size. Consider accumulating fragments in a `Vector[ByteString]` (or a `ByteStringBuilder`) and concatenating once in `finishFrame()`. Same applies to `UncompressedFrame.append` below.'

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
```diff
@@ -0,0 +1,345 @@
...
+        accepted = false
+    }
+
+    if (accepted) {
+      Some(Negotiated(WebSocketExtension(ExtensionName, responseParams), serverNoContext, clientNoContext, settings))
+    } else None
+  }
+
+  private def validWindowBits(value: String): Boolean =
+    value.length <= 2 && value.forall(_.isDigit) && {
```
He-Pin, line 132: Nit: if this ever gets called with an empty string (it won't today because the empty case is handled before reaching here), `value.toInt` would throw `NumberFormatException`. A quick `value.nonEmpty &&` guard at the start would make this more defensive for future callers.

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
```diff
@@ -0,0 +1,345 @@
...
+
+        override def onPush(): Unit = {
+          pending = handler(grab(in)).iterator
+          pushOrPull()
+        }
+
+        override def onPull(): Unit =
+          pushOrPull()
+
+        override def onUpstreamFinish(): Unit = {
```
He-Pin, line 320: If upstream finishes while a fragmented compressed message is still in progress (`compressedMessageInProgress == true` or `compressedFrame.isDefined`), the pending data is silently dropped here. Not sure if that's the right behavior — maybe emit a protocol error or at least log a warning so it's not a silent data loss?

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/Handshake.scala
```diff
@@ -122,26 +123,47 @@ private[http] object Handshake {
             case OptionVal.Some(p) => p.protocols
             case _                 => Nil
           }
+          val clientRequestedExtensions = headers.collect {
+            case extensions: `Sec-WebSocket-Extensions` => extensions.extensions
+          }.flatten
+          val perMessageDeflate =
+            PerMessageDeflate.negotiate(
+              clientRequestedExtensions,
+              settings.asInstanceOf[WebSocketSettingsImpl].compression)
```
He-Pin, line 132: This unchecked cast will throw a `ClassCastException` if someone provides a custom `WebSocketSettings` implementation that isn't `WebSocketSettingsImpl`. Might be worth a pattern match with a graceful fallback (e.g. skip compression negotiation for unknown settings types), or at least documenting that `WebSocketSettings` must be a `WebSocketSettingsImpl`.

http-core/src/main/resources/reference.conf
```diff
@@ -348,6 +348,44 @@ pekko.http {
...
+
+        # Maximum size of a decompressed WebSocket message. If this value is
+        # exceeded while inflating a compressed message, the connection is closed
+        # with a WebSocket protocol error.
+        # Set to 0 to disable this limit.
+        max-allocation = 64k
```
He-Pin, line 362: 64 KB feels quite conservative for a default. Real-world WebSocket apps (collaborative editing, gaming, large JSON payloads) routinely exceed this. Tomcat has no limit by default, Jetty uses 128 KB for text message size. Something like 256 KB or 1 MB might be a more practical default while still protecting against decompression bombs. Users who need tighter limits can always lower it.

He-Pin (commented):
Left a few inline suggestions on the core implementation files.

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
mkurz: > The 1024-byte buffer is allocated fresh on every inflate() call. Since this method runs per frame within a single connection, it might be worth making this a class-level field to reduce GC pressure.

codex was suggesting the same, but we came to the conclusion its not worth it. anyway, let me see again.

mkurz:
@He-Pin Give me a minute, so my agent can answer your agent ;) 
btw, which model are you using?

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
He-Pin: in the Artery Transport, there is a bytebufPoll. I think that would help.

He-Pin (approved):
overall lgtm , would it possible to coodinate with https://github.com/apache/pekko/pull/2409 ?

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/Handshake.scala
```diff
@@ -122,26 +123,47 @@ private[http] object Handshake {
             case OptionVal.Some(p) => p.protocols
             case _                 => Nil
           }
+          val clientRequestedExtensions = headers.collect {
+            case extensions: `Sec-WebSocket-Extensions` => extensions.extensions
+          }.flatten
```
He-Pin, line 128: Would it be possible to avoid the flatten and do it in one call?

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
```diff
@@ -0,0 +1,345 @@
...
+  final case class Negotiated(
+      responseExtension: WebSocketExtension,
+      serverNoContextTakeover: Boolean,
+      clientNoContextTakeover: Boolean,
+      settings: WebSocketCompressionSettingsImpl) {
+    def bidiFlow: BidiFlow[FrameEventOrError, FrameEventOrError, FrameEvent, FrameEvent, NotUsed] =
+      BidiFlow.fromFlows(inflaterFlow, deflaterFlow)
+
+    def frameEventBidiFlow(
+        maskRandom: () => Random): BidiFlow[FrameEvent, FrameEvent, FrameEvent, FrameEvent, NotUsed] =
```
He-Pin, line 64: Use RandomGenerator in Java 17?

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
```diff
@@ -0,0 +1,345 @@
...
+
+    override def close(): Unit =
+      deflater.end()
+  }
+
+  private trait LifecycleMapConcat[-In, +Out] extends (In => immutable.Iterable[Out]) {
+    def close(): Unit
+  }
+
+  private final class LifecycleMapConcatStage[In, Out](
```
He-Pin, line 298: is it possible upstream to pekko

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
```diff
@@ -0,0 +1,345 @@
...
+      if (removeTail && bytes.endsWith(EmptyStoredBlock)) bytes.dropRight(EmptyStoredBlock.length) else bytes
+    }
+
+    override def close(): Unit =
+      deflater.end()
+  }
+
+  private trait LifecycleMapConcat[-In, +Out] extends (In => immutable.Iterable[Out]) {
+    def close(): Unit
+  }
```
He-Pin, line 296: I think we should avoid this `MapConcat` name because there is one in the pekko stream too.

He-Pin: And there is a `statefulMap` which can be used

http-core/src/main/scala/org/apache/pekko/http/scaladsl/model/ws/WebSocketUpgrade.scala
```diff
@@ -81,13 +121,34 @@ trait WebSocketUpgrade extends jm.ws.WebSocketUpgrade {
...
   def handleMessagesWith(
       handlerFlow: Graph[FlowShape[jm.ws.Message, jm.ws.Message], ? <: Any], subprotocol: String): HttpResponse =
     handleMessages(JavaMapping.toScala(handlerFlow), subprotocol = Some(subprotocol))
 
+  /**
+   * Java API
+   *
+   * @since 2.0.0
+   */
+  def handleMessagesWith(
```
He-Pin, line 146: java doc

http-core/src/main/resources/reference.conf
mkurz: I would be  ok with 256 KB - 1 MB seems a bit too much. 
@pjfanning what do you think?

http-core/src/main/resources/reference.conf
He-Pin: yes, but that can come up later, we are using something like this later.

pjfanning:
> overall lgtm , would it possible to coodinate with [apache/pekko#2409](https://github.com/apache/pekko/pull/2409) ?

2409 is not making progress and we can always add zstd later - I don't see why we can't just use the compression here initially

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
pjfanning: * unless there is an immediate use case for this in pekko core, this should be left here
* even if we do end up moving this to pekko core, I would prefer not to block this PR - we can always delete this private code from here and use the pekko core version later on

mkurz: agree. resolving.

http-core/src/main/resources/reference.conf
pjfanning: I have no strong preference. 256Kb can be the starting point. If we get requests to increase the default later on, we can do that. The value is a configurable but a default setting that only some people need to worry about is ideal.

He-Pin:
I'm fine with this, Which I'm asking is will @mkurz introduce some common interface into pekko and used it here.

mkurz:
I will address the comments later or tomorrow

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
```diff
@@ -0,0 +1,345 @@
...
+        val buffer = new Array[Byte](1024)
+        var count = inflater.inflate(buffer)
+        while (count > 0) {
+          decompressedMessageBytes += count
+          if (settings.maxAllocation > 0 && decompressedMessageBytes > settings.maxAllocation)
+            throw new ProtocolException("WebSocket decompressed message exceeds configured maximum allocation")
+          output.write(buffer, 0, count)
+          count = inflater.inflate(buffer)
+        }
+        ByteString.fromArray(output.toByteArray)
```
He-Pin, line 208: There is an fromUnsafeArray iirc

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
```diff
@@ -0,0 +1,345 @@
...
+        bypassFrameInProgress = !data.lastPart
+        data :: Nil
+      case data: FrameData if compressedFrame.isDefined =>
+        compressedFrame = compressedFrame.map(_.append(data.data))
+        if (data.lastPart) finishFrame() else Nil
+      case other => other :: Nil
+    }
+
+    private def finishFrame(): immutable.Iterable[FrameEventOrError] = {
+      val frame = compressedFrame.get
```
He-Pin, line 183: Netty (DeflateDecoder.java:48-49, 122-123) handles empty EMPTY_DEFLATE_BLOCK

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
He-Pin, line 199: is it possible we reuse this buffer?

He-Pin:
I think we can merge this, and improvement can come up later, wdyt @pjfanning I checked the netty's implemenation, the current is great and correct. but yes, netty can have better performance.

pjfanning:
merged the existing PR - we can mark this as experimental in the next release if there are still issues that need addressing at that point

mkurz:
- https://github.com/playframework/playframework/pull/14075

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
mkurz: Addressed by #1173. It uses one reusable 8 KiB byte array owned by each materialized InflaterFlow, rather than a shared pool. Since each flow is per connection and processed serially, this removes the per-call allocation without adding pool acquire/release overhead.

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
mkurz: Addressed by #1173: DeflaterFlow now owns and reuses an 8 KiB buffer instead of allocating a new 1 KiB buffer for every deflate() call.

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
mkurz: Addressed by #1173: fragments are accumulated in a `Vector[ByteString]` and compacted once using a size-hinted `ByteStringBuilder`. The single-fragment case returns the existing ByteString directly.

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
mkurz: Addressed by #1173 by adding the defensive `value.nonEmpty` check before parsing the window-bits value.

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
mkurz: #1173 now verifies that inflater and deflater resources are released even when the stream finishes with incomplete message state. It intentionally leaves the resulting wire and stream semantics unchanged. I plan to address those separately: incomplete inbound compressed state should produce a FrameError and WebSocket close 1002, while incomplete outbound state should fail the stream stage.

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/Handshake.scala
mkurz: Addressed by #1173: the handshake now pattern-matches `WebSocketSettingsImpl`. Unknown custom implementations skip compression negotiation instead of causing a `ClassCastException`.

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/Handshake.scala
mkurz: Addressed by #1173 by replacing `collect(...).flatten` with a single `flatMap`.

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
mkurz: Yes. Addressed by #1173: the buffer is now an 8 KiB field of InflaterFlow and is reused across inflate operations.

http-core/src/main/resources/reference.conf
mkurz: I have now opened follow-up #1172, which adds application-controlled per-message compression selection. The compression decision is evaluated once per message and retained across continuation frames, including when control frames are interleaved.

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
mkurz: This follows the existing WebSocket masking API: `WebSocketSettings.randomFactory` and `Masking` currently use `java.util.Random`. Using `RandomGenerator` consistently would require a broader `WebSocket` settings/masking API change rather than a compression-specific change, so I would leave that for a separate PR.

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
mkurz: I looked at both built-in operators. `statefulMapConcat` only invokes its completion callback on normal upstream completion, so it cannot guarantee that `Inflater.end()` and `Deflater.end()` run after upstream failure or downstream cancellation. `statefulMap` has broader lifecycle handling but is one-to-one, so we would need to emit an `Iterable` and add another flattening stage.

The private `LifecycleMapConcatStage` directly supports zero or more outputs and calls `close()` from `postStop`, covering completion, failure, cancellation, and protocol errors. #1173 adds tests for all of those termination paths. I therefore kept the custom stage; the Lifecycle prefix is intended to distinguish it from Pekko Stream’s regular mapConcat operator.

http-core/src/main/scala/org/apache/pekko/http/scaladsl/model/ws/WebSocketUpgrade.scala
mkurz: The complete method documentation is present on `org.apache.pekko.http.javadsl.model.ws.WebSocketUpgrade`. The corresponding methods in the Scala DSL trait are bridge implementations of that Java API and follow the existing convention of using the Java API marker. The new public methods also have `@since 2.0.0`.

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
mkurz: Addressed in the merged implementation: the freshly created output arrays are wrapped using `ByteString.fromArrayUnsafe`. They are not modified after ownership is transferred to the `ByteString`. Input uses `toArrayUnsafe` before being passed to the JDK inflater or deflater, which only reads that input.

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
mkurz: Addressed in the merged implementation and covered by explicit tests. An empty inbound compressed message is inflated and delivered as an empty message, while an empty outbound message produces the expected DEFLATE block and successfully round-trips through an inflater. Final compressed frames also have the RFC 7692 tail restored before inflation.

http-core/src/main/resources/reference.conf
mkurz: Increased to 256kb in https://github.com/apache/pekko-http/pull/1173

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
mkurz: Correction to my previous reply: after comparing this more closely with Netty and general WebSocket shutdown semantics, I no longer plan a compression-specific behavioral change.

Netty releases incomplete encoder and decoder state during handler removal or channel inactivity without inspecting the in-progress compression state or synthesizing a 1002 close. A Close frame may also terminate a connection while a fragmented data message is unfinished, and after a real TCP EOF sending another Close frame is generally impossible.

#1173 addresses the actionable part by verifying that inflater and deflater resources are released for incomplete inbound and outbound state, including incomplete logical messages and physical frames.

Any future change to truncated-message reporting should be implemented in the general WebSocket framing layer and cover compressed and uncompressed fragmented messages consistently.

PR #1122, opened 2026-07-05
Remove deprecated StatusCodes.PayloadTooLarge and UnprocessableEntity (since 1.1.0)

He-Pin:
## Motivation
`PayloadTooLarge` and `UnprocessableEntity` status codes were deprecated since pekko-http 1.1.0 in favor of `ContentTooLarge` and `UnprocessableContent` respectively.

## Modification
Removed deprecated vals from Scala `StatusCodes` object and Java `StatusCodes` class. Added MiMa filters for binary compatibility exceptions.

## Result
Deprecated API removed. Users should use `ContentTooLarge` and `UnprocessableContent`.


pjfanning:
I, personally, prefer to keep deprecated methods around for a while and these were only deprecated recently.

PR #1123, opened 2026-07-05
Remove deprecated HttpMethod.custom 4-arg overload (since 1.4.0)

He-Pin:
## Motivation
The 4-arg overload of `HttpMethod.custom` was deprecated since pekko-http 1.4.0 in favor of the 5-arg overload with `contentLengthAllowed` parameter.

## Modification
Removed `HttpMethod.custom(name, safe, idempotent, requestEntityAcceptance)` from Scala and Java APIs. Removed private `oldContentLengthCondition` helper. Added MiMa filters.

## Result
Deprecated API removed. Users should use the 5-arg `custom` overload.


pjfanning:
I, personally, prefer to keep deprecated methods around for a while and these were only deprecated recently. 1.4.0 has not been released yet.

He-Pin:
Ok ,let's keep it for the next major release

PR #1126, opened 2026-07-05
Remove deprecated ParserSettings.apply/create (since Akka HTTP 10.2.0)

He-Pin:
## Motivation
`ParserSettings.apply` and `ParserSettings.create` were deprecated since Akka HTTP 10.2.0 in favor of `forServer` and `forClient`.

## Modification
Removed deprecated `apply` methods from Scala API and `create` methods from Java API. Updated test to use `forServer()`. Added MiMa filters.

## Result
Deprecated API removed. Users should use `ParserSettings.forServer()` or `ParserSettings.forClient()`.


pjfanning (approved):
lgtm

PR #1127, opened 2026-07-05
Remove deprecated ConnectionContext.defaultPort (since Akka HTTP 10.2.0)

He-Pin:
## Motivation
`ConnectionContext.defaultPort` was deprecated since Akka HTTP 10.2.0 as an internal method left only for binary compatibility.

## Modification
Removed `defaultPort` from `ConnectionContext` trait and its overrides in `HttpsConnectionContext` and `HttpConnectionContext`. Added MiMa filters.

## Result
Deprecated internal method removed.


pjfanning (approved):
lgtm

PR #1132, opened 2026-07-06
Replace Reflections with ClassGraph in tests

He-Pin:
### Motivation
The Reflections-based test scanners rely on deep classpath reflection. ClassGraph provides the same test metadata scanning without retaining the broader MethodHandle/VarHandle refactoring that reviewers called out as unnecessary complexity.

### Modification
- Replace `org.reflections` with `io.github.classgraph:classgraph:4.8.184` for docs and HTTP test scanners.
- Update `ApiMayChangeDocCheckerSpec` to find `@ApiMayChange` classes and methods through ClassGraph metadata.
- Update `DirectivesConsistencySpec` to read directive methods and `@CorrespondsTo` metadata through ClassGraph while preserving direct-interface lookup semantics.
- Close ClassGraph `ScanResult` instances after each ScalaTest suite.

### Result
The PR is now scoped to replacing Reflections with ClassGraph in test scanners, while keeping unrelated reflection code in its simpler existing form.

### Tests
- `scalafmt --mode diff-ref=b0d3009b375b536397a776880301cc1943ef4645` - success
- `scalafmt --list --mode diff-ref=b0d3009b375b536397a776880301cc1943ef4645` - success
- `git diff --check` - success
- JDK 17: `sbt "docs / Test / testOnly docs.ApiMayChangeDocCheckerSpec" "http-tests / Test / testOnly org.apache.pekko.http.javadsl.DirectivesConsistencySpec"` - success
- `qodercli -p --output-format stream-json --cwd "$PWD" --attachment /tmp/project-review.diff ...` - No must-fix findings

### References
None - internal test dependency migration

### DIFF
--- a/docs/src/test/scala/docs/ApiMayChangeDocCheckerSpec.scala
+++ b/docs/src/test/scala/docs/ApiMayChangeDocCheckerSpec.scala
@@ -13,15 +13,10 @@
 
 package docs
 
-import java.lang.reflect.Method
-
 import org.apache.pekko.annotation.ApiMayChange
-import org.reflections.Reflections
-import org.reflections.scanners.{ MethodAnnotationsScanner, Scanners, TypeAnnotationsScanner }
-import org.reflections.util.{ ClasspathHelper, ConfigurationBuilder }
+import io.github.classgraph.{ ClassGraph, MethodInfo }
 import org.scalatest.Assertion
 
-import scala.collection.mutable
 import scala.io.Source
 import scala.jdk.CollectionConverters._
 import org.scalatest.matchers.should.Matchers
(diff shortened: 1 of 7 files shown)


jrudolph:
As long as there's no clear performance win (which is already arguably irrelevant for tests), I'd keep the simpler existing versions.

In general, a change like that *can* help in hot loops or wherever inlining can help with further optimizations but in many cases, there are other inlining hazards so that there's no real benefit of complicating the code for that.

He-Pin:
In this repo we can replace it with ClassGraph, because the newer JDK will report an error for deep reflection.

pjfanning (approved):
lgtm - code changes are easy enough to follow and the lib change only affects tests

He-Pin:
Yes, I want to reduce the reflection usage .

Issue #1133, opened 2026-07-06
OOM issue with zgc


teads-lionel-ngo:
Hello there, 

I am considering switching the gc used in production to zgc for my app but it constantly crashed with a OOM error. I have been actively chasing the root cause of the OOM and I believe there may be some issues when using pekko http with some garbage collectors. 
To make it short, ZGC doesn't reclaim off-heap memory because my app has a fairly large heap with very moderate pressure. My app uses `decodeRequest` & `encodeResponse`. 
For example, `decodeRequest` uses a [GZipCompressor](https://github.com/apache/pekko-http/blob/fad7678899cf1ed36cc4921ecf0d421a69f980ae/http/src/main/scala/org/apache/pekko/http/scaladsl/coding/GzipCompressor.scala#L71) and from my understanding for every incoming request, it creates a new [Inflater](https://github.com/apache/pekko-http/blob/fad7678899cf1ed36cc4921ecf0d421a69f980ae/http/src/main/scala/org/apache/pekko/http/scaladsl/coding/GzipCompressor.scala#L74). The issue for my use case is that there is no explicit call `.end()` to properly release the resource leaving it to the garbage collector to reclaim it. Same goes for `encoreResponse`.
One workaround I found is using the flag `ZCollectionInterval` to force the gc to release off-heap memory but at the expense of CPU. I also tried to do a custom hacky `decoderRequest` and it also worked.

So here's my question: is it an intended behaviour in pekko http to let the gc handle this?

He-Pin:
I’ve also run into OOM issues when using ZGC in production; it requires a large memory footprint, and in some cases, we need to be more proactive about explicitly setting references to `null`. We encountered similar issues in our code previously, but since we started using the Alibaba JDK—which utilizes jemalloc—we haven't run into this problem again.

He-Pin:
Another problem is that the default glibc is bad, and we should use jemalloc for the JVM. The best we can do on our side is to do some pooling I think, you are welcome to submit a PR.

jrudolph:
Ultimately, it would be better to move to a Java-native gzip implementation. That would also allow negotiation for the websocket compression params.

pjfanning:
@teads-lionel-ngo could you provide your workaround, the decodeRequest?

teads-lionel-ngo:
@pjfanning I asked claude to implement a custom hacky `decodeRequest` with a pool and explicit closure of the inflater. Code is AI generated so I wouldn't blindly trust it.

[...]

PR #1134, opened 2026-07-06
GzipDecompressor: fix inflater memory leak

pjfanning:
See #1133 

`GzipDecompressor` allocates a new `Inflater` per gzip stream but did not explicitly release its native memory. Under collectors with relaxed off-heap reclamation such as ZGC, repeated request decompression can accumulate unreclaimed native buffers and eventually OOM.

- **Inflater lifecycle cleanup**
  - Add explicit `Inflater.end()` cleanup for `GzipDecompressor`
  - Make cleanup idempotent so termination paths can safely converge
  - Run cleanup from stage shutdown (`postStop`) to cover normal completion, failure, and cancellation

- **Truncation handling**
  - Release the inflater before failing truncated gzip streams
  - Preserve the existing `Truncated GZIP stream` behavior while ensuring native resources are not left behind

- **Targeted coverage**
  - Add focused tests that verify the inflater is released on:
    - successful decode completion
    - early downstream cancellation
    - truncation failure

```scala
private def cleanupInflater(): Unit =
  if (!inflaterEnded) {
    inflaterEnded = true
    inflater.end()
  }

override def postStop(): Unit = cleanupInflater()

override def onTruncation(): Unit = {
  cleanupInflater()
  failStage(new ZipException("Truncated GZIP stream"))
}
```

### DIFF
--- a/http/src/main/scala/org/apache/pekko/http/scaladsl/coding/GzipCompressor.scala
+++ b/http/src/main/scala/org/apache/pekko/http/scaladsl/coding/GzipCompressor.scala
@@ -70,12 +70,26 @@ private[coding] object GzipCompressor {
 @InternalApi
 private[coding] class GzipDecompressor(
     maxBytesPerChunk: Int = Decoder.MaxBytesPerChunkDefault) extends DeflateDecompressorBase(maxBytesPerChunk) {
+  protected[coding] def createInflater(): Inflater = new Inflater(true)
+
   override def createLogic(attr: Attributes) = new ParsingLogic {
-    private val inflater = new Inflater(true)
-    private val crc32: CRC32 = new CRC32
+    private[this] val inflater = createInflater()
+    private[this] val crc32: CRC32 = new CRC32
+    private[this] var inflaterEnded = false
+
+    private def cleanupInflater(): Unit =
+      if (!inflaterEnded) {
+        inflaterEnded = true
+        inflater.end()
+      }
+
+    override def postStop(): Unit = cleanupInflater()
 
     trait Step extends ParseStep[ByteString] {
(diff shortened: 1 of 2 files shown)


http/src/main/scala/org/apache/pekko/http/scaladsl/coding/GzipCompressor.scala
```diff
@@ -70,12 +70,26 @@ private[coding] object GzipCompressor {
...
   override def createLogic(attr: Attributes) = new ParsingLogic {
-    private val inflater = new Inflater(true)
-    private val crc32: CRC32 = new CRC32
+    private[this] val inflater = createInflater()
+    private[this] val crc32: CRC32 = new CRC32
+    private[this] var inflaterEnded = false
+
+    private def cleanupInflater(): Unit =
+      if (!inflaterEnded) {
+        inflaterEnded = true
```
He-Pin, line 82: Maybe we can mark it as null ? instead of an inflaterEnded

pjfanning: I prefer the boolean to making `inflater` a var and then checking for nulls. The nulls approach makes cleanup harder.

```
    private def cleanupInflater(): Unit =
      if (!inflaterEnded) {
        inflaterEnded = true
        inflater.end()
      }
```

This is not thread-safe.
```
    private def cleanupInflater(): Unit =
      if (inflater != null) {
        inflater.end()
      }
```

This is thread-safe but introduces a local variable.
```
    private def cleanupInflater(): Unit = {
      val localInflater = inflater
      if (localInflater != null) {
        localInflater.end()
      }
   }
```

pjfanning:
@He-Pin do you think is ok to merge?

He-Pin (approved):
lgtm

PR #1135, opened 2026-07-06
use ByteString toArrayUnsafe in new deflater code

pjfanning:
Code looks like it should be ok to use toArrayUnsafe, that no other code should be able to modify the array data.

toArrayUnsafe() on ByteString is only really helpful for ByteStrings directly backed by 1 byte array with no offset. Otherwise, it gives you the same as toArray.

I don't know if the costs outweigh the benefits but if used an InflaterInputStream, we can get an InputStream from a ByteString that can avoid array copies for all ByteString types - but with extra expense to manage the stream, etc.

`new InflaterInputStream(byteString.asInputStream)`

### DIFF
--- a/http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
@@ -195,7 +195,7 @@ private[http] object PerMessageDeflate {
       try {
         val input = if (appendTail) data ++ EmptyStoredBlock else data
         inflater.setInput(input.toArrayUnsafe())
-        val output = new ByteArrayOutputStream()
+        val output = new ByteArrayOutputStream(1024)
         val buffer = new Array[Byte](1024)
         var count = inflater.inflate(buffer)
         while (count > 0) {
@@ -205,7 +205,7 @@ private[http] object PerMessageDeflate {
           output.write(buffer, 0, count)
           count = inflater.inflate(buffer)
         }
-        ByteString.fromArray(output.toByteArray)
+        ByteString.fromArrayUnsafe(output.toByteArray)
       } catch {
         case ex: DataFormatException =>
           throw new ProtocolException(s"Invalid WebSocket compressed message: ${ex.getMessage}")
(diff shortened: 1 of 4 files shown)


pjfanning:
@mkurz does this look ok?

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
```diff
@@ -194,7 +194,7 @@ private[http] object PerMessageDeflate {
     private def inflate(data: ByteString, appendTail: Boolean): ByteString = {
       try {
         val input = if (appendTail) data ++ EmptyStoredBlock else data
-        inflater.setInput(input.toArray)
+        inflater.setInput(input.toArrayUnsafe())
```
He-Pin, line 197: can we use the asInputStream here

pjfanning: would need something like `new InflaterInputStream(byteString.asInputStream)`

He-Pin (approved):

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
pjfanning: actually it's CRC32 that is what uses the bytes from the ByteString not the Inflater directly

http-core/src/main/scala/org/apache/pekko/http/impl/engine/ws/PerMessageDeflate.scala
```diff
@@ -275,15 +275,15 @@ private[http] object PerMessageDeflate {
     }
 
     private def deflate(data: ByteString, removeTail: Boolean): ByteString = {
-      deflater.setInput(data.toArray)
-      val output = new ByteArrayOutputStream()
+      deflater.setInput(data.toArrayUnsafe())
+      val output = new ByteArrayOutputStream(1024)
```
pjfanning, line 279: ByteArrayOutputStream starts with 32 bytes by default and you can end up allocating multiple new arrays while you grow the data. Starting with a reasonable value can help.

PR #1137, opened 2026-07-06
DeflateDecompressor: fix inflater memory leak

pjfanning:
Motivation:
DeflateDecompressor allocates an Inflater per stream but never explicitly releases its native memory. Under GCs with relaxed off-heap reclamation (e.g. ZGC), repeated request decompression can accumulate unreclaimed native buffers and eventually OOM.

Modification:
- Add createInflater(noWrap) factory method to DeflateDecompressor to allow test injection of a custom Inflater
- Track the current inflater in a var within createLogic
- Add idempotent cleanupInflater() that calls inflater.end()
- Override postStop() to call cleanupInflater(), covering normal completion, failure, and cancellation
- Add focused tests in DeflateSpec verifying the inflater is released on successful decode, early cancellation, and truncation

Result:
The Inflater's native memory is reclaimed promptly when the stage terminates, regardless of termination path.

References:
Refs #1134 and #1133

### DIFF
--- a/http-tests/src/test/scala/org/apache/pekko/http/scaladsl/coding/DeflateSpec.scala
+++ b/http-tests/src/test/scala/org/apache/pekko/http/scaladsl/coding/DeflateSpec.scala
@@ -89,7 +89,6 @@ class DeflateSpec extends CoderSpec {
     }
   }
 
-  @nowarn("msg=deprecated")
   private def decodeWith(inflater: TrackingInflater, bytes: ByteString): ByteString =
     decoderWith(inflater).decode(bytes)(SystemMaterializer(system).materializer).awaitResult(3.seconds.dilated)
 
--- a/http-tests/src/test/scala/org/apache/pekko/http/scaladsl/coding/DeflateSpec.scala
+++ b/http-tests/src/test/scala/org/apache/pekko/http/scaladsl/coding/DeflateSpec.scala
@@ -14,6 +14,7 @@
 package org.apache.pekko.http.scaladsl.coding
 
 import java.io.{ InputStream, OutputStream }
+import java.util.concurrent.{ CountDownLatch, TimeUnit }
 import java.util.concurrent.atomic.AtomicInteger
 import java.util.zip._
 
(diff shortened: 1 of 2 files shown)


http/src/main/scala/org/apache/pekko/http/scaladsl/coding/DeflateCompressor.scala
```diff
@@ -103,14 +103,27 @@ private[coding] object DeflateCompressor {
...
+
+    override def postStop(): Unit = cleanupInflater()
 
     /** Step that probes if the deflate stream contains a zlib wrapper */
     case object ProbeWrapping extends ParseStep[ByteString] {
       override def onTruncation(): Unit = completeStage()
 
       override def parse(reader: ByteStringParser.ByteReader): ParseResult[ByteString] = {
         val inflater = examineAndBuildInflater(reader.remainingData)
+        currentInflater = inflater
```
He-Pin, line 127: Can we close the previous inflater before replacing it here? Inflate loops back to ProbeWrapping after inflater.finished(), so concatenated deflate streams can run this path more than once. As written, the second assignment overwrites the first inflater and postStop can only end() the last one. A small helper that calls the idempotent cleanup before assignment, then resets the guard for the new inflater, would keep the leak fix complete.

http-tests/src/test/scala/org/apache/pekko/http/scaladsl/coding/DeflateSpec.scala
```diff
@@ -61,6 +64,54 @@ class DeflateSpec extends CoderSpec {
...
+    }
+    "release the inflater when decoding is truncated" in {
+      val inflater = new TrackingInflater
+      // Truncated deflate streams complete without exception (completeStage on truncation)
+      decodeWith(inflater, streamEncode(smallTextBytes).dropRight(5))
+      inflater.endCalls.get() shouldEqual 1
+    }
+  }
+
+  @nowarn("msg=deprecated")
```
He-Pin: This currently breaks the Scala 2.13 CI: @nowarn annotation does not suppress any warnings at this line. It looks like the deprecated StreamDecoder warning is only introduced by decoderWith, so this annotation should be dropped or moved to the expression that actually needs it.

He-Pin (changes requested):
Thanks for the fix. I found two things that should be fixed before merge: one is the Scala 2.13 compile failure, and the other is a lifecycle hole when the parser loops back for another deflate stream.

He-Pin (approved):

PR #1142, opened 2026-07-07
Ensure deflater/inflater end() is called in DeflateCompressor and GzipCompressor

pjfanning:
more changes related to #1133

- DeflateCompressor: add deflaterEnded flag + idempotent endDeflater() method; call endDeflater() in finishWithBuffer() instead of deflater.end() directly
- GzipDecompressor: add postStop() calling inflater.end() + add createInflater() factory method for testability
- StreamUtils.byteStringTransformer: add optional cleanup callback, call it in postStop() when the stage is stopped before onUpstreamFinish()
- Encoder.singleUseEncoderFlow: pass cleanup callback that calls endDeflater() on DeflateCompressor instances to handle stream cancellation/failure
- DeflateSpec: add tests for deflater cleanup on normal finish and cancellation
- GzipSpec: add tests for inflater cleanup on normal finish and cancellation, and deflater cleanup on normal finish and cancellation

### DIFF
--- a/http-core/src/main/scala/org/apache/pekko/http/impl/util/StreamUtils.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/impl/util/StreamUtils.scala
@@ -43,12 +43,18 @@ private[http] object StreamUtils {
    * Creates a transformer that will call `f` for each incoming ByteString and output its result. After the complete
    * input has been read it will call `finish` once to determine the final ByteString to post to the output.
    * Empty ByteStrings are discarded.
+   * If the stage is stopped before the input is fully consumed (e.g. on downstream cancellation or upstream failure),
+   * `cleanup` is called to release any resources held by the transformer.
    */
   def byteStringTransformer(
-      f: ByteString => ByteString, finish: () => ByteString): GraphStage[FlowShape[ByteString, ByteString]] =
+      f: ByteString => ByteString,
+      finish: () => ByteString,
+      cleanup: () => Unit = () => ()): GraphStage[FlowShape[ByteString, ByteString]] =
     new SimpleLinearGraphStage[ByteString] {
       override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
         new GraphStageLogic(shape) with InHandler with OutHandler {
+          private var finished = false
+
           override def onPush(): Unit = {
             val data = f(grab(in))
             if (data.nonEmpty) push(out, data)
@@ -58,10 +64,13 @@ private[http] object StreamUtils {
           override def onPull(): Unit = pull(in)
 
           override def onUpstreamFinish(): Unit = {
+            finished = true
             val data = finish()
             if (data.nonEmpty) emit(out, data)
             completeStage()
           }
+
+          override def postStop(): Unit = if (!finished) cleanup()
 
           setHandlers(in, out, this)
         }
--- a/http-core/src/main/scala/org/apache/pekko/http/impl/util/StreamUtils.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/impl/util/StreamUtils.scala
@@ -64,10 +64,13 @@ private[http] object StreamUtils {
           override def onPull(): Unit = pull(in)
 
           override def onUpstreamFinish(): Unit = {
-            finished = true
-            val data = finish()
-            if (data.nonEmpty) emit(out, data)
-            completeStage()
+            try {
+              val data = finish()
+              if (data.nonEmpty) emit(out, data)
+              completeStage()
+            } finally {
+              finished = true
+            }
           }
 
           override def postStop(): Unit = if (!finished) cleanup()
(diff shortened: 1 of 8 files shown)


He-Pin (commented):
Overall this looks like a solid cleanup. The idempotent `endDeflater()` pattern is clean, and the tests actually exercise the cancellation path properly (nice touch with `awaitEnd` to avoid the postStop race).

A couple of minor thoughts inline. Also — the PR title still has "WIP" in it, might want to update that before merging.

http-core/src/main/scala/org/apache/pekko/http/impl/util/StreamUtils.scala
```diff
@@ -58,11 +64,14 @@ private[http] object StreamUtils {
           override def onPull(): Unit = pull(in)
 
           override def onUpstreamFinish(): Unit = {
+            finished = true
```
He-Pin: Setting `finished = true` *before* `finish()` runs means that if `finish()` itself throws, `postStop()` won't call `cleanup()`. In the current compressor use case this is harmless because `finishWithBuffer()` already called `endDeflater()` before anything could throw, and `cleanup()` would be a no-op anyway. But since `byteStringTransformer` is a general-purpose helper, someone might later pass a `cleanup` that does more than guard an idempotent resource release — and then this becomes a subtle leak.

Might be worth either moving `finished = true` to *after* the `finish()` call (so `postStop` still runs cleanup on failure), or wrapping in try-finally. Low priority though given the current callers.

http/src/main/scala/org/apache/pekko/http/scaladsl/coding/DeflateCompressor.scala
```diff
@@ -60,10 +60,20 @@ private[coding] class DeflateCompressor private[coding] (compressionLevel: Int)
...
 
+  private[coding] def endDeflater(): Unit =
+    if (!deflaterEnded) {
+      deflaterEnded = true
+      deflater.end()
+    }
+
+  private[coding] override def cleanup(): Unit = endDeflater()
+
+  private var deflaterEnded = false
```
He-Pin: nit: `deflaterEnded` is declared after the methods that reference it (lines 67-73). Works fine in Scala of course, but reads a bit bottom-up. Might be cleaner to declare it next to `deflater` on line 35 since they're logically paired.

pjfanning: I moved the declaration

http-core/src/main/scala/org/apache/pekko/http/impl/util/StreamUtils.scala
pjfanning: I added a try with finally.

http-core/src/main/scala/org/apache/pekko/http/impl/util/StreamUtils.scala
```diff
@@ -58,11 +64,17 @@ private[http] object StreamUtils {
...
           override def onUpstreamFinish(): Unit = {
-            val data = finish()
-            if (data.nonEmpty) emit(out, data)
-            completeStage()
+            try {
+              val data = finish()
+              if (data.nonEmpty) emit(out, data)
+              completeStage()
+            } finally {
+              finished = true
```
He-Pin: Setting the finished flag in the finally block prevents postStop() from calling cleanup() if finish() throws. In that failure path the stage still stops, but the finished guard is already true, so the deflater leak this PR is trying to fix can still happen. Please move the flag assignment onto the success path, right before completeStage(). That keeps the happy path idempotent while still letting postStop() clean up on exceptions.

pjfanning: I moved it before the `if (data.nonEmpty) emit(out, data)` - is that ok?

http-core/src/main/scala/org/apache/pekko/http/impl/util/StreamUtils.scala
```diff
@@ -64,13 +64,10 @@ private[http] object StreamUtils {
...
-            try {
-              val data = finish()
-              if (data.nonEmpty) emit(out, data)
-              completeStage()
-            } finally {
-              finished = true
-            }
+            val data = finish()
+            finished = true
+            if (data.nonEmpty) emit(out, data)
```
He-Pin, line 69: then we should add the `finished = true` to the emit(out, data, () => finished)` wdyt

pjfanning: @He-Pin I don't think we should refactor emit to take an extra function param. I think that would be too large a change.

pjfanning:
@He-Pin is it ok to merge this as is? - as it improves the happy path and that is the leak that is affecting the user in #1133

He-Pin (approved):
lgtm, and I will take another look if I can help any on this.

PR #1154, opened 2026-07-10
Ensure byte string transformers clean up on every stage stop

He-Pin:
### Motivation
PR #1142 added cleanup for byte string transformers, but the stage marked itself finished as soon as finish returned. If the final ByteString was waiting for downstream demand and downstream cancelled, postStop observed the finished flag and skipped cleanup.

The current Deflate and Gzip finish path already closes its Deflater before returning, so this follow-up hardens the generic stage lifecycle rather than claiming a second normal-path native leak.

### Modification
Remove the finished flag and invoke the cleanup callback unconditionally from postStop, making cleanup the single finalizer for normal completion, failure, cancellation, and materializer shutdown.

Add directional StreamUtils tests for normal completion and for cancellation while a final emission is pending.

### Result
Transformer resources are finalized on every stage termination path. The only production cleanup callback is idempotent, so normal completion still calls Deflater.end exactly once.

### References
Refs #1133

Refs #1142


pjfanning (approved):
lgtm

pjfanning:
@He-Pin can you back port this to 1.4.x branch?

PR #1155, opened 2026-07-11
Backport: Fix byte string transformer cleanup on stage stop (1.4.x)

He-Pin:
### Motivation
Backport of #1154 to the 1.4.x branch.

On the byte-string transformer, `finished` was set before the pending final emission was delivered. If downstream cancelled in that window, `postStop` skipped `cleanup`, leaking transformer resources.

### Modification
Cherry-pick of commit 18796ca69492bcb84b98b52005161fe18c4a055c from `main`:

- Treat `cleanup` as the stage finalizer by invoking it unconditionally from `postStop`.
- Drop the fragile `finished` flag.
- Add directional tests for normal completion and for cancellation while the final emission is pending.

Existing compressor cleanup is idempotent, so the unconditional call is safe.

### Result
All stage termination paths release transformer resources on 1.4.x without relying on a fragile `finished` flag.

### Tests
- Delegated to CI for the 1.4.x branch (mirrors validation already done on `main` in #1154).
- On `main`: `sbt "http-core / Test / testOnly org.apache.pekko.http.impl.util.StreamUtilsSpec"` (7 passed), `sbt "http-tests / Test / testOnly org.apache.pekko.http.scaladsl.coding.DeflateSpec org.apache.pekko.http.scaladsl.coding.GzipSpec"` (54 passed), `sbt checkCodeStyle`, `sbt headerCreateAll`/`+headerCheckAll`, and `scalafmt --list --mode diff-ref=origin/main` all passed.

### References
- Backport of #1154
- Refs #1133; follow-up to #1142

### DIFF
--- a/http-core/src/main/scala/org/apache/pekko/http/impl/util/StreamUtils.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/impl/util/StreamUtils.scala
@@ -44,8 +44,7 @@ private[http] object StreamUtils {
    * Creates a transformer that will call `f` for each incoming ByteString and output its result. After the complete
    * input has been read it will call `finish` once to determine the final ByteString to post to the output.
    * Empty ByteStrings are discarded.
-   * If the stage is stopped before the input is fully consumed (e.g. on downstream cancellation or upstream failure),
-   * `cleanup` is called to release any resources held by the transformer.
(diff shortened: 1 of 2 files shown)


pjfanning (approved):
Lgtm

PR #1161, opened 2026-07-13
Create stage-release-candidate.yml

pjfanning:
part of releasing 1.4.0

### DIFF
--- a/.github/workflows/stage-release-candidate.yml
+++ b/.github/workflows/stage-release-candidate.yml
@@ -1,0 +1,354 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+name: Stage release candidate
+
+on:
+  workflow_dispatch:
+    inputs:
+      source-tar-to-svn:
+        description: "Stage the source tarball to svn (old)"
+        default: true
+        type: boolean
+      source-tar-to-atr:
+        description: "Stage the source tarball to ATR (new)"
+        default: true
+        type: boolean
+      jars:
+        description: "Stage the binary jars to nexus"
+        default: true
+        type: boolean
+      email-template:
+        description: "Generate vote email template"
+        default: true
+        type: boolean
+
+permissions:
+  contents: read
+
+jobs:
+  # Automating the step at https://github.com/apache/pekko-site/wiki/Pekko-Release-Process#build-the-source-release-candidate
+  # Partly based on https://github.com/apache/daffodil/blob/main/.github/workflows/release-candidate.yml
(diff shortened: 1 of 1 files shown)


He-Pin (commented):
Reviewing line by line. The workflow structure is solid and follows the established pattern from the pekko repo. A few findings below.

.github/workflows/stage-release-candidate.yml
```diff
@@ -0,0 +1,355 @@
...
+          # To be enabled after this workflow has been tested:
+          #if [[ "$REF" != "v"* ]]; then
+          #  echo "Trigger this workflow on a version tag"
+          #  exit 1
+          #fi
+          if [[ "$REF" != *"-RC"* ]]; then
+            echo "Trigger this workflow on an RC tag"
+            exit 1
+          fi
+          export VERSION=$(echo $REF | sed -e "s/\(.*\)-.*/\\1/")
```
He-Pin, line 194: This sed pattern is missing the leading `.` that all the other jobs use. With tag `v1.4.0-RC1`:

```
# This step (line 194):
echo "v1.4.0-RC1" | sed -e "s/\(.*\)-.*/\1/"   → v1.4.0  ← keeps the v prefix

# Other jobs (lines 61, 136, 248):
echo "v1.4.0-RC1" | sed -e "s/.\(.*\)-.*/\1/"   → 1.4.0   ← correct
```

The "Build, sign and stage artifacts" step (line 230) uses the correct pattern, so artifacts are not mis-versioned — but this check step prints `Version: v1.4.0` in the log, which is misleading when debugging. This same copy-paste bug also exists in apache/pekko line 193.

Suggest aligning with the other jobs:
```yaml
export VERSION=$(echo $REF | sed -e "s/.\(.*\)-.*/\\1/")
```

.github/workflows/stage-release-candidate.yml
```diff
@@ -0,0 +1,355 @@
...
+          echo "VERSION=$VERSION"
+          echo "RC_VERSION=$RC_VERSION"
+          echo "BRANCH=$BRANCH"
+
+          export DISCUSS=$(curl 'https://lists.apache.org/api/stats.lua?list=dev&domain=pekko.apache.org' | jq ".emails.[] | .subject, .mid" | grep -A1 "$MODULE $VERSION" | tail -1 | tr -d \")
+          echo "DISCUSS=$DISCUSS"
+          export DISCUSS_THREAD=https://lists.apache.org/thread/$(curl "https://lists.apache.org/api/thread.lua?id=$DISCUSS&find_parent=true" | jq .thread.mid | tr -d \")
+          echo "DISCUSS_THREAD=$DISCUSS_THREAD"
+
+          export RELEASE_NOTES=https://github.com/apache/pekko-http/pull/$(curl https://api.github.com/repos/apache/pekko-http/pulls?state=all | jq ".[] | .title, .number" | grep -A1 "Release notes for $VERSION" | tail -1)
```
He-Pin, line 264: The GitHub API default pagination is 30 items per page. This repo has 1000+ PRs, so if the "Release notes for $VERSION" PR isn't among the 30 most recently updated ones, this grep will fail silently and `RELEASE_NOTES` will be a broken URL.

Adding `&per_page=100` helps but is still fragile. A more reliable approach would be the search API:
```bash
gh pr list --repo apache/pekko-http --state all \
  --search "Release notes for $VERSION in:title" \
  --json number --jq '"https://github.com/apache/pekko-http/pull/" + (.[0].number | tostring)'
```

Minor concern since for 1.4.0 the release-notes PR will likely be recent enough, but worth hardening before this workflow is reused for future releases.

pjfanning: this is a copy of a script we have in lots of places and it's been working ok - I'd prefer to keep the script as is since this PR does not target main and the improvements should go there first

The generated email is just a template, easily fixed by hand.

This CI script simply doesn't need to be bomb-proof.

He-Pin (approved):

PR #1163, opened 2026-07-13
use java8 in 1.x builds

pjfanning:
java 17 has become the norm in 2.x builds


He-Pin (approved):

He-Pin:
I think we can do it with `release 8 ` too

PR #1172, opened 2026-07-23
Add selective WebSocket message compression

mkurz:
## Summary

Adds server-side per-message compression selection when `permessage-deflate` has been negotiated.

Applications can now decide independently for every outbound text or binary message whether it should be compressed. This supports policies such as skipping sensitive, already-compressed, or small messages while retaining compression for the rest of the WebSocket connection.

## API

The Scala API accepts a `Message => Boolean` selector, while the Java API accepts a `Predicate<Message>`.

Selector overloads are available for:

- `WebSocketUpgrade.handleMessages`
- `WebSocketUpgrade.handleMessagesWithSinkSource`
- `handleWebSocketMessages`
- `handleWebSocketMessagesForProtocol`
- `handleWebSocketMessagesForOptionalProtocol`

The internal low-level frame API additionally accepts a `FrameStart => Boolean` selector.

## Behavior

- The selector is evaluated exactly once for each outbound text or binary message.
- It is evaluated only when `permessage-deflate` was negotiated.
- Control frames are never passed to the selector or compressed.
- One decision is retained across all fragments of a streamed or fragmented message, including when control frames are interleaved.
- Sending an uncompressed message does not modify the DEFLATE context, so later compressed messages continue with the correct context-takeover state.
- Existing APIs retain their current behavior and compress every outbound data message after successful negotiation.

Returning `true` selects compression for the message. It does not enable or negotiate compression by itself.

## Implementation

The high-level message API evaluates the selector before rendering and carries the result on the initial frame for the compression stage.

The low-level frame API evaluates its selector directly for the initial text or binary frame. The compression stage retains that choice until the final continuation frame.

This is similar to Netty's per-message WebSocket extension filter while using Pekko HTTP's existing streaming frame pipeline.

## Compatibility

All existing overloads and internal entry points remain available and delegate to the previous compress-all behavior.

The new Scala trait methods are concrete delegating methods, so existing implementations do not need to implement them.

## Testing

Coverage includes:

- Alternating compressed and uncompressed messages
- Scala and Java selectors
- Strict and streamed messages
- Fragmented messages with interleaved control frames
- Context takeover across skipped messages
- Selector evaluation count
- No selector evaluation without negotiation
- High-level and low-level frame APIs
- Scala and Java routing-directive delegation
- Subprotocol preservation
- Binary compatibility on Scala 2.13 and Scala 3

Validation performed with:

```text
sbt applyCodeStyle
sbt +mimaReportBinaryIssues
sbt '+http-tests/testOnly org.apache.pekko.http.scaladsl.server.directives.WebSocketDirectivesSpec'
sbt 'http-core/testOnly org.apache.pekko.http.impl.engine.ws.WebSocketServerSpec'
```

## References

- #61
- #1114

### DIFF
--- a/docs/src/test/java/docs/http/javadsl/server/WebSocketCoreExample.java
+++ b/docs/src/test/java/docs/http/javadsl/server/WebSocketCoreExample.java
@@ -21,6 +21,7 @@ import java.util.Optional;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Predicate;
 import org.apache.pekko.NotUsed;
 import org.apache.pekko.actor.ActorSystem;
 import org.apache.pekko.http.javadsl.ConnectionContext;
@@ -32,6 +33,7 @@ import org.apache.pekko.http.javadsl.model.StatusCodes;
 import org.apache.pekko.http.javadsl.model.ws.Message;
 import org.apache.pekko.http.javadsl.model.ws.TextMessage;
 import org.apache.pekko.http.javadsl.model.ws.WebSocketRequest;
+import org.apache.pekko.http.javadsl.model.ws.WebSocketUpgrade;
 import org.apache.pekko.http.javadsl.settings.ClientConnectionSettings;
 import org.apache.pekko.http.javadsl.settings.ServerSettings;
 import org.apache.pekko.http.javadsl.settings.WebSocketSettings;
@@ -122,6 +124,25 @@ public class WebSocketCoreExample {
   }
 
   // #websocket-handler
+
+  static HttpResponse selectiveCompression(
+      WebSocketUpgrade upgrade, Flow<Message, Message, NotUsed> handler) {
+    // #websocket-selective-compression
+    Predicate<Message> shouldCompress =
+        message -> {
+          if (!message.isStrict())
+            return true; // Compress streamed messages without buffering them.
+          if (message.isText()) {
+            return ByteString.fromString(message.asTextMessage().getStrictText()).size() >= 1024;
+          } else {
+            return message.asBinaryMessage().getStrictData().size() >= 1024;
+          }
+        };
+
+    HttpResponse response = upgrade.handleMessagesWith(handler, shouldCompress);
+    // #websocket-selective-compression
+    return response;
+  }
 
   {
     ActorSystem system = null;
--- a/docs/src/test/java/docs/http/javadsl/server/WebSocketCoreExample.java
+++ b/docs/src/test/java/docs/http/javadsl/server/WebSocketCoreExample.java
@@ -128,16 +128,7 @@ public class WebSocketCoreExample {
   static HttpResponse selectiveCompression(
       WebSocketUpgrade upgrade, Flow<Message, Message, NotUsed> handler) {
     // #websocket-selective-compression
-    Predicate<Message> shouldCompress =
-        message -> {
-          if (!message.isStrict())
-            return true; // Compress streamed messages without buffering them.
-          if (message.isText()) {
-            return ByteString.fromString(message.asTextMessage().getStrictText()).size() >= 1024;
-          } else {
-            return message.asBinaryMessage().getStrictData().size() >= 1024;
-          }
-        };
+    Predicate<Message> shouldCompress = Message::isText;
 
     HttpResponse response = upgrade.handleMessagesWith(handler, shouldCompress);
     // #websocket-selective-compression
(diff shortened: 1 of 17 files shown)


mkurz:
This is needed for Play so we can implement a threshold:
- https://github.com/playframework/playframework/pull/14120

docs/src/test/java/docs/http/javadsl/server/WebSocketCoreExample.java
```diff
@@ -123,6 +125,25 @@ public static TextMessage handleTextMessage(TextMessage msg) {
...
 
+  static HttpResponse selectiveCompression(
+      WebSocketUpgrade upgrade, Flow<Message, Message, NotUsed> handler) {
+    // #websocket-selective-compression
+    Predicate<Message> shouldCompress =
+        message -> {
+          if (!message.isStrict())
+            return true; // Compress streamed messages without buffering them.
+          if (message.isText()) {
+            return ByteString.fromString(message.asTextMessage().getStrictText()).size() >= 1024;
```
pjfanning: this seems complicated - is there no simpler way to do this? or some simple enhancement that would make it easier

http-core/src/test/scala/org/apache/pekko/http/impl/engine/ws/WebSocketServerSpec.scala
```diff
@@ -14,12 +14,15 @@
...
 import java.io.ByteArrayOutputStream
+import java.util.concurrent.atomic.AtomicInteger
+import java.util.function.Predicate
 import java.util.zip.Deflater
 import java.util.zip.Inflater
 
 import org.apache.pekko
 import pekko.actor.ActorSystem
 import pekko.http.scaladsl.model.ws._
+import pekko.http.javadsl.model.ws.{ Message => JavaMessage, TextMessage => JavaTextMessage }
```
pjfanning, line 26: move above scaladsl import above - alphabetic

mkurz: Done, see commit

docs/src/test/java/docs/http/javadsl/server/WebSocketCoreExample.java
mkurz: if you mean the example is too complicated, I simplified it to a straightforward policy that compresses text messages and leaves binary messages uncompressed, see the new commit.

The previous example demonstrated a payload-size threshold. Providing a generic threshold helper is less straightforward for streamed messages because their total size is unknown when the compression decision must be made, before the first frame is emitted. Such a helper would need to define a policy such as “compress streamed messages unconditionally and apply the threshold only to strict messages,” or buffer streamed data until the threshold is reached, which adds buffering and latency.

I therefore kept this PR focused on the general selector API and made the documentation example simpler. A built-in threshold policy could be considered separately if desired.

pjfanning: I was more worried about the `ByteString.fromString(message.asTextMessage().getStrictText()).size() >= 1024` and thought that it might be possible to get the same result with simpler code or to provide an API that that calculates the TextMessage length.

mkurz: ok, however I would prefer to keep a reusable payload-size API out of this PR and consider it separately if there is broader demand.

codex also has an opinion on that:
`getStrictText().length()` would count UTF-16 code units, while the compression threshold in the example was based on the UTF-8 bytes sent on the wire. That is why the example used `ByteString.fromString(...).size()`.

We could add an API such as `getStrictTextUtf8ByteLength`, but a general `TextMessage.length` would be ambiguous, and streamed messages do not have a known total length without consuming them. Such a helper would also still need to encode the string, which the WebSocket renderer subsequently does again.

Since the threshold was only intended as an example of a possible selector, I simplified the example instead.

mkurz: IMHO we can resolve this thread

mkurz:
any chance this gets merged within the next 5 hours so it makes it into the next nightly (so tomorrow I can build against that in Play)? (no pressure however, just checking)

pjfanning (approved):
lgtm

pjfanning:
@mkurz could you fix the merge conflicts and I'll merge.
If anyone spots any issues post merge, we can look at fixing them.
It will be a few weeks before we get around to a 2.0.0-M2 RC because I want to get the Pekko Core 2.0.0-M4 out first.

mkurz:
Thanks!

pjfanning:
2.0.0-M1+205-35174606-SNAPSHOT is published

PR #1191, opened 2026-07-29
escape html in FileAndResourceDirectives

pjfanning:
Added escapeHtml methods to the DirectoryListing object that escape the five HTML special characters (&, <, >, ", '). Applied to all user-controlled strings in the directory listing HTML:

- path in `<title>` and `<h1>` tags
- path in the "../" parent directory link href
- path + name in file/directory link hrefs (via the two-argument overload, avoiding an intermediate string concatenation)
- name as link text content

Pretty unlikely but there is a theoretical chance that filenames could have problematic XSS type names.
Claude AI thinks that we should harden the code as a result.

### DIFF
--- a/http/src/main/scala/org/apache/pekko/http/scaladsl/server/directives/FileAndResourceDirectives.scala
+++ b/http/src/main/scala/org/apache/pekko/http/scaladsl/server/directives/FileAndResourceDirectives.scala
@@ -400,6 +400,29 @@ object DirectoryListing {
       |</html>
       |""".stripMarginWithNewline("\n").split('$')
 
+  private def escapeHtml(s: String): String = escapeHtml(s, "")
+
+  private def escapeHtml(s1: String, s2: String): String = {
+    val sb = new java.lang.StringBuilder(s1.length + s2.length + 16)
+    def appendEscaped(s: String): Unit = {
+      var i = 0
+      while (i < s.length) {
+        s.charAt(i) match {
+          case '&'  => sb.append("&amp;")
+          case '<'  => sb.append("&lt;")
+          case '>'  => sb.append("&gt;")
+          case '"'  => sb.append("&quot;")
+          case '\'' => sb.append("&#39;")
+          case c    => sb.append(c)
+        }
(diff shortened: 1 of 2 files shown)


http/src/main/scala/org/apache/pekko/http/scaladsl/server/directives/FileAndResourceDirectives.scala
```diff
@@ -412,15 +429,24 @@ object DirectoryListing {
       def maxNameLength(seq: Seq[(File, String)]) = if (seq.isEmpty) 0 else seq.map(_._2.length).max
       val maxNameLen = math.max(maxNameLength(directoryFilesAndNames) + 1, maxNameLength(fileFilesAndNames))
       val sb = new java.lang.StringBuilder
-      sb.append(html(0)).append(path).append(html(1)).append(path).append(html(2))
+      val escapedPath = escapeHtml(path)
+      sb.append(html(0)).append(escapedPath).append(html(1)).append(escapedPath).append(html(2))
       if (!isRoot) {
         val secondToLastSlash = path.lastIndexOf('/', path.lastIndexOf('/', path.length - 1) - 1)
```
He-Pin, line 435: Nit: `escapeHtml(path)` here is called once per file/directory entry, but you already computed `escapedPath` a few lines above. Could just reuse it:

```scala
sb.append("<a href=\"").append(escapedPath).append(escapedName).append("\">").append(escapedName).append("</a>")
```

Avoids redundant escaping in directories with many entries.

He-Pin (commented):
Overall LGTM — the escaping logic is correct and covers all user-controlled insertion points. One small suggestion inline.

He-Pin (approved):

http/src/main/scala/org/apache/pekko/http/scaladsl/server/directives/FileAndResourceDirectives.scala
pjfanning: escapedPath is already handled this way

PR #1194, opened 2026-07-29
add warnings about limitations of idle-timeout

pjfanning:
The idle-timeout does catch completely idle connections, but an attacker sending one byte per second would keep the connection alive indefinitely since data is flowing. The request timeout only starts after the entity is complete.

The real mitigations users can apply:
- Lower idle-timeout for untrusted clients
- Use max-content-length to bound total entity size
- Add a reverse proxy (nginx, etc.) that has its own slowloris protections

Adding a per-entity receive timeout to the framework would be a significant feature.

### DIFF
--- a/docs/src/main/paradox/common/timeouts.md
+++ b/docs/src/main/paradox/common/timeouts.md
@@ -15,6 +15,15 @@ on a connection for over `idle-timeout` time, the connection will be automatical
 This setting should be used as a last-resort safeguard to prevent unused or stuck connections from consuming resources for
 an indefinite time.
 
+This is a bidirectional inactivity timeout. It triggers only when no
+bytes are sent or received for the configured duration. Traffic in
+either direction, including a client sending bytes at intervals shorter
+than the timeout, keeps the connection alive; it is not an overall
+request-receive timeout. Lowering this value and configuring
+`pekko.http.server.parsing.max-content-length` reduce exposure but do not
+bound request duration. Use an appropriately configured reverse proxy or
+application-level entity timeout when a hard receive deadline is required.
+
 The setting works the same way for server and client connections and it is configurable independently using the following keys:
 
 ```
--- a/http-core/src/main/resources/reference.conf
+++ b/http-core/src/main/resources/reference.conf
@@ -45,6 +45,11 @@ pekko.http {
 
     # The time after which an idle connection will be automatically closed.
     # Set to `infinite` to completely disable idle connection timeouts.
+    #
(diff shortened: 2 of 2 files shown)


mkurz (changes requested):
**Findings**

1. **Medium: the timeout direction is described incorrectly.**

   The new comment says it triggers when “no data is received.” Pekko uses `bidirectionalIdleTimeout`, so traffic in either direction resets it. This should say “when no bytes are sent or received.” [Implementation](https://github.com/apache/pekko-http/blob/4d35672ddc516fed049e5f5e4fda9e6260c951b9/http-core/src/main/scala/org/apache/pekko/http/impl/engine/HttpConnectionIdleTimeoutBidi.scala#L43-L57)

2. **Medium: the suggested mitigations could imply stronger protection than they provide.**

   Lowering `idle-timeout` only forces an attacker to send bytes more frequently. `max-content-length` limits entity bytes, but does not impose a receive deadline, protect slowly delivered headers, or help until the entity stream is materialized. [Entity-limit semantics](https://github.com/apache/pekko-http/blob/4d35672ddc516fed049e5f5e4fda9e6260c951b9/docs/src/main/paradox/common/http-model.md#L219-L236)

   The comment should explicitly say these settings reduce exposure but do not bound request duration. The reverse-proxy or application-level timeout recommendation from the PR body should be retained.

3. **Low: the canonical timeout documentation remains incomplete.**

   The same limitation should be added to `docs/src/main/paradox/common/timeouts.md`, where idle-timeout is documented publicly. That page already correctly describes the timeout as bidirectional. [Timeout documentation](https://github.com/apache/pekko-http/blob/4d35672ddc516fed049e5f5e4fda9e6260c951b9/docs/src/main/paradox/common/timeouts.md#L9-L24)

Suggested wording:

```hocon
# This is a bidirectional inactivity timeout. It triggers only when no
# bytes are sent or received for the configured duration. Traffic in
# either direction, including a client sending bytes at intervals shorter
# than the timeout, keeps the connection alive; it is not an overall
# request-receive timeout. Lowering this value and configuring
# pekko.http.server.parsing.max-content-length reduce exposure but do not
# bound request duration. Use an appropriately configured reverse proxy or
# application-level entity timeout when a hard receive deadline is required.
```

pjfanning:
Thanks @mkurz - changed as suggested.

I'll have a look at docs/src/main/paradox/common/timeouts.md

nvollmar (approved):

PR #1218, opened 2026-08-22
Compare path elements when checking that a served file is below the base directory

pjfanning:
### Motivation
`checkIsSafeDescendant` in `FileAndResourceDirectives` decides whether a requested file is really below the served directory by comparing the two canonical paths as plain strings:

```scala
if (!canonicalFinalPath.startsWith(baseFile.getCanonicalPath))
```

A string prefix is not a path prefix. With `/var/www` served, the location `/var/www-private/secret` satisfies `startsWith` without being contained in the served directory. The segment filter in `safeJoinPaths` does not catch this, since no path segment is suspicious — the path only moves out of the directory when `getCanonicalPath` resolves a symbolic link that points at such a sibling. The file is then served.

### Modification
Compare the canonical paths element by element through `java.nio.file.Path` rather than as strings. The base directory itself still compares as contained, which the directory listing in `getFromBrowseableDirectory` relies on (its root listing resolves to exactly the base path).

### Result
Only files that are genuinely below the served directory are served. A symbolic link into a sibling directory is rejected with the warning that was already there, whatever the sibling is called. Requests that were previously served are unaffected, and the existing symlink and traversal tests still pass.

### Tests
- `sbt "http-tests/testOnly org.apache.pekko.http.scaladsl.server.directives.FileAndResourceDirectivesSymlinkSpec"` - pass; new test serves through a symlink pointing at a sibling directory named after the served one, and asserts the request is rejected. Verified that it fails without the change (the file is served, `handled` is `true`).
- `sbt http-tests/test` - pass (1484 tests)
- `sbt http/mimaReportBinaryIssues` - pass
- `sbt http/scalafmt http-tests/Test/scalafmt` - clean

### References
None - tightens the containment check for file and resource directives

### DIFF
--- a/http-tests/src/test/scala/org/apache/pekko/http/scaladsl/server/directives/FileAndResourceDirectivesSpec.scala
+++ b/http-tests/src/test/scala/org/apache/pekko/http/scaladsl/server/directives/FileAndResourceDirectivesSpec.scala
@@ -195,6 +195,18 @@ class FileAndResourceDirectivesSpec extends RoutingSpec with Inspectors with Ins
       shouldReject("..%c0%af", warnings = 0)
       shouldReject("..%c1%9c", warnings = 0)
     }
+    "reject requests whose path can never name a file" in {
+      // a percent-encoded NUL decodes into the path segment, but no file-system path may contain it, so
+      // canonicalization throws; that must surface as a rejection, not as an error escaping to the exception handler
+      def route(uri: String) =
+        mapRequestContext(_.withUnmatchedPath(Path("/" + uri))) { _getFromDirectory("someDir") }
+
+      EventFilter.warning(pattern = ".* points to a location that is not part of .*", occurrences = 1).intercept {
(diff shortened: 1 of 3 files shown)


pjfanning:
Updated after review: rebased onto main and pushed 1144ef34a addressing the findings.

- `checkIsSafeDescendant` now catches `InvalidPathException`/`IOException` and treats such paths as not contained — previously a Windows segment like `a<b` (or `%00` on any platform, pre-existing) escaped as a 500 instead of the clean rejection, skipping the traversal warning. New directional test uses a `%00` segment; verified failing with the fix stashed.
- The scaladoc above `safeDirectoryChildPath` now states the element-wise containment contract instead of the string-prefix semantics the fix removed, and notes that Windows canonicalization does not resolve NTFS symlinks/junctions — that link-escape class remains open there (pre-existing, `Path.toRealPath` would be the principled follow-up).
- Symlink spec fixture no longer writes content nothing asserts on.

He-Pin (approved):

PR #1221, opened 2026-08-23
Parse Content-Length as digits with surrounding whitespace only

pjfanning:
### Motivation
`ContentLengthParser` skips whitespace anywhere in the field value:

```scala
if (DIGIT(c)) { ... } else if (WSP(c)) recurse(ix + 1, result)
```

So `Content-Length: 1 2` is read as `12`, `Content-Length: 5 5` as `55`, and an empty `Content-Length:` as `0`. The field value is `1*DIGIT` surrounded by optional whitespace (RFC 9110, section 8.6 and RFC 9112, section 5).

The other ways two implementations could disagree about a body length are already closed here — two Content-Length headers with different values are rejected, `Transfer-Encoding` accepts only a single `chunked`, and `chunked` together with a `Content-Length` is rejected — which is what makes this one worth tightening: it is the remaining case where pekko-http reads a length that a proxy or another server in the request path reads differently, or rejects.

### Modification
Skip whitespace before and after the digits, require at least one digit, and end the value at the first non-digit.

### Result
A value with whitespace between digits, or with no digit at all, is rejected with the same "Illegal \`Content-Length\` header value" error that other malformed values already produce. Leading and trailing whitespace still parse as before.

This rejects two inputs that were accepted before: whitespace inside the digits, and an empty value that used to be read as `0`. Both are malformed per the grammar, but it is a behaviour change for anyone who was relying on the lenient reading.

### Tests
- `sbt "http-core/testOnly org.apache.pekko.http.impl.engine.parsing.RequestParserCRLFSpec org.apache.pekko.http.impl.engine.parsing.RequestParserLFSpec org.apache.pekko.http.impl.engine.parsing.ResponseParserSpec"` - pass, with 2 new tests, one for whitespace inside the value and one for an empty value; both fail without the change. Leading whitespace stays covered by the existing `Content-length:    17` tests, so the accepted cases are pinned too.
- `sbt http-core/test` - pass (1218 tests)
- `sbt http-tests/test` - pass (1483 tests)
- `sbt http-core/mimaReportBinaryIssues` - pass
- `sbt http-core/scalafmt http-core/Test/scalafmt` - clean

### References
None - tightens Content-Length parsing to the grammar

### DIFF
--- a/http-core/src/main/scala/org/apache/pekko/http/impl/engine/parsing/SpecializedHeaderValueParsers.scala
+++ b/http-core/src/main/scala/org/apache/pekko/http/impl/engine/parsing/SpecializedHeaderValueParsers.scala
@@ -33,21 +33,32 @@ private[parsing] object SpecializedHeaderValueParsers {
   def specializedHeaderValueParsers = Seq(ContentLengthParser)
 
   object ContentLengthParser extends HeaderValueParser("Content-Length", maxValueCount = 1) {
+    // The field value is `1*DIGIT`, surrounded by optional whitespace (RFC 9110, section 8.6 and RFC 9112,
+    // section 5). Whitespace within the digits must not be skipped: a value like `1 2` would then be read as `12`
+    // here while another implementation in the request path rejects it or reads it as `1`.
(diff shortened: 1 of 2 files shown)


Philippus (approved):
nice catch!

PR #1242, opened 2026-08-27
Add a draft threat model and wire it for discoverability

potiuk:
### Motivation

The Pekko PMC asked the ASF Security team for a draft threat model to refine, as
preparation for a Claude security scan. This is the companion to
[the same change in `apache/pekko`](https://github.com/apache/pekko/pulls) — that model
covers the actor, stream, remoting and cluster layers, and this one does not restate them.

`docs/src/main/paradox/security.md` already carries a **"Security model"** section, and it
makes the most consequential statement in this document: applications *"should not be
exposed to the public internet directly"*, and DoS resistance is claimed only as
*"pretty well under most known Denial of Service attacks"*. But that is three sentences,
and the quantitative envelope that actually bounds untrusted input — the parsing limits in
`http-core`'s `reference.conf` — is not connected to it anywhere.

Separately, the `AGENTS.md` → `SECURITY.md` chain that a scanning agent follows to locate
a project's model does not resolve today: there is no `SECURITY.md` in the repo.

### Modification

- **`THREAT_MODEL.md`** — a v0 draft. **§5a** collects the shipped parsing limits,
  timeouts and CORS defaults as the model's quantitative spine; **§8** states the six
  default-on properties they underwrite. Every claim is tagged `(documented)` with a
  citation or `(inferred)` with a matching question in §14.
- **`SECURITY.md`** — reporting policy (`security@apache.org`; Pekko does not operate a
  separate project security list), plus the three points that catch most reporters.
- **`AGENTS.md`** — a two-line `## Security` section pointing at both.

Nothing `security.md` asserts has been dropped or weakened — **§15 back-maps** each
existing statement to the section that now carries it.

### Result

The chain `AGENTS.md` → `SECURITY.md` → `THREAT_MODEL.md` resolves on `main`, and triagers
get a citable document with a closed set of dispositions (§13).

Worth noting in contrast to the `apache/pekko` model: **every §8 property here is
default-on**. Pekko HTTP's parsing limits, timeouts, strict URI/cookie parsing and
response-splitting protection all ship enabled. The soft spot is not the defaults, it is
the boundary of the DoS claim.

**§14 holds 10 questions, each stated as a proposed answer** — confirming or correcting is
enough. The two worth reading first:

- **Q1 — where exactly is the DoS line?** `security.md`'s claim is deliberately graded, and
  triage needs a split. *Proposed:* a request **within** all §5a limits that causes
  super-linear CPU or memory is `VALID`; anything requiring limits to be raised, or
  requiring request *volume* rather than request *content*, is out of model because volume
  defence is the fronting proxy's job. Is that the intended reading?
- **Q2 — the `http-cors` defaults.** The module ships `allowed-origins = "*"` together with
  `allow-credentials = yes`. Per its own `reference.conf`, that combination does **not**
  send a literal `*` — it echoes the request's `Origin` and allows credentials. Two things
  bound how alarming that is: `cors()` is opt-in, and the interaction is documented. Is the
  default deliberate, or should it change?

**This is a draft and should not be treated as canonical until at least Q1 and Q2 are
answered.** Happy to fold answers in, or hand it over entirely — the PMC owns the document.

### Tests

Not run - docs only

### References

None - requested by the Pekko PMC chair on the ASF security scan thread


pjfanning:
Thanks Jarek. My plan is to concentrate on https://github.com/apache/pekko/pull/3478 and to return to this when we make progress on that. If you prefer, we can merge this and discuss with the Pekko PMC how to complete it.

potiuk:
Sure. Tak your time - and indeed `pekko` one is more important .

pjfanning:
I verified the model's claims against the source and pushed 17cbbf6bc with corrections — three things did not survive the check:

1. **§5 "writes no files" was false.** `fileUploadAll` creates temp files via `Files.createTempFile` + `deleteOnExit`, and `storeUploadedFile(s)` streams entity bytes to an application-chosen destination (`FileUploadDirectives.scala:178`). The scan behind §14 Q9 missed it. §5 now states the accurate boundary (writes happen only through upload directives the application installs) and Q9 records the correction.

2. **Q3 is now answered — as a correction, and it found a real bug.** `checkIsSafeDescendant` compared canonical paths as *strings*: with `/var/www` served, a symlink resolving into `/var/www-private/` passed `startsWith` and the file was served. That is exactly the §5b.4 "walked out of" class, disposition `VALID`, and #1218 (in flight) fixes it by comparing path elements. That fix also settles the draft's symlink residual: an escape from the root is a `VALID` finding, not an operator responsibility. §9, §15 and the status/tally lines updated to match — the "all ten answered" claim is now actually true (the status previously also said "Q1-Q8 answered" with Q3 still ending in an open question).

3. **The commit pin was self-referential.** `444d939` is this PR's own first commit; the pin now points at `85d7243`, the `main` commit the branch is based on, which is where the cited line numbers were verified.

Also added to §12: #1217 proposes replacing the per-upload `deleteOnExit` with a shutdown hook registered by Pekko HTTP itself, which on merge invalidates the §5 "registers no shutdown hook" claim and its §11a non-finding — whichever of these lands second needs to reconcile.

For the record, the rest verified clean: the full §5a table against `reference.conf`, the CORS echo-with-credentials behaviour (`CorsSettingsImpl.scala:63-67`), the P7 CONTINUATION accumulation bound (`HeaderDecompression.scala:171`), `secure_==`, and the `extractClientIP`/`extractDirectClientIP` split.

pjfanning:
Pushed 12c07b502 folding in the post-review state of the two in-flight PRs the model references:

- **§14 Q3 / §9**: the symlink-rejection claim is now scoped to platforms where `File.getCanonicalPath` resolves links. On Windows it does not resolve NTFS symlinks/junctions (a caveat #1218 now documents at the function), so that link-escape residual is on record with `Path.toRealPath` as the principled follow-up, and the operator's link-free-tree choice carries containment there. Also noted: #1218 now rejects path segments no file-system path may contain (NUL, `<` on Windows) instead of letting them escape as server errors.
- **§5**: the upload-directive note reflects #1217's rework — one temp directory per actor system removed by a `CoordinatedShutdown` task after the drain, replacing per-file `deleteOnExit`. (The earlier §12 note already records that this rework keeps the "registers no shutdown hook" claim true.)

potiuk:
This is better than what we drafted, and the three corrections are the useful part — so, briefly, on each:

**§5 "writes no files".** Straightforwardly wrong, and worth naming as a class rather than just a fix. The draft asserted a negative ("writes no files") from a scan that hadn't enumerated the upload directives; `fileUploadAll` and `storeUploadedFile(s)` were always there to be found. A model claiming a capability is *absent* needs the same evidence as one claiming it's present, and this one didn't have it.

**Q3 → #1218.** This is the one that matters. A string `startsWith` on canonical paths passing `/var/www-private/` against a `/var/www` root is a real escape, and the draft had it as an operator responsibility — i.e. the model would have taught a future scan *not* to report it. Your reclassification to `VALID` is right, and the Windows caveat (`getCanonicalPath` not resolving NTFS junctions, `Path.toRealPath` as the principled fix) is the kind of platform-specific qualification a generated draft reliably misses.

**The self-referential pin.** Fair, and embarrassing — pinning `444d939` pinned this PR's own first commit rather than the `main` commit the claims were verified against. `85d7243` is correct.

On merging: yes, merge it. You offered earlier to land it and finish with the PMC, and I think that's right — the document is now yours in substance, and the parts still moving (#1217's `CoordinatedShutdown` rework, #1218's element-wise comparison) are yours to reconcile as they land. §12 already records which claim each one invalidates, which is the part that usually gets lost.

One thing worth recording rather than asking about: these three corrections are the same failure mode several projects reported to us independently this month — a claim inferred without walking the enforcement path to check it. The §5 one is going back to the scan team as a concrete example of it, described by shape rather than by project.

pjfanning:
@samueleresca do you think this is ok to merge? It can be a living document that we update over time, including if we get reports that seem invalid but that the threat model seems to support as being an issue.

samueleresca:
> @samueleresca do you think this is ok to merge? It can be a living document that we update over time, including if we get reports that seem invalid but that the threat model seems to support as being an issue.


I have been busy. I take a final look later but otherwise it is good to go

THREAT_MODEL.md
```diff
@@ -0,0 +1,392 @@
...
+- **`allowed-origins = "*"` does not mean "no credentials are exposed".** Combined with the shipped `allow-credentials = yes`, it echoes the caller's `Origin` (§5a).
+- **`remote-address-attribute` is not the client IP behind a proxy.** It is the socket peer.
+- **`max-content-length` is not a global memory bound.** It bounds one entity; concurrent connections multiply it.
+- **A rejection is not a failure.** The routing DSL's rejection mechanism is control flow, not a security control — an unhandled rejection can fall through to a different route.
+
+### Well-known attack classes left to the caller
+
+- **Request smuggling / desync** between a fronting proxy and Pekko HTTP — inherently a two-party property; strict parsing (P4) helps but cannot settle it alone.
+- **Slow-loris and connection exhaustion** — `idle-timeout` and `max-connections` (P5) bound what one connection holds and how many are accepted, but exhausting those bounds by volume is disclaimed (§14 Q1). A single connection that evades `idle-timeout` while holding resources is the in-scope version.
+- **Decompression bombs** in request bodies, where the application enables decoding.
```
samueleresca: nit: Decompression bombs left to the caller is not entirely true. The operator still has options to configure some maximums in the request (e.g., decode-max-bytes-per-chunk and `decode-max-size`: https://github.com/apache/pekko-http/blob/85d7243e745d68c987b79c5c4d56958f582a5c34/http/src/main/resources/reference.conf#L41-L50

samueleresca (approved):
looks good to me. I left a nit but feel free to address it or not. Let's keep the scans going and worst case we can revisit the model post adjustments.

THREAT_MODEL.md
pjfanning: I updated this

samueleresca (approved):

PR #1264, opened 2026-08-31
feat: bound the size of an incoming HTTP/2 frame

pjfanning:
### Motivation

`Http2FrameParsing` read a frame's 24-bit length field and then took that many bytes with no upper bound:

```scala
val length = reader.readShortBE() << 8 | reader.readByte()
...
val payload = reader.take(length)
```

So a peer could declare a length up to the field's maximum of 16 MiB - 1 and make the frame parser buffer that much for a single frame. This happens *before* the HPACK and entity limits apply, so `max-header-list-size`, `incoming-stream-level-buffer-size` and `incoming-connection-level-buffer-size` do not bound it — and it is multiplied by the number of connections.

### Modification

Add a `max-frame-size` setting to the HTTP/2 server and client settings, defaulting to `512kB`, and reject a larger frame with a `FRAME_SIZE_ERROR` on its frame header, before the payload is buffered. The value is validated against the bounds RFC 9113 §4.2 sets for `SETTINGS_MAX_FRAME_SIZE` (16 KiB … 16 MiB - 1).

**The setting is a limit on what is accepted, not an advertisement**, and that is deliberate. An earlier revision of this branch *did* advertise it as `SETTINGS_MAX_FRAME_SIZE`, and h2spec's "4.2 Frame Size" case showed that to be actively harmful: DATA frames are flow controlled and the initial window is 64 KiB, so a peer that sizes its first frame to a larger advertised value trips a `FLOW_CONTROL_ERROR` before the window has grown. h2spec did exactly that and the conformance test went red.

Advertising nothing keeps a well-behaved peer at the 16 KiB default it already assumes, and the configured value is then pure leniency for peers that exceed it — we always accept at least as much as any peer is told it may send. That also preserves pekko's existing behaviour of accepting frames well above 16 KiB, which its own tests rely on.

Why 512 kB and not the RFC ceiling: defaulting to 16 MiB would leave the default posture exactly as it is today and fix nothing unless an operator tuned it. 512 kB matches `incoming-stream-level-buffer-size`, so the parser's transient buffering stays the same order as the stream buffer instead of 32× larger, while still allowing 32× the protocol default.

### Result

A single frame can no longer make the parser hold up to 16 MiB; the bound is 512 kB by default and configurable. Frames from a spec-compliant peer are unaffected.

### Tests

- `sbt "http2-tests/testOnly ...Http2ServerSpec ...Http2ClientSpec ...H2SpecIntegrationSpec"` — pass (251 tests), including the h2spec conformance suite. A new test sends a frame one byte over a configured 16 KiB limit and expects `GOAWAY(FRAME_SIZE_ERROR)`; verified it fails with the size check disabled.
- Two existing tests deliberately send a single 512001-byte frame to exceed the 512 kB stream buffer: "fail if more data is received than stream-level window allows" and "release connection-level flow control accounting when a stream-level window is exceeded" (added by #1281). Both now raise `max-frame-size` to 1 MiB in order to reach the flow-control check they are actually about.
- `sbt "+http-core/mimaReportBinaryIssues"` — pass on 2.13.18 and 3.3.8. The filters cover the new settings members (including the concrete javadsl `withMaxFrameSize`/`getMaxFrameSize`, which MiMa still reports as `ReversedMissingMethodProblem`), the replaced internal `Http2FrameParsing` constructor, and the `copy$default$N` shifts from the new private case-class field.
- `sbt "+http-core/compile"` — pass on 2.13.18 and 3.3.8.
- Native `scalafmt` on the changed files — clean.

### References

None - bounds how much a single HTTP/2 frame can buffer

### DIFF
--- a/http-core/src/main/mima-filters/2.0.x.backwards.excludes/max-frame-size.excludes
+++ b/http-core/src/main/mima-filters/2.0.x.backwards.excludes/max-frame-size.excludes
@@ -1,0 +1,22 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# new max-frame-size setting
+ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.pekko.http.javadsl.settings.Http2ClientSettings.maxFrameSize")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.pekko.http.javadsl.settings.Http2ServerSettings.withMaxFrameSize")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.pekko.http.scaladsl.settings.Http2ClientSettings.maxFrameSize")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.pekko.http.scaladsl.settings.Http2ServerSettings.maxFrameSize")
--- a/http-core/src/main/mima-filters/2.0.x.backwards.excludes/max-frame-size.excludes
+++ b/http-core/src/main/mima-filters/2.0.x.backwards.excludes/max-frame-size.excludes
@@ -17,6 +17,17 @@
 
 # new max-frame-size setting
 ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.pekko.http.javadsl.settings.Http2ClientSettings.maxFrameSize")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.pekko.http.javadsl.settings.Http2ClientSettings.withMaxFrameSize")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.pekko.http.javadsl.settings.Http2ServerSettings.getMaxFrameSize")
 ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.pekko.http.javadsl.settings.Http2ServerSettings.withMaxFrameSize")
 ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.pekko.http.scaladsl.settings.Http2ClientSettings.maxFrameSize")
 ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.pekko.http.scaladsl.settings.Http2ServerSettings.maxFrameSize")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.pekko.http.scaladsl.settings.Http2CommonSettings.maxFrameSize")
+
+# internal frame parser now takes the limit as a constructor parameter
+ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.http.impl.engine.http2.framing.Http2FrameParsing.this")
+
+# the new field shifts the positions of the private settings case class parameters after it
+ProblemFilters.exclude[IncompatibleResultTypeProblem]("org.apache.pekko.http.scaladsl.settings.Http2ClientSettings#Http2ClientSettingsImpl.copy$default$8")
+ProblemFilters.exclude[IncompatibleResultTypeProblem]("org.apache.pekko.http.scaladsl.settings.Http2ClientSettings#Http2ClientSettingsImpl.copy$default$11")
+ProblemFilters.exclude[IncompatibleResultTypeProblem]("org.apache.pekko.http.scaladsl.settings.Http2ServerSettings#Http2ServerSettingsImpl.copy$default$9")
--- a/http-core/src/main/resources/reference.conf
+++ b/http-core/src/main/resources/reference.conf
@@ -263,6 +263,20 @@ pekko.http {
       # effective limit for a well-behaved peer is somewhat stricter than the configured value.
       max-header-list-size = 64 KiB
 
+      # The largest frame payload this endpoint accepts, in bytes. A larger incoming frame is rejected with a
+      # FRAME_SIZE_ERROR on its frame header, before the payload is buffered, so this bounds how much a single frame
+      # can make this endpoint hold at once. The length field of a frame header allows up to 16 MiB - 1.
+      #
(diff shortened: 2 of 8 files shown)


http2-tests/src/test/scala/org/apache/pekko/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -657,6 +669,10 @@ class Http2ServerSpec extends Http2SpecWithMaterializer("""
           })
         "fail if more data is received than stream-level window allows".inAssertAllStagesStopped(
           new WaitingForRequestData {
+            // the single frame below is deliberately bigger than the stream-level buffer, and so also bigger than the
+            // default max-frame-size; raise that limit so the frame reaches the flow-control check this test is about
+            override def settings: ServerSettings = super.settings.mapHttp2Settings(_.withMaxFrameSize(1024 * 1024))
```
He-Pin, line 810: One more test needs this same override: `"release connection-level flow control accounting when a stream-level window is exceeded"`, a few tests below this one. It sends the identical single `512001`-byte `DataFrame` and only overrides `withIncomingConnectionLevelBufferSize`, so with `max-frame-size = 512kB` (HOCON `kB` = 1000, i.e. 512,000) the frame is now rejected here with `GOAWAY(FRAME_SIZE_ERROR)` instead of reaching the flow-control check, and `expectRST_STREAM(TheStreamId, FLOW_CONTROL_ERROR)` plus the window assertion below it both fail.

#1281 added that test on Sep 11, eleven days after these commits, and the PR still reports `MERGEABLE` since the hunks do not overlap — so it only breaks after merge. The PR body's "it was the only test in the suite that needed adjusting" needs updating as well.

http-core/src/main/mima-filters/2.0.x.backwards.excludes/max-frame-size.excludes
```diff
@@ -0,0 +1,22 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# new max-frame-size setting
+ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.pekko.http.javadsl.settings.Http2ClientSettings.maxFrameSize")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.pekko.http.javadsl.settings.Http2ServerSettings.withMaxFrameSize")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.pekko.http.scaladsl.settings.Http2ClientSettings.maxFrameSize")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.pekko.http.scaladsl.settings.Http2ServerSettings.maxFrameSize")
```
He-Pin, line 24: Two filters short of the precedent for this exact change shape. `http2-max-header-list-size.excludes` (#1216 — same structure: a new abstract javadsl/scaladsl member pair plus their concrete companions) carries six, and two of them cover concrete javadsl members that this file omits:

```
ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.pekko.http.javadsl.settings.Http2ClientSettings.withMaxFrameSize")
ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.pekko.http.javadsl.settings.Http2ServerSettings.getMaxFrameSize")
```

MiMa reports `ReversedMissingMethodProblem` for additions to traits even when they have a body, which is why #1216 filtered `withMaxHeaderListSize` and `getMaxHeaderListSize` despite both being concrete. The PR body records only `sbt http-core/mimaReportBinaryIssues`, not the cross-build `+mimaReportBinaryIssues` that CI runs — please re-run that after adding these.

http-core/src/main/resources/reference.conf
```diff
@@ -254,6 +254,20 @@ pekko.http {
...
+      # can make this endpoint hold at once. The length field of a frame header allows up to 16 MiB - 1.
+      #
+      # This is a limit, not an invitation: no larger SETTINGS_MAX_FRAME_SIZE is advertised, so a peer that follows
+      # the spec keeps to the 16 KiB default of RFC 9113, section 4.2 and the extra room here is only leniency for
+      # peers that do not. Advertising a value far above the initial 64 KiB flow-control window would invite frames
+      # that cannot be received without breaking flow control anyway.
+      #
+      # RFC 9113, section 4.2 constrains this to be between 16 KiB and 16 MiB - 1. The amount of request data
+      # buffered overall is bounded separately by the incoming-connection-level-buffer-size and
+      # incoming-stream-level-buffer-size settings above.
```
He-Pin: Nit: `incoming-connection-level-buffer-size` and `incoming-stream-level-buffer-size` are defined *below* this block, not above — `max-frame-size` sits right after `max-header-list-size`. Same wording in the client block further down.

He-Pin (changes requested):
The bound itself is correct. `ByteReader.readShortBE()` and `readByte()` both return masked `Int`s (`pekko/stream/.../ByteStringParser.scala:242-247,216-222`), so `length` is always a valid non-negative 24-bit value and the check cannot be bypassed by a large declared length; `take(n)` throws `NeedMoreData` and `ParsingLogic` keeps buffering, so the unbounded 16 MiB buffering it addresses is real. `IllegalHttp2FrameSize` maps to `ErrorCode.FRAME_SIZE_ERROR` and closes the connection, matching RFC 9113 §4.2, and not advertising `SETTINGS_MAX_FRAME_SIZE` is consistent with `Http2Demux`, which advertises only MAX_CONCURRENT_STREAMS / MAX_HEADER_LIST_SIZE / ENABLE_PUSH — so no spec-compliant peer is affected.

Two things block it: a second oversized-frame test that landed on `main` after this branch, and two missing MiMa filters. Details inline.

http2-tests/src/test/scala/org/apache/pekko/http/impl/engine/http2/Http2ServerSpec.scala
pjfanning: Good catch, thanks. Rebased onto main and added the same `withMaxFrameSize(1024 * 1024)` override to that test in 7d5398ef2; both instances of it pass now. PR body updated too.

http-core/src/main/mima-filters/2.0.x.backwards.excludes/max-frame-size.excludes
pjfanning: Added both, and re-ran `sbt "+http-core/mimaReportBinaryIssues"` — clean on 2.13.18 and 3.3.8. The rebase also pulled in the 1.0.0 comparison from `mima-1.2.0-additional-checks.excludes`, which flagged five more problems from this PR (the replaced internal `Http2FrameParsing` constructor, the abstract `Http2CommonSettings.maxFrameSize`, and three `copy$default$N` shifts from the new case-class field). Those are filtered in the same file, same shape as the `maxHeaderListSize` ones.

http-core/src/main/resources/reference.conf
pjfanning: Fixed in both blocks.

pjfanning:
@He-Pin would you have time to re-review this?

PR #1266, opened 2026-08-31
feat: bound the number of parts in a multipart entity

pjfanning:
### Motivation

`BodyPartParser` had no limit on how many body parts one multipart entity may contain. Each part costs a set of parsed headers and an entity of its own, so a body packed with minimal parts — a boundary, a short `Content-Disposition` and an empty body, on the order of 40–60 bytes each — amplifies the work and allocation a request of a given size causes. A body at the default `max-content-length` of 8m runs to **well over a hundred thousand parts**, all of which the strict and form-field paths materialise (a `VectorBuilder`, a field map, a `HttpEntity.Strict` per part).

`max-content-length` bounds the bytes but not that amplification. Peer frameworks grew an explicit part limit for the same reason — Commons FileUpload's `FileCountLimit` (FILEUPLOAD-279), Spring's `maxParts`, Jetty's `maxFormKeys`.

### Modification

Add a `max-part-count` parser setting, enforced where the parser starts the headers of a new part, failing the entity once it is exceeded. The three call sites that begin a part now go through `parsePartHeaderLines`; the recursive calls that continue the current part's headers are unaffected, as is the closing boundary, which does not start a part.

### On the default

**10000 is deliberately generous, and I'd welcome a steer on it.** It is chosen to stay above the largest part count the test suite exercises — an existing case, "many small parts received in one go", that parses **5000** parts — rather than to be the tightest useful bound. It still cuts the worst case by more than an order of magnitude, and an application that knows its forms are small can set it far lower.

I did not want to silently drop a documented capability for a low-severity amplification issue, which is why I sized the default around that test rather than adjusting the test. If you'd rather have a tighter default (1000 would match Jetty's form-key limit and is ample for real forms), that test can raise its own limit via config instead — say the word and I'll flip it.

### Result

A multipart body can no longer be packed with an unbounded number of parts; the count is bounded by configuration.

### Tests

- `sbt "http-tests/testOnly ...MultipartUnmarshallersSpec* ...MarshallingSpec ...FileUploadDirectivesSpec"` — pass (95 tests), including the existing 5000-part case. A new test configures a limit of 2, sends 3 parts and expects the entity to fail; verified it fails with the limit check disabled, in both the CRLF and LF variants.
- `sbt http-core/mimaReportBinaryIssues`, `sbt http/mimaReportBinaryIssues` — pass, with an exclude for the new method on the internal `BodyPartParser.Settings` (following the existing `max-chunk-count.excludes` precedent).
- `sbt "+http-core/compile"` — pass on 2.13.18 and 3.3.8.
- Native `scalafmt` clean.

### References

None - bounds the number of parts in a multipart entity

### DIFF
--- a/http-core/src/main/mima-filters/2.0.x.backwards.excludes/max-part-count.excludes
+++ b/http-core/src/main/mima-filters/2.0.x.backwards.excludes/max-part-count.excludes
@@ -17,3 +17,5 @@
 
 # new max-part-count setting
 ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.pekko.http.impl.engine.parsing.BodyPartParser#Settings.maxPartCount")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.pekko.http.javadsl.settings.ParserSettings.getMaxPartCount")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.pekko.http.scaladsl.settings.ParserSettings.maxPartCount")
--- a/http-core/src/main/mima-filters/2.0.x.backwards.excludes/max-part-count.excludes
+++ b/http-core/src/main/mima-filters/2.0.x.backwards.excludes/max-part-count.excludes
@@ -1,0 +1,19 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# new max-part-count setting
+ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.pekko.http.impl.engine.parsing.BodyPartParser#Settings.maxPartCount")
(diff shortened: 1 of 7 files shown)


samueleresca (changes requested):
It is possible to bypass the check by using a multi-part with empty headers and body:
```
diff --git a/http-tests/src/test/scala/org/apache/pekko/http/scaladsl/unmarshalling/MultipartUnmarshallersSpec.scala b/http-tests/src/test/scala/org/apache/pekko/http/scaladsl/unmarshalling/MultipartUnmarshallersSpec.scala
index a3e3f6622..812cfbac7 100644
--- a/http-tests/src/test/scala/org/apache/pekko/http/scaladsl/unmarshalling/MultipartUnmarshallersSpec.scala
+++ b/http-tests/src/test/scala/org/apache/pekko/http/scaladsl/unmarshalling/MultipartUnmarshallersSpec.scala
@@ -268,6 +268,21 @@ trait MultipartUnmarshallersSpec extends PekkoSpecWithMaterializer {
           1.second.dilated).getMessage shouldEqual
         "multipart entity contains more than the configured limit of 2 parts"
       }
+      "more empty parts than the configured limit (should be rejected, but is not)" in {
+        // Reproduces the max-part-count bypass: a run of consecutive empty parts (boundary
+        // immediately followed by another boundary, no headers/body in between) is emitted via
+        // the `BoundaryHeader` branch of parseHeaderLines, which recurses into parseHeaderLines
+        // directly instead of parsePartHeaderLines - so partCount is never incremented past 1 and
+        // the configured max-part-count limit is never enforced for this entity shape.
+        implicit val parserSettings: ParserSettings = ParserSettings(system).withMaxPartCount(2)
+        val body = ("--12345" + lineFeed) * 5 + "--12345--"
+
+        Await.result(
+          Unmarshal(HttpEntity(`multipart/mixed`.withBoundary("12345"), ByteString(body)))
+            .to[Multipart.General].failed,
+          1.second.dilated).getMessage shouldEqual
+        "multipart entity contains more than the configured limit of 2 parts"
+      }
       "a stray boundary" in {
         Await.result(
```

samueleresca (approved):

PR #1307, opened 2026-09-18
Remove external sbt-dependency-graph plugin

### LINKED ISSUE #452: Remove dependency on `sbt-dependency-graph`
Turns out, this line

https://github.com/apache/incubator-pekko-http/blob/a6f2ff64a8a028b39a662b460da1ba4c1fe81105/project/plugins.sbt#L20

is not needed anymore since sbt 1.4, because it's build in now.
See https://www.scala-sbt.org/1.x/docs/sbt-1.4-Release-Notes.html#sbt-dependency-graph+is+in-sourced

Actually for advanced usage you can replace it with 
```
addDependencyTreePlugin
```

However it's not that easy, because you make use of apis here:
https://github.com/apache/incubator-pekko-http/blob/a6f2ff64a8a028b39a662b460da1ba4c1fe81105/project/ValidatePullRequest.scala#L17-L18

which in sbt are now [internal](https://github.com/sbt/sbt/blob/v1.9.8/main/src/main/scala/sbt/internal/graph/model.scala#L67)...
(Maybe it's easy to fix, but didn't try, leave that up to you)

mkurz:
- Fixes #452

### Motivation

sbt has included dependency-tree support since 1.4. Pekko HTTP uses sbt 1.13.0, so the legacy `net.virtual-void` dependency is no longer needed.

The pull-request validator still relied on graph model classes supplied by that plugin.

### Modification

- Replace `sbt-dependency-graph` with `addDependencyTreePlugin`.
- Read module IDs directly from sbt's public `UpdateReport` data.
- Preserve dependency analysis across the Compile, Test, Runtime, Provided, and Optional configurations.
- Preserve inclusion of the current project when checking changed modules.

### Result

The build uses the dependency-tree implementation supplied for sbt 1.13.0 and no longer depends on the legacy plugin APIs.

Pull-request impact analysis retains the previous project-selection behavior.

### Tests

- `sbt -batch "show sbtVersion"` — passed; loaded the build and compiled 21 meta-build sources with sbt 1.13.0.
- Scalafmt — passed.
- `git diff --check` — passed.
- Test suites not run per request.


mkurz:
Green.

raboof (approved):
seems like a nice cleanup!

pjfanning (approved):
lgtm

Philippus (approved):

mkurz:
Any eta for v2.0.0-M2? I recommend to cut before end of September because starting with Oct. 1st sonatype will enforce publishing limits: https://central.sonatype.org/news/20260716_publisher_pro_update/

